PR #54, opened 2023-07-20
Classification utils2

borgr:


elronbandel (commented):
Have you changed their names in the tests?

examples/add_metric_to_catalog.py
elronbandel, line 1: changed also in the tests depend on this name?

prepare/splitters/missing_split.py
elronbandel, line 1: Have you tested these works?

elronbandel: Looks like the slice split using wrong syntax

src/unitxt/artifact.py
elronbandel, line 1: what made you do this?

src/unitxt/card.py
elronbandel, line 1: kw_only cannot be used since it is compatible only with python 3.10. 
In the near future we will change the dataclass backbone and it will allow this behavior by default.

src/unitxt/task.py
elronbandel, line 1: amazing. should have been merged imdiatly and serpratly

src/unitxt/operators.py
elronbandel, line 1: This is great. should be added immediately.

elronbandel (changes requested):
Please read the comments. try your best, if possible to break to two pull requests, essential stuff and datasets (where each dataset is separate) again only if possible

PR #76, opened 2023-07-25
new metric rouge

gitMichal:


elronbandel (approved):
Rouge in the old unitxt had some preprocessing that is not included here. (something to do with separation of sentences) this might affect the results.

PR #88, opened 2023-07-27
Github actions: Install the project itself.

eladven:


borgr (approved):
Sometimes code just makes you smile
This, however, is not the first example that pops into mind

PR #120, opened 2023-08-03
Lmeval

borgr:
Added multiple choice (and QA) and many many datasets.
Sharing templates should be done in a different way (but not supported currently)
Function to ease creation should move to separate files in the future?


elronbandel:
fixed in different pr

PR #174, opened 2023-08-15
Add wer metric

yoavkatz:
Added word error rate metric and adjusted tests to deal with metrics which are low (e.g 0) when correct and high (e.g. 1) when not.


elronbandel (approved):

yoavkatz:
Fixed in another pull request.

PR #212, opened 2023-09-19
Add recipe with index access and catalog access

elronbandel:


perlitz (approved):

yoavkatz (approved):
Approve. Suggest to added error checks (added specific comments)

PR #235, opened 2023-09-26
Fix bug in key val template

elronbandel:


yoavkatz (approved):

yoavkatz:
As we discussed, you should add a check that empty templates is used only if there is only one field.

PR #239, opened 2023-10-01
Add class balancing to standard recipe

elronbandel:
The balancing is at the "target" level after rendering so every type of class or answer can be balanced


src/unitxt/operators.py
```diff
@@ -734,3 +735,52 @@ def process(self, instance: Dict[str, Any], stream_name: str = None) -> Dict[str
...
+
+    def process(self, stream: Stream, stream_name: str = None) -> Generator:
+        counter = collections.Counter()
+
+        for instance in stream:
+            counter[self.signature(instance)] += 1
+
+        lowest_count = counter.most_common()[-1][-1]
+
+        max_total_instances = lowest_count
```
yoavkatz, line 776: max_total_instance_per_signiture

yoavkatz (commented):
Should look like this in the API:

card=wnli,train_balancer=operator.balance.determistics,max_train_instances=50,max_test_instance=100

tests/test_operators.py
```diff
@@ -328,3 +329,22 @@ def test_take_by_field(self):
             targets=targets,
             tester=self,
         )
+
+    def test_detirministic_balancer(self):
```
yoavkatz, line 333: deterministic

src/unitxt/operators.py
```diff
@@ -734,3 +735,60 @@ def process(self, instance: Dict[str, Any], stream_name: str = None) -> Dict[str
             dict_set(instance, field, new_values, use_dpath=True, set_multiple=True)
 
         return instance
+
+
+class StreamRefiner(SingleStreamOperator):
+    max_instances: int = None
```
yoavkatz, line 741: Why is this a class member? should it be a local variable?

elronbandel: It is to allow control over the max instances of all the streams instead of having to define max per stream

yoavkatz (approved):

PR #240, opened 2023-10-02
Changed score to 0-1 instead of 0-100

yoavkatz:


elronbandel (approved):

yoavkatz:
Now support both normalized (0-1) and normal (0-100) score.

elronbandel (approved):

PR #250, opened 2023-10-09
Add augmentation capabilties

yoavkatz:
Created support for augmentation in unitxt.

The use is be:

card=cards.sst,template_card_index=0,demos_pool_size=100,num_demos=0,augmentor=augmentors.augment_whitespace

Augmentors are FieldOperator over the 'source' field , and are applied after rendering is done.

Added a first augmentor called AugmentWhitespace that replaces each whitespace is 1-3 spaces, tabs or new lines.

Also improved some error messages.


yoavkatz:
The API was changed to support two ways of augmenting: the full input provided to the model, or only  specific input fields defined in the task (in a new optional field called "argumentable_inputs".

"augmentors.augment_whitespace_model_input" vs "augmentors.augment_whitespace_task_input"

OfirArviv:
> augmentors.augment_whitespace_task_input

can you give an example? It seems like its the same field and not a new field

prepare/cards/sst2.py
```diff
@@ -29,9 +29,7 @@
 add_to_catalog(one_sentence_classification_templates, "templates.one_sent_classification", overwrite=True)
 
 one_sentence_classification_task = FormTask(
-    inputs=["choices", "sentence"],
-    outputs=["label"],
-    metrics=["metrics.accuracy"],
+    inputs=["choices", "sentence"], outputs=["label"], metrics=["metrics.accuracy"], augmentable_inputs=["sentence"]
```
OfirArviv, line 32: what is the augmentation?

src/unitxt/catalog/tasks/one_sent_classification.json
```diff
@@ -9,5 +9,8 @@
     ],
```
OfirArviv, line 9: what is the augmentation type?

src/unitxt/catalog/augmentors/augment_whitespace_task_input.json
```diff
@@ -0,0 +1,4 @@
+{
```
OfirArviv, line 1: how will it look if its outside code?

src/unitxt/task.py
```diff
@@ -11,6 +11,13 @@ class FormTask(Tasker, StreamInstanceOperator):
     inputs: List[str]
     outputs: List[str]
     metrics: List[str]
+    augmentable_inputs: List[str] = []
+
+    def verify(self):
```
OfirArviv, line 16: please explain, could it not be its not the same?

prepare/cards/sst2.py
yoavkatz: There are two modes of augmentation.    One where the entire input provide to the model is augmented (called model input augmentation) .  This includes the any system prompt,  instructions or demonstrations.   The second type of augmentation (called task input augmentation) , augments only specific fields in the task's input.   Augmentable inputs are the subsets of the task's input field that undergo  augmentation.  For example, in sst ,  only the 'sentence' input is augmentwed and not the "choices", which is fixed.

src/unitxt/catalog/augmentors/augment_whitespace_task_input.json
yoavkatz: The augmentor object has two booleans "augment_model_input" and "augment_task_input".   Only one can be set (this is verified in the code).  The  standard template passes the needed input (model_input, ot task_input) to the augmentor , depending on these booleans.  The augmentor changes the relevant text.

In the catalog we save the augmentor , including the value of these flags, which determine it's behavior.

The users of the standard template, just calls with "augmentor=augmentors.augment_whitespace_model_input" or "augmentor=augmentors.augment_whitespace_model_input", and this creates the relevant augmentor based.

prepare/cards/sst2.py
OfirArviv: yes, but is it whitespace augmentation? where is it stated?

tests/test_operators.py
```diff
@@ -386,3 +387,36 @@ def test_length_balancer(self):
...
+
+    def test_augment_whitespace_task_input_with_error(self):
+        text = "The dog ate my cat"
+        inputs = [{"inputs": {"text": text}}]
+        operator = AugmentWhitespace(augment_task_input=True)
+        operator.set_task_input_fields(["sentence"])
+        with self.assertRaises(ValueError):
+            outputs = apply_operator(operator, inputs)
+
+    def test_augment_whitespace_task_input(self):
```
OfirArviv, line 412: add a test that its really only affecting the input and not the rest

OfirArviv (changes requested):

tests/test_operators.py
yoavkatz: There is such a test.  But it requires the standard recipe.  So it's in test_recipe.py:

  def test_standard_recipe_with_augmentor_on_task_input(self):
        recipe = StandardRecipeWithIndexes(
            card="cards.sst2",
            augmentor="augmentors.augment_whitespace_task_input",
            template_card_index=0,
            max_train_instances=0,
            max_test_instances=2,
        )
        stream = recipe()
        sample = list(stream["test"])[1]
        source = sample["source"]
        pattern = "Given this sentence: (.*), classify if it is negative, positive.\s*"
        result = re.match(pattern, sample["source"], re.DOTALL)
        assert result, f"Unable to find '{pattern}' in '{source}'"
        result = result.group(1)
        original_text = "unflinchingly bleak and desperate "
        assert (
            result != original_text
        ), f"Augmented text '{result}' is equal to text without '{original_text}' and was not augmented"
        normalized_output_source = result.split()
        normalized_input_source = original_text.split()
        assert (
            normalized_output_source == normalized_input_source
        ), f"{normalized_output_source} is not equal to f{normalized_input_source}"

OfirArviv (approved):

PR #253, opened 2023-10-16
Add load limit

yoavkatz:
Added a env new variable called UNITEXT_IBM_COS_LOADER_LINE_LIMIT which is used to limit the number of lines downloaded from COS.  This is used to speedup generating datacards (using the prepare/cards/*.py files) .  

This works only for datsets that are in jsonl format.   

In another change avoid placing random file names in output
(To simplify diffs comparison)

### DIFF
--- a/requirements/tests.rqr
+++ b/requirements/tests.rqr
@@ -1,3 +1,4 @@
 bert_score
 transformers
 sentence_transformers
+ibm-cos-sdk
--- a/src/unitxt/catalog.py
+++ b/src/unitxt/catalog.py
@@ -100,7 +100,7 @@ class GithubCatalog(LocalCatalog):
 def verify_legal_catalog_name(name):
     assert re.match(
         r"^[\w" + COLLECTION_SEPARATOR + "]+$", name
-    ), f'Catalog name should be alphanumeric, ":" should specify dirs (instead of "/"). Name: {name}'
+    ), f'Artifict name ("{name}") should be alphanumeric. Use "." for nesting (e.g. myfolder.my_artifact)'
 
 
 def add_to_catalog(
--- a/src/unitxt/catalog.py
+++ b/src/unitxt/catalog.py
@@ -56,7 +56,9 @@ class LocalCatalog(Catalog):
             return False
         return os.path.exists(path) and os.path.isfile(path)
 
-    def save_artifact(self, artifact: Artifact, artifact_identifier: str, overwrite: bool = False):
+    def save_artifact(
+        self, artifact: Artifact, artifact_identifier: str, overwrite: bool = False, verbose: bool = True
+    ):
         assert isinstance(artifact, Artifact), f"Input artifact must be an instance of Artifact, got {type(artifact)}"
         if not overwrite:
             assert (
@@ -65,7 +67,8 @@ class LocalCatalog(Catalog):
         path = self.path(artifact_identifier)
         os.makedirs(Path(path).parent.absolute(), exist_ok=True)
         artifact.save(path)
-        print(f"Artifact {artifact_identifier} saved to {path}")
+        if verbose:
+            print(f"Artifact {artifact_identifier} saved to {path}")
 
 
 class EnvironmentLocalCatalog(LocalCatalog):
@@ -101,12 +104,19 @@ def verify_legal_catalog_name(name):
 
 
 def add_to_catalog(
-    artifact: Artifact, name: str, catalog: Catalog = None, overwrite: bool = False, catalog_path: str = None
+    artifact: Artifact,
+    name: str,
+    catalog: Catalog = None,
+    overwrite: bool = False,
+    catalog_path: str = None,
+    verbose=True,
 ):
     if catalog is None:
         if catalog_path is None:
             catalog_path = default_catalog_path
         catalog = LocalCatalog(location=catalog_path)
     verify_legal_catalog_name(name)
-    catalog.save_artifact(artifact, name, overwrite=overwrite)  # remove collection (its actually the dir).
+    catalog.save_artifact(
+        artifact, name, overwrite=overwrite, verbose=verbose
+    )  # remove collection (its actually the dir).
     # verify name
--- a/src/unitxt/catalog.py
+++ b/src/unitxt/catalog.py
@@ -100,7 +100,7 @@ class GithubCatalog(LocalCatalog):
 def verify_legal_catalog_name(name):
     assert re.match(
         r"^[\w" + COLLECTION_SEPARATOR + "]+$", name
-    ), 'Catalog name should be alphanumeric, ":" should specify dirs (instead of "/").'
+    ), f'Catalog name should be alphanumeric, ":" should specify dirs (instead of "/"). Name: {name}'
 
 
 def add_to_catalog(
--- a/src/unitxt/loaders.py
+++ b/src/unitxt/loaders.py
@@ -20,6 +20,13 @@ except ImportError:
 
 
 class Loader(SourceOperator):
+    # The loader_limit an optional parameter used to control the maximum number of instances to load from the the source.
+    # It is usually provided to the loader via the recipe (see standard.py)
+    # The loader can use this value to limit the amount of data downloaded from the source
+    # to reduce loading time.  However, this may not always be possible, so the
+    # loader may ingore this.  In any case, the recipe, will limit the number of instances in the returned
+    # stream after, after load is complete.
+    loader_limit: int = None
     pass
 
 
@@ -58,7 +65,6 @@ class LoadFromIBMCloud(Loader):
     endpoint_url_env: str
     aws_access_key_id_env: str
     aws_secret_access_key_env: str
-    loader_line_limit_env = "UNITEXT_IBM_COS_LOADER_LINE_LIMIT"
     bucket_name: str
     data_dir: str
     data_files: Sequence[str]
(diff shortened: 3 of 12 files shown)


OfirArviv (commented):
what is the use case? In general I'm not in favor on environment params unless there is a good reason

yoavkatz:
> what is the use case? In general I'm not in favor on environment params unless there is a good reason

The use case is that if I want to 'prepare' json cards from all the .py files , I don't really need to load all the data (can be 100MB) from COS.  This can reduce the time considerably - saving a minute for each card generated.

src/unitxt/loaders.py
```diff
@@ -19,6 +20,13 @@
 
 
 class Loader(SourceOperator):
+    # The loader_limit an optional parameter used to control the maximum number of instances to load from the the source.
+    # It is usually provided to the loader via the recipe (see standard.py)
+    # The loader can use this value to limit the amount of data downloaded from the source
+    # to reduce loading time.  However, this may not always be possible, so the
+    # loader may ingore this.  In any case, the recipe, will limit the number of instances in the returned
```
OfirArviv, line 27: is the loader per card?
if so, isn't it possible to error out if its not possible to limit the load?
because the source is fixed in the card
Why not to throw an error?

src/unitxt/loaders.py
```diff
@@ -62,13 +70,24 @@ class LoadFromIBMCloud(Loader):
...
-        print(f"Downloading {item_name} from {bucket_name} COS to {local_file}")
+        print(f"Downloading {item_name} from {bucket_name} COS")
         try:
             response = cos.Object(bucket_name, item_name).get()
             size = response["ContentLength"]
+            body = response["Body"]
         except Exception as e:
             raise Exception(f"Unabled to access {item_name} in {bucket_name} in COS", e)
 
+        if self.loader_limit != None:
```
OfirArviv, line 81: you are still downloading everything, no? What am I missing?

src/unitxt/operators.py
```diff
@@ -858,6 +858,9 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
         for instance in stream:
             counter[self.signature(instance)] += 1
 
+        if len(counter) == 0:
```
OfirArviv, line 861: why can we get 0? should we allow that?

src/unitxt/standard.py
```diff
@@ -132,10 +158,21 @@ class StandardRecipeWithIndexes(BaseRecipe):
     def prepare(self):
         assert (
             self.template_card_index is None or self.template is None
```
OfirArviv, line 160: is this related  another issue?

src/unitxt/standard.py
```diff
@@ -132,10 +158,21 @@ class StandardRecipeWithIndexes(BaseRecipe):
...
-        ), "Specify either template or template_card_index"
+        ), f"Specify either template ({self.template}) or template_card_index ({self.template_card_index}) but not both"
+        assert (
+            not self.template_card_index is None or not self.template is None
```
OfirArviv, line 163: confusing  logic

src/unitxt/standard.py
```diff
@@ -50,12 +52,36 @@ def verify(self):
                 raise ValueError(
                     f"demos_pool_size must be bigger than num_demos={self.num_demos}, Got demos_pool_size={self.demos_pool_size}"
                 )
+            if self.loader_limit and self.demos_pool_size > self.loader_limit:
```
OfirArviv, line 55: you have if self.loader_limit in all of them. Do a single if on top of it all.

Also what if train_validatation > the limit for example?

src/unitxt/loaders.py
yoavkatz: Loader limit is set to each of the loaders by the calling recipe.

src/unitxt/operators.py
yoavkatz: It can happen if by mistake someone filters too many results.  In most cases, it should not happen, once I added the checks in the standard recipe verify method, but it can happen .    The question is more general - can streams be empty.   But we we allow empty streams, we need operators to behave correctly over them.

src/unitxt/standard.py
yoavkatz: I'll move the if to the top.

self.max_validation_instances is checked below.

requirements/tests.rqr
```diff
@@ -1,3 +1,4 @@
 bert_score
 transformers
 sentence_transformers
+ibm-cos-sdk
```
OfirArviv, line 4: we test ibm-cos here? how?

src/unitxt/standard.py
```diff
@@ -50,12 +52,35 @@ def verify(self):
...
+                raise ValueError(
+                    f"demos_pool_size must be bigger than loader_limit{self.loader_limit}, Got demos_pool_size={self.demos_pool_size}"
+                )
```
OfirArviv, line 58: this one is not in the general if

src/unitxt/test_utils/card.py
```diff
@@ -26,44 +27,38 @@ def test_metrics_exist(card):
...
         assert json.dumps(card_.to_dict(), sort_keys=True) == json.dumps(
             card.to_dict(), sort_keys=True
         ), "Card loaded is not equal to card stored"
 
 
-def load_examples_from_common_recipe(card, tested_split):
-    if card.templates:
-        num_templates = len(card.templates)
+def load_examples_from_standard_recipe(card, tested_split, template_card_index):
+    print("=" * 80)
```
OfirArviv, line 39: is this intentional?

src/unitxt/test_utils/card.py
```diff
@@ -26,44 +27,38 @@ def test_metrics_exist(card):
...
-    recipe = CommonRecipe(
+    recipe = StandardRecipe(
         card=card,
         demos_pool_size=100,
         demos_taken_from=tested_split,
         num_demos=3,
-        template_item=template_item,
-        instruction_item=0 if num_instructions else None,
+        template_card_index=template_card_index,
+        instruction_card_index=0 if num_instructions else None,
```
OfirArviv, line 49: this is never None according to line 48

tests/test_recipe.py
```diff
@@ -112,6 +112,77 @@ def test_standard_recipe_with_balancer(self):
...
+            instruction="instructions.models.llama",
+            template="templates.key_val",
+            format="formats.user_agent",
+            demos_pool_size=5,
+            num_demos=1,
+            loader_limit=10,
+        )
+
+        stream = recipe()
+        self.assertEqual(len(list(stream["train"])), 5)
```
OfirArviv, line 127: this is confusing, why only 5 and not 10 also?

tests/test_recipe.py
```diff
@@ -112,6 +112,77 @@ def test_standard_recipe_with_balancer(self):
...
+    def test_standard_recipe_with_template_errors(self):
+        with self.assertRaises(AssertionError) as cm:
+            recipe = StandardRecipeWithIndexes(card="cards.wnli")
+        self.assertEqual(str(cm.exception), "Specify either template or template_card_index in card")
+
+        with self.assertRaises(AssertionError) as cm:
+            recipe = StandardRecipeWithIndexes(
+                card="cards.wnli", template="templates.key_val", template_card_index=100
+            )
+        self.assertTrue("Specify either template" in str(cm.exception))
```
OfirArviv, line 172: you are missing " or template_card_index in card" I think , and the assert is weird, why 2?

OfirArviv: maybe a better idea is to add a short description to each test, to explain the scenario

requirements/tests.rqr
yoavkatz: We test the API using mocks. See test_loaders.py.

src/unitxt/standard.py
yoavkatz: Right, because it also requires multi  conditions on the poll size number of demos, which are in the inclosing if statements.

src/unitxt/test_utils/card.py
yoavkatz: Yes. It makes the output more readible.

src/unitxt/test_utils/card.py
yoavkatz: num_instructions is never None, but 
  
  "if num_instructions " does not check if the (int) variable is None, it checks if it's non zero:
  
  https://stackoverflow.com/questions/73241173/what-is-the-meaning-of-the-if-0-in-python

tests/test_recipe.py
yoavkatz: Because 5 elements are taken to the demo pool.  Added comment.

tests/test_recipe.py
yoavkatz: Changed to regex search and added short description

src/unitxt/catalog.py
```diff
@@ -97,16 +100,23 @@ def __contains__(self, artifact_identifier: str):
 def verify_legal_catalog_name(name):
     assert re.match(
         r"^[\w" + COLLECTION_SEPARATOR + "]+$", name
-    ), 'Catalog name should be alphanumeric, ":" should specify dirs (instead of "/").'
+    ), f'Catalog name should be alphanumeric, ":" should specify dirs (instead of "/"). Name: {name}'
```
OfirArviv, line 103: I think this error is phrased in a confusing way, but it is not a blocker

src/unitxt/standard.py
```diff
@@ -48,14 +50,37 @@ def verify(self):
                 )
             if self.demos_pool_size < self.num_demos:
                 raise ValueError(
-                    f"demos_pool_size must be bigger than num_demos={self.num_demos}, Got demos_pool_size={self.demos_pool_size}"
+                    f"demos_pool_size must be bigger than num_demos ({self.num_demos}), Got demos_pool_size={self.demos_pool_size}"
+                )
+            if self.loader_limit and self.demos_pool_size > self.loader_limit:
```
OfirArviv, line 55: I thin it should be >=. We don't want all the examples to go into the demo pool either, right?

OfirArviv (approved):

src/unitxt/catalog.py
yoavkatz: Improved

src/unitxt/standard.py
yoavkatz: Not clear - if you don't do fine/prompt tuning it's not a problem to take all training to the demo pool.

PR #254, opened 2023-10-16
Fixed bug in random initialization

yoavkatz:
Default seed was not set, so initially each run
generated different selection.

Was less observed, because used of nested_seed() set the seed.

Also added unit test.


matanor (approved):
LGTM

PR #260, opened 2023-10-22
ethos classification

naamaz:


prepare/cards/ethos.py
```diff
@@ -35,6 +35,16 @@
                 output_format="{label}",
                 postprocessors=["processors.take_first_non_empty_line"],
             ),
+            InputOutputTemplate(
+                input_format="Given this sentence: {text}. Classify if it contains hatespeech. Choices: {choices}. I would classify this sentence as: ",
```
yoavkatz, line 39: It's "hate speech" more comon than "hatespeech"? Or is this the expected output?

naamaz: These were the original choices and therefore this is the expected output. It works this way. I can try "hatespeech" instead and check. should work as well

yoavkatz (changes requested):
Overall looks good.  Minor changes request.

prepare/cards/ethos.py
yoavkatz: The choices are "hate speech" and "not hate speech" so why are the instructions "Classify if it contains hatespeech"? (without a space).  Maybe it does not make a difference, but it seems wrong.

yoavkatz (approved):
I approve, but (1) I think you can still call it   "processors.hate_or_not"  or even "processors.hate_speech_or_not_hate_speech" (and" not string_or_not_hate", which is not clear)

src/unitxt/processors.py
```diff
@@ -74,3 +74,22 @@ def process(self, instance):
...
+    def process(self, instance):
+        non_empty_line = instance.lower()
+        match = re.search(r"[.,!?;]", non_empty_line)
+        if match:
+            # Extract text up to the first punctuation
+            non_empty_line = non_empty_line[: match.start()]
+        return non_empty_line
+
+
+class HateOrNot(BaseFieldOperator):
```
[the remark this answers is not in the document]
naamaz: Done. Chose "hate speech" since it is not common to use "hatespeech" as far as I could see.

yoavkatz (approved):

PR #264, opened 2023-10-31
Fixes slash in IBM COS

### LINKED ISSUE #262: Remove slash (/) from object key in LoadFromIBMCloud
The slash character is not allowed as a valid character in the object key when storing files in IBM COS thus the retrieval of files with this character will lead to an object not being found.

Making `data_dir` as an optional parameter will solve this issue.

gennosuke2k7:
Fixes #262


yoavkatz (approved):

yoavkatz:
Hi. Code looks good, but you need to run a black formatter or simply apply

 pre-commit run --all-files

PR #266, opened 2023-11-01
Fixed single stream operators and enhanced merge functionality

yoavkatz:
Before the change, the StreamInstanceOperator overrided the _process_stream method, and lost the common functionality of checking if operator should be applied based on `apply_to_streams` and `dont_apply_to_streams`.

Now the code was refactored to the checks is always applied.

Another change is to allow specifying which streams to merge, in the MergeStream operator.
Also added unit tests.


tests/test_operators.py
```diff
@@ -170,6 +171,37 @@ def test_split_by_value(self):
...
+            self.assertDictEqual(input_dict, output_dict)
+
+        # test with parameters
+        input_multi_stream = MultiStream(
+            {"test": [{"field": "test1"}], "validation": [{"field": "validation1"}], "train": [{"field": "train1"}]}
+        )
+        output_multi_stream = MergeStreams(
+            streams_to_merge=["test", "train"], new_stream_name="merged", add_origin_stream_name=False
+        )(input_multi_stream)
+        self.assertListEqual(list(output_multi_stream.keys()), ["merged"])
```
matanor, line 198: suggestion: this is the same code as above, IIUC, for comparing a stream to expected value. maybe consider a function for that.

src/unitxt/operator.py
```diff
@@ -170,7 +173,7 @@ def _process_multi_stream(self, multi_stream: MultiStream) -> MultiStream:
     def _process_single_stream(self, stream: Stream, stream_name: str = None) -> Stream:
         return Stream(self._process_stream, gen_kwargs={"stream": stream, "stream_name": stream_name})
 
-    def is_should_be_processed(self, stream_name):
+    def _is_should_be_processed(self, stream_name):
         if (
             self.apply_to_streams is not None
```
matanor, line 178: perhaps add tests for checking the functionality of filtering streams? (i.e. of `self.apply_to_streams`)

src/unitxt/operator.py
```diff
@@ -185,10 +188,7 @@ def is_should_be_processed(self, stream_name):
         ) and stream_name not in self.dont_apply_to_streams
 
     def _process_stream(self, stream: Stream, stream_name: str = None) -> Generator:
-        if self.is_should_be_processed(stream_name):
-            yield from self.process(stream, stream_name)
-        else:
-            yield from stream
+        yield from self.process(stream, stream_name)
```
matanor, line 191: why was the check of `is_should_be_processed` removed here?

matanor (changes requested):
Added a few comments / questions. Please have a look.

tests/test_operators.py
yoavkatz: Done.

src/unitxt/operator.py
yoavkatz: Because the check was moved _process_multi_stream, is calls this method.

The issue was that this method was overriden by derived classes, and that's why the functionally did not work in derived classes. The _process_multi_stream method is not overriden.

matanor (approved):
LGTM

PR #267, opened 2023-11-01
New Yes-No template

matanor:


src/unitxt/templates.py
```diff
@@ -181,6 +181,61 @@ def get_postprocessors(self) -> List[str]:
...
+            return self.input_format.format(**data)
+        except KeyError as e:
+            raise KeyError(
+                f"Available inputs are {inputs.keys()} but input format "
+                f"requires a different one: {self.input_format}"
+            ) from e
+
+    def process_outputs(self, outputs: Dict[str, object]) -> str:
+        try:
+            gold_class_name = outputs[self.label_field]
```
yoavkatz, line 225: Need to add a check and an error message If label_field is not in outputs.

src/unitxt/templates.py
```diff
@@ -181,6 +181,61 @@ def get_postprocessors(self) -> List[str]:
...
+            if self.class_name in gold_class_name:
```
yoavkatz, line 226: Add assert and error message that gold class name is a list which is either empty or has one element.

matanor: Can it contain two or more elements? (for multi-label classification)

yoavkatz: I see.  So maybe call it 'gold_class_names'.

matanor: Fixed

matanor:
@yoavkatz tks for the review, please see the changes following your comment.
also, added tests.

yoavkatz (approved):
I approve, but I think that checking exceptions should be with their expected message. Otherwise, we may mask unreadable messages.


        with self.assertRaises(AssertionError) as cm:
            recipe = StandardRecipeWithIndexes(card="cards.wnli")
        self.assertEqual(str(cm.exception), "Specify either template or template_card_index in card")

        # Check either template or template index was specified , but not both
        with self.assertRaises(AssertionError) as cm:
            recipe = StandardRecipeWithIndexes(
                card="cards.wnli", template="templates.key_val", template_card_index=100
            )
        self.assertTrue(
            not re.match("Specify either template (.*) or template_card_index (.*) but not both", str(cm.exception))
            is None
        )

matanor:
> self.assertTrue(
>         not re.match("Specify either template (.*) or template_card_index (.*) but not both", str(cm.exception))
>         is None
>     )

good idea, i improved the tests to account for the exception messages.

yoavkatz (approved):

PR #271, opened 2023-11-05
Add debug card capabilities

yoavkatz:
Added debug_card utility to help debug cards
    
    At the core, added an ability to run a sequence operator
    only for a given number of steps.
    
    The debug_card utility iterates over all steps in a card and
    prints the for each step all the streams (splits) in the output
    and the first example in each stream.
    
    test_card(card,debug=True  )

### DIFF
--- a/prepare/cards/belebele.py
+++ b/prepare/cards/belebele.py
@@ -182,5 +182,5 @@ for lang in language_codes:
         templates=CONTEXT_MMLU_TEMPLATES_NO_INTRO,
     )
     if lang == "acm_Arab":
-        test_card(card, tested_split="test")
+        test_card(card, demos_taken_from="test", debug=True)
     add_to_catalog(card, f"cards.belebele.{lang}", overwrite=True)
--- a/prepare/cards/sst2.py
+++ b/prepare/cards/sst2.py
@@ -49,5 +49,5 @@ card = TaskCard(
     templates="templates.one_sent_classification",
 )
 
-test_card(card, debug=False)
+test_card(card, debug=True)
 add_to_catalog(card, "cards.sst2", overwrite=True)
--- a/prepare/cards/sst2.py
+++ b/prepare/cards/sst2.py
@@ -49,5 +49,5 @@ card = TaskCard(
     templates="templates.one_sent_classification",
 )
 
-test_card(card)
+test_card(card, debug=False)
 add_to_catalog(card, "cards.sst2", overwrite=True)
--- a/src/unitxt/operator.py
+++ b/src/unitxt/operator.py
@@ -396,7 +396,7 @@ class SourceSequentialOperator(SequentialOperator):
         return super().__call__()
 
     def process(self, multi_stream: Optional[MultiStream] = None) -> MultiStream:
-        assert (self.num_steps() > 0, "Calling process on a SourceSequentialOperator without any steps")
+        assert self.num_steps() > 0, "Calling process on a SourceSequentialOperator without any steps"
         multi_stream = self.steps[0]()
         for operator in self.steps[1 : self._get_max_steps()]:
             multi_stream = operator(multi_stream)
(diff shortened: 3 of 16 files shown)


prepare/cards/belebele.py
```diff
@@ -182,5 +182,5 @@
         templates=CONTEXT_MMLU_TEMPLATES_NO_INTRO,
     )
     if lang == "acm_Arab":
-        test_card(card, tested_split="test")
+        test_card(card, demos_taken_from="test", debug=True)
```
matanor, line 185: just checking, should commit with `debug=True`?

prepare/cards/sst2.py
```diff
@@ -49,5 +49,5 @@
     templates="templates.one_sent_classification",
 )
 
-test_card(card)
+test_card(card, debug=False)
```
matanor, line 52: again, just checking, if that's the default, should it be removed?

src/unitxt/operator.py
```diff
@@ -176,16 +177,17 @@ def _process_single_stream(self, stream: Stream, stream_name: str = None) -> Str
...
             raise ValueError(
                 f"Stream '{stream_name}' can be in either apply_to_streams or dont_apply_to_streams not both."
             )
 
-        return (
-            self.apply_to_streams is None or stream_name in self.apply_to_streams
-        ) and stream_name not in self.dont_apply_to_streams
+        return (self.apply_to_streams is None or stream_name in self.apply_to_streams) and (
+            self.dont_apply_to_streams is None or stream_name not in self.dont_apply_to_streams
+        )
```
matanor, line 190: this part maybe deserves some u-tests (maybe open an issue for that?).

src/unitxt/operator.py
```diff
@@ -351,22 +353,42 @@ def process(self, instance: dict, multi_stream: MultiStream) -> dict:
...
+
+    def set_max_steps(self, max_steps):
+        assert (
+            max_steps <= self.num_steps()
+        ), f"Max steps requested ({max_steps}) is larger than defined steps {self.num_steps()}"
+        assert max_steps >= 1, f"Max steps requested ({max_steps}) is less than 1"
+        self.max_steps = max_steps
+
+    def get_last_step_description(self):
+        last_step = self.max_steps - 1 if not self.max_steps is None else len(self.steps) - 1
```
matanor, line 378:
```suggestion
        last_step = self.max_steps - 1 if self.max_steps is not None else len(self.steps) - 1
```

matanor: (improves readability IMO)

src/unitxt/operator.py
```diff
@@ -351,22 +353,42 @@ def process(self, instance: dict, multi_stream: MultiStream) -> dict:
...
+        description = str(self.steps[last_step])
+        return re.sub(r"\w+=None, ", "", description)
+
+    def _get_max_steps(self):
+        return self.max_steps if not self.max_steps is None else len(self.steps)
```
matanor, line 383:
```suggestion
        return self.max_steps if self.max_steps is not None else len(self.steps)
```

matanor: i think also just `self.max_steps if self.max_steps else len(self.steps)` should work

src/unitxt/splitters.py
```diff
@@ -183,6 +183,10 @@ def process(self, instance: Dict[str, object], multi_stream: MultiStream) -> Dic
             raise Exception(f"Unable to fetch instances from '{self.source_stream}' to '{self.target_field}'")
 
 
+class CreateDemoSplit(SpreadSplit):
+    pass
```
matanor, line 187: what is this part for?

src/unitxt/standard.py
```diff
@@ -108,7 +108,7 @@ def prepare(self):
             sampler.set_size(self.num_demos)
 
             self.steps.append(
-                SpreadSplit(
+                CreateDemoSplit(
```
matanor, line 111: same as above, why the new class?

matanor (approved):
LGTM, left just a few minor comments

prepare/cards/sst2.py
yoavkatz: Left as an example.

src/unitxt/operator.py
yoavkatz: There are tests for it.  See  test_stream_refiner.

src/unitxt/splitters.py
yoavkatz: To give a meaningful name to that part of the pipeline in debug mode.

PR #273, opened 2023-11-06
Add (micro/macro) precision and recall to CustomF1

yoavkatz:
(Currently only aggregates results (and not per class), so not too add too many scores.


matanor (approved):
LGTM

yoavkatz:
The default behavior in sckit learn on divide by zero is to return ‘0’ with a warning.   There is the flag to change to change to 1.0, 0.0 (without warning) and nan (which is a new option).

Set our behavior to 0 as well.

PR #274, opened 2023-11-06
add empty instruction

OfirArviv:


OfirArviv:
@yoavkatz  is this the right way to do it? Conceptually I want it to be the default value for "no instruction"

yoavkatz:
LGTM.  Note you need to commit to unitxt with signoff (-S).

yoavkatz (approved):

PR #281, opened 2023-11-07
Updates to YesNoTemplate

matanor:
Change `YesNoTemplate` to accept a `class_field`, which contains the name of the field with the queried class_name. It is now expected in both the inputs and the outputs to the template.

This replaces the `class_name` parameter, that directly included the queried class name.


yoavkatz (changes requested):
I think the user experience (as well as code and the tests) were made a big more complex, because two errors are catched by the same exception.  I think they can easily be separated.

matanor:
> I think the user experience (as well as code and the tests) were made a big more complex, because two errors are catched by the same exception. I think they can easily be separated.

Sure, separated the tests.

yoavkatz (approved):

PR #284, opened 2023-11-07
Fix mmlu templates

perlitz:


prepare/processors/processors.py
```diff
@@ -17,17 +17,3 @@
 operator3 = StringOrNotString(string="hate speech")
 
 add_to_catalog(operator3, "processors.hate_speech_or_not_hate_speech", overwrite=True)
-
```
yoavkatz, line 20: Why was this code deleted?

prepare/templates/qa/multiple_choice/templates.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def replace_if_context_not_there(s, oldvalue, newvalue):
+    if "{context}" in s:
+        return s
+    else:
+        return s.replace(oldvalue, newvalue)
+
+
+templates_with_context = {
```
yoavkatz, line 33: i must say this is a convulated way to define the templates (inherited in from past code).
Wouldn't it be simpler and more readible, just to write the 4 templates as they are expected to appear?

prepare/templates/qa/multiple_choice/templates.py
```diff
@@ -0,0 +1,104 @@
...
+# )
+
+for k, v in templates_with_context.items():
+    template = InputOutputTemplate(input_format=v, output_format="{label}", postprocessors=["processors.first_character"])
+    add_to_catalog(template, f"templates.qa.multiple_choice.context.{k}", overwrite=True)
+
+
+# context no intro
+templates_context_no_intro = {
+    key: val.replace("The following are multiple choice questions (with answers) about {topic}.", "").strip()
```
yoavkatz, line 62: Here too.  Why not specify the 3 templates?  This replace does not effect the lm_eval_harness template, so we are just adding a duplicate template.

yoavkatz (changes requested):
See below.  One crticial point on deleting the implemention of 
"processors.first_character",  and two points to clarify the code.

yoavkatz (approved):

PR #285, opened 2023-11-07
Added new operators

yoavkatz:
- FilterByListsOfValues :  Filters a stream, yielding only instances that  whose field values are included in the specified values lists.

- Intersect  : Intersects the value of a field, which must be a list,  with a given list

- RemoveValues : Removes values in  of a field, which must be a list,  by a given list 

Added tests and improved error checks

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -686,7 +686,7 @@ class FilterByListsOfValues(SingleStreamOperator):
                 yield instance
 
 
-class IntersectWithList(FieldOperator):
+class Intersect(FieldOperator):
     """
     Intersects the value of a field, which must be a list, with a given list
     Args:
@@ -703,6 +703,25 @@ class IntersectWithList(FieldOperator):
         if not isinstance(value, list):
             raise ValueError(f"The value in field is not a list but '{value}'")
         return [e for e in value if e in self.allowed_values]
+
+
+class RemoveValues(FieldOperator):
+    """
+    Removes elements in a field, which must be a list, using a given list of unallowed
+    Args:
+        unallowed_values (list) - removed_values
+    """
+
+    unallowed_values: List[Any]
+
+    def verify(self):
+        if not isinstance(self.unallowed_values, list):
+            raise ValueError(f"The unallowed_values is not a list but '{self.unallowed_values}'")
+
+    def process_value(self, value: Any) -> Any:
+        if not isinstance(value, list):
+            raise ValueError(f"The value in field is not a list but '{value}'")
+        return [e for e in value if not e in self.unallowed_values]
 
 
 class Unique(SingleStreamReducer):
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -735,6 +735,9 @@ class RemoveValues(FieldOperator):
 
     def verify(self):
         super().verify()
+        if self.process_every_value:
+            raise ValueError(f"'process_every_value=True' is not supported in RemoveValues operator")
+
         if not isinstance(self.unallowed_values, list):
             raise ValueError(f"The unallowed_values is not a list but '{self.unallowed_values}'")
 
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -212,12 +212,19 @@ class FieldOperator(StreamInstanceOperator):
                     default=self.get_default,
                     not_exist_ok=self.not_exist_ok,
                 )
-            except TypeError as e:
-                raise TypeError(f"Failed to get {from_field} from {instance}")
-            if self.process_every_value:
-                new_value = [self.process_value(value) for value in old_value]
-            else:
-                new_value = self.process_value(old_value)
+            except Exception as e:
+                raise ValueError(
+                    f"{self.__class__.__name__}: Failed to get '{from_field}' from {instance} due to : {e}"
+                )
+            try:
+                if self.process_every_value:
+                    new_value = [self.process_value(value) for value in old_value]
+                else:
+                    new_value = self.process_value(old_value)
+            except Exception as e:
+                raise ValueError(
+                    f"{self.__class__.__name__}: Failed to process '{from_field}' from {instance} due to : {e}"
+                )
             if self.use_query and is_subpath(from_field, to_field):
                 dict_delete(instance, from_field)
             dict_set(instance, to_field, new_value, use_dpath=self.use_query, not_exist_ok=True)
@@ -644,10 +651,10 @@ class ApplyOperatorsField(StreamInstanceOperator, ArtifactFetcherMixin):
 
 class FilterByValues(SingleStreamOperator):
     """
-    Filters a stream, yielding only instances that match specified values.
+    Filters a stream, yielding only instances that match specified values in the provided fields.
 
     Args:
-        values (Dict[str, Any]): The values that instances should match to be included in the output.
+        values (Dict[str, Any]): For each field, the values that instances should match to be included in the output.
     """
 
     values: Dict[str, Any]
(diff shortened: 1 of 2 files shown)


src/unitxt/operators.py
```diff
@@ -658,6 +665,65 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+
+    allowed_values: List[Any]
+
+    def verify(self):
+        if not isinstance(self.allowed_values, list):
+            raise ValueError(f"The allowed_values is not a list but '{self.allowed_values}'")
+
+    def process_value(self, value: Any) -> Any:
+        if not isinstance(value, list):
+            raise ValueError(f"The value in field is not a list but '{value}'")
```
matanor, line 704: what if `process_every_value` is `True`? In this case the `value` is only one of the values in the field, and not the entire value of the field. In this case, shouldn't we allow list values? (e.g. the field value is [ ["a", "b"], "c" ].

I think at least the error message should be f"The processed value is not a list but '{value}'".

src/unitxt/operators.py
```diff
@@ -658,6 +665,65 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+class RemoveValues(FieldOperator):
+    """
+    Removes elements in a field, which must be a list, using a given list of unallowed
+    Args:
+        unallowed_values (list) - removed_values
+    """
+
+    unallowed_values: List[Any]
+
+    def verify(self):
```
matanor, line 717:
```suggestion
    def verify(self):
        super.verify()
```

src/unitxt/operators.py
```diff
@@ -658,6 +665,65 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+        if not isinstance(self.unallowed_values, list):
+            raise ValueError(f"The unallowed_values is not a list but '{self.unallowed_values}'")
+
+    def process_value(self, value: Any) -> Any:
+        if not isinstance(value, list):
```
matanor, line 722: Same comment as above

tests/test_operators.py
```diff
@@ -91,6 +94,119 @@ def test_filter_by_values(self):
...
+            {"label": "b"},
+        ]
+        with self.assertRaises(ValueError) as cm:
+            test_operator(
+                operator=Intersect(field="label", allowed_values=["c"]),
+                inputs=inputs,
+                targets=targets,
+                tester=self,
+            )
+        print(cm.exception)
```
matanor, line 152: remove print?

tests/test_operators.py
```diff
@@ -91,6 +94,119 @@ def test_filter_by_values(self):
...
+        )
+
+        with self.assertRaises(ValueError) as cm:
+            test_operator(
+                operator=FilterByListsOfValues(values={"b": "5"}), inputs=inputs, targets=targets, tester=self
+            )
+        self.assertEqual(str(cm.exception), "The filter for key ('b') in FilterByListsOfValues is not a list but '5'")
+
+    def test_intersect(self):
+        inputs = [
```
matanor, line 120: suggestion: add tests when `process_every_value` is `True`

tests/test_operators.py
```diff
@@ -91,6 +94,119 @@ def test_filter_by_values(self):
...
+                targets=targets,
+                tester=self,
+            )
+        print(cm.exception)
+        self.assertEqual(
+            str(cm.exception),
+            "Intersect: Failed to process 'label' from {'label': 'b'} due to : The value in field is not a list but 'b'",
+        )
+
+    def test_remove_values_by_list(self):
```
matanor, line 158:
```suggestion
    def test_remove_values(self):
```

matanor: To match the operator name

tests/test_operators.py
```diff
@@ -91,6 +94,119 @@ def test_filter_by_values(self):
...
+            {"label": "b"},
+        ]
+        with self.assertRaises(ValueError) as cm:
+            test_operator(
+                operator=RemoveValues(field="label", unallowed_values=["c"]),
+                inputs=inputs,
+                targets=targets,
+                tester=self,
+            )
+        print(cm.exception)
```
matanor, line 191: remove print?

tests/test_operators.py
```diff
@@ -91,6 +94,119 @@ def test_filter_by_values(self):
...
+        )
+
+        with self.assertRaises(ValueError) as cm:
+            test_operator(
+                operator=RemoveValues(field="label2", unallowed_values=["c"]),
+                inputs=inputs,
+                targets=targets,
+                tester=self,
+            )
+        print(cm.exception)
```
matanor, line 204: remove print?

matanor (changes requested):
hi @yoavkatz looks good, please see my comments..

tests/test_operators.py
yoavkatz: Done

src/unitxt/operators.py
```diff
@@ -658,6 +665,65 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
                 yield instance
 
 
+class FilterByListsOfValues(SingleStreamOperator):
```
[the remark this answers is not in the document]
yoavkatz: Best to  it's better to keep it consistent.    I think the action we want to is filter the dataset.  I'll change the name of the parameter to 'required_values' in both classes, so it will be clear.

src/unitxt/operators.py
```diff
@@ -658,6 +665,65 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    values: Dict[str, Any]
+
+    def verify(self):
+        for key, value in self.values.items():
+            if not isinstance(value, list):
+                raise ValueError(f"The filter for key ('{key}') in FilterByListsOfValues is not a list but '{value}'")
+
+    def process(self, stream: Stream, stream_name: str = None) -> Generator:
+        for instance in stream:
+            if any(instance[key] in value for key, value in self.values.items()):
```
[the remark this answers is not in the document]
yoavkatz: Added checks and tests.

tests/test_operators.py
yoavkatz: Done

tests/test_operators.py
yoavkatz: Done.

tests/test_operators.py
yoavkatz: I disable this feature for intersect because it would be confusing with nested lists.

src/unitxt/operators.py
```diff
@@ -212,12 +212,19 @@ def process(self, instance: Dict[str, Any], stream_name: str = None) -> Dict[str
...
+                    f"{self.__class__.__name__}: Failed to get '{from_field}' from {instance} due to : {e}"
+                )
+            try:
+                if self.process_every_value:
+                    new_value = [self.process_value(value) for value in old_value]
+                else:
+                    new_value = self.process_value(old_value)
+            except Exception as e:
+                raise ValueError(
+                    f"{self.__class__.__name__}: Failed to process '{from_field}' from {instance} due to : {e}"
```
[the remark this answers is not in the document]
matanor: right, i made an error in the suggestion, syntax should be similar to:
```
        except KeyError as e:
            raise RuntimeError(
                f"Available inputs are {inputs.keys()} but input format "
                f"requires a different one: {self.input_format}"
            ) from e
```
what this does is to set the exception `e` as the direct cause of the raise exception. 
in second thought, i would keep your phrasing of the explicit error message. just perhaps consider adding `from e`.

src/unitxt/operators.py
```diff
@@ -644,20 +651,96 @@ def process(self, instance: Dict[str, Any], stream_name: str = None) -> Dict[str
...
+    required_values: Dict[str, Any]
+
+    def process(self, stream: Stream, stream_name: str = None) -> Generator:
+        for instance in stream:
+            filter = False
+            for key, value in self.required_values.items():
+                if not key in instance:
+                    raise ValueError(f"Required filter field ('{key}') in FilterByValues is not found in {instance}")
+                if instance[key] != value:
+                    filter = True
```
matanor, line 669: `break` at this point? or is the intention to keep the loop so all the keys are checked for their presence?

src/unitxt/operators.py
yoavkatz: I decided not to allow nested list (process_every_value is True) .   It's too complicated.

src/unitxt/operators.py
yoavkatz: I did not think about it originally, but I indeed think it's best to see the keys are in the instance.

src/unitxt/operators.py
```diff
@@ -658,6 +665,65 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    """
+    Intersects the value of a field, which must be a list, with a given list
+    Args:
+        allowed_values (list) - list to intersect
+    """
+
+    allowed_values: List[Any]
+
+    def verify(self):
+        if not isinstance(self.allowed_values, list):
```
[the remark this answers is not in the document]
yoavkatz: Right. Fixed.

src/unitxt/operators.py
yoavkatz: Fixed.

src/unitxt/operators.py
yoavkatz: Same as above.  Prevented (process_every_value is True) .

src/unitxt/operators.py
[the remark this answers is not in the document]
yoavkatz: Done.

src/unitxt/operators.py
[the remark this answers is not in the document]
yoavkatz: Done.

tests/test_operators.py
```diff
@@ -80,16 +83,170 @@ def test_flatten_instances(self):
...
+        ]
+
+        targets = [
+            {"label": ["b"]},
+            {"label": []},
+            {"label": ["b", "f"]},
         ]
 
-        test_operator(operator=FilterByValues(values={"a": 1}), inputs=inputs, targets=targets, tester=self)
+        test_operator(
```
matanor, line 154: suggestion (not crucial): add `test_operator_and_excpect_exception`, or similar. since this seems to repeat several times.

matanor (approved):
LGTM. left two minor comments.

tests/test_operators.py
yoavkatz: Good.  I'll fix this in the next operator I write. I want the operators to be available to Ilya.

PR #288, opened 2023-11-08
fix for issue 286

dafnapension:
a simple fix..  perhaps good enough for now.

### DIFF
--- a/src/unitxt/loaders.py
+++ b/src/unitxt/loaders.py
@@ -39,12 +39,18 @@ class LoadHF(Loader):
     cached = False
 
     def process(self):
-        dataset = hf_load_dataset(
-            self.path, name=self.name, data_dir=self.data_dir, data_files=self.data_files, streaming=self.streaming
-        )
+        try:
+            dataset = hf_load_dataset(
+                self.path, name=self.name, data_dir=self.data_dir, data_files=self.data_files, streaming=self.streaming
+            )
+        except NotImplementedError:
+            dataset = hf_load_dataset(
+                self.path, name=self.name, data_dir=self.data_dir, data_files=self.data_files, streaming=False
+            )
+            for split in dataset.keys():
+                dataset[split] = dataset[split].to_iterable_dataset()
 
         return MultiStream.from_iterables(dataset)
-
 
 class LoadCSV(Loader):
     files: Dict[str, str]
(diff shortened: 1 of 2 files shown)


perlitz:
Hi @dafnapension!
nice to git-meet :) 
you will need to run `pre-commit install` inside the project to make the formatters work when you commit, this way pass the you will pass the Code Quality & Security / verify-pre-commit-executed (pull_request) tests

src/unitxt/loaders.py
```diff
@@ -39,13 +39,19 @@ class LoadHF(Loader):
     cached = False
 
     def process(self):
-        dataset = hf_load_dataset(
-            self.path, name=self.name, data_dir=self.data_dir, data_files=self.data_files, streaming=self.streaming
-        )
+        try:
```
perlitz, line 42: can't you just LoadHF with streaming=False?
and

if not streaming:
    for split in dataset.keys():
        dataset[split] = dataset[split].to_iterable_dataset()

?

perlitz:
Also, you will have to add test to improve the coverage

dafnapension:
Thank you, @perlitz ,
Can you please point for me, novice, to the unittest that covers **any** line from LoadHF?

src/unitxt/loaders.py
yoavkatz: I think you can do as you suggest:
if (self.streaming):
    dataset = hf_load_dataset(
                self.path, name=self.name, data_dir=self.data_dir, data_files=self.data_files, streaming=self.streaming
            )
else:
      dataset = hf_load_dataset(
                self.path, name=self.name, data_dir=self.data_dir, data_files=self.data_files, streaming=False
            )
            for split in dataset.keys():
                dataset[split] = dataset[split].to_iterable_dataset()
                
Then the card creator , can have specify  

loader = LoadHF(path="GEM/xlsum", name="igbo", streaming=False) 


If the error message when stream = True and reading zip files is not clear, you can catch the exception, and raise a new one. 


In any case, best to have a test that checks that the exceptions is generated.

yoavkatz (changes requested):
Hi Dafna, please see the conversation.

src/unitxt/loaders.py
dafnapension: Thank you, @yoavkatz , I do cover the exception catch in the test. I see these lines are all green in the coverage report.

dafnapension: @yoavkatz , can you, then, approve this PR?

yoavkatz (approved):

yoavkatz:
Closing to add sign to solve DCO issue.

PR #290, opened 2023-11-08
add a card preparer for dataset xlsum

dafnapension:


dafnapension:
this script can run all the way through only after pull request 288 is merged

prepare/cards/xlsum.py
```diff
@@ -0,0 +1,38 @@
...
+            metrics=["metrics.rouge"],
+        ),
+        templates=TemplatesList(
+            [
+                InputOutputTemplate(input_format="{text}", output_format="{target}"),
+            ]
+        ),
+    )
+    if lang == langs[0]:
+        test_card(card, debug=True)
```
yoavkatz, line 37: set debug = False by default to speed up regressions.

prepare/cards/xlsum.py
```diff
@@ -0,0 +1,38 @@
...
+langs = configs
+
+for lang in langs:
+    card = TaskCard(
+        loader=LoadHF(path="GEM/xlsum", name=lang),
+        preprocess_steps=[
+            # SplitRandomMix({"train": "train[95%]", "validation": "train[5%]", "test": "validation"}),
+            # CopyFields(field_to_field=[["answers/text", "answer"]], use_query=True),
+        ],
+        task=FormTask(
```
yoavkatz, line 25: In general, we should move to a summarization task that is defined outside the card.  We can setup a meeting to discuss.

yoavkatz (changes requested):
Changes in the comments.

dafnapension:
open  PR #301 instead, and recommitted  with signature

PR #305, opened 2023-11-13
nDCG metric (ranking quality)

arielge:


prepare/metrics/ndcg.py
```diff
@@ -0,0 +1,40 @@
...
+from src.unitxt.metrics import NDCG, MetricPipeline
+from src.unitxt.test_utils.metrics import test_metric
+
+# Normalized Discounted Cumulative Gain
+metric = MetricPipeline(
+    main_score="nDCG",
+    preprocess_steps=[
+        CopyFields(field_to_field=[("references/0", "references")], use_query=True),
+        CastFields(
+            fields={"prediction": "float", "references": "float"},
```
yoavkatz, line 12: I commend you for the use of operators to define the metrics, but in this case, I'm not sure .  
Wouldn't it be more readable, if the code was just in the metric, with clearer error message (e.g. to check that the there is only one reference).

yoavkatz (changes requested):
See comments below.

prepare/metrics/ndcg.py
arielge: Thanks, I am not yet so familiar with all the moving parts in unitxt, so I do not have a strong opinion on the matter.
In this case I followed the example from here - https://github.com/IBM/unitxt/blob/main/prepare/metrics/spearman.py

yoavkatz: Yes. The difference is that  in spearman the pipeline reuses an existing metric wrapper (HuggingfaceMetric), here you are creating a new metric in any case. 

But I don't have a strong opinion.   I'll approve.

yoavkatz (approved):

PR #307, opened 2023-11-14
Adjust F1 to yes no template

matanor:
The [YesNoTemplate](https://github.com/IBM/unitxt/blob/main/src/unitxt/templates.py#L184) produces outputs of "Yes" and "No" for binary classification (for a question of 'does some text belong to some class?'). These correspond to 'class name' and 'none' output for existing binary templates (e.g., asking 'classify some text as class name or none').
The special handling of 'none' gold labels within the [F1MultiLabel](https://github.com/IBM/unitxt/blob/main/src/unitxt/metrics.py#L392) metric should also apply to 'no' labels. 
This pull request makes this adjustment, and also updates the relevant tests.


yoavkatz (approved):

matanor:
Decided not to commit, but rather adjust the YesNoTemplate to the metric. Instead of adjusting the metric to the template.

PR #310, opened 2023-11-14
almostEvil first attempt

dafnapension:


dafnapension:
Hi @yoavkatz ,
This card preparer is somewhat different: question answer type, rather than multi-choice like amazon, or summarizers like xlsum and mlsum that you already reviewed for me.
Also, in here, I had to do the split by myself, because it only has train.

gitMichal:
thanks @dafnapension , i wonder if the metric here should not be ROUGE rather than squad ?

dafnapension:
Thanks, @gitMichal , I agree. I change to rouge.

yoavkatz (approved):

PR #311, opened 2023-11-15
Documentation update

yoavkatz:


docs/concepts.rst
```diff
@@ -20,7 +20,7 @@ Operator
 
 An operator is a class that takes multiple streams as input and produces multiple streams as output.
 Every modification of the data in the stream is done by an operator.
-Every opertor should be doing a single task and its name should reflect its operation.
+Every opertor should perform single task and its name should reflect its operation.
```
matanor, line 23:
```suggestion
Every opertor should perform a single task and its name should reflect its operation.
```

docs/introduction.rst
```diff
@@ -7,6 +7,11 @@
 Introduction
 ===================
 
-Unitxt is a Python library for getting data prepared for utilization by language models. In one line of code, it prepares a dataset or mixtures-of-datasets into an sequence to sequence input-output format for training and evaluation. Our aspiration is to be simple, adaptable, and transparent.
+Unitxt is a Python library for getting data ready for utilization in training, evaluation and inference of language models.
+It provides a set of reusable building blocks and methodology for defining datasets and metrics.
 
-Unitxt builds on the principle of separation. Separation allows adding a dataset without knowing anything about the models using it. It allows for training without worrying about preprocessing, switching models without the need to load the data differently, and changing formats (instruction, in-context learning, etc.) without changing anything else.
+In one line of code, it prepares a dataset or mixtures-of-datasets into an sequence to sequence input-output format for training and evaluation. Our aspiration is to be simple, adaptable, and transparent.
```
matanor, line 13:
```suggestion
In one line of code, it prepares a dataset or mixtures-of-datasets into a sequence to sequence input-output format for training and evaluation. Our aspiration is to be simple, adaptable, and transparent.
```

docs/no_installation_usage.rst
```diff
@@ -1,25 +1,76 @@
...
+    })
+  })
+
+This prints the source text (input to the model) of the first sample in the training set:
+
+.. code-block:: python
+
+    print(dataset['train'][0]['source'])
+
+    'Given this sentence: I stuck a pin through a carrot. When I pulled the pin out, it had a hole., classify if this sentence: The carrot had a hole. is entailment, not entailment.\n'
```
matanor, line 41: a general question: i wonder if the input text in the templates should be within two " characters. For the above example, it seems lacking. Though maybe LLMs don't care about that.

docs/no_installation_usage.rst
```diff
@@ -1,25 +1,76 @@
...
 
 Unitxt enable the loading of datasets in different forms if they have a dataset card stored in
-the unitxt catalog. For example, here we load wnli in 5 shots format:
+the unitxt catalog. For example, here we load wnli in 3 shots format:
+
+.. code-block:: python
+
+  dataset = load_dataset('unitxt/data', 'card=cards.wnli,num_demos=3,demos_pool_size=100')
+
+Now the source source text (input to the model) of the first sample in the training set has in-context examples:
```
matanor, line 61:
```suggestion
Now the source text (input to the model) of the first sample in the training set has in-context examples:
```

matanor (approved):
LGTM, found 2-3 typos.
I did not go over `catalog.rst`. Should I? Or is it auto generated from the catalog?

docs/concepts.rst
```diff
@@ -2,7 +2,7 @@
 Concepts
 ==============
 
-The main building blocks of the library are the `Stream`, `Operator` and the `Atifact` classes.
+The main building blocks of the library are the `Stream`, `Operator` and the `Artifact` classes.
```
dafnapension, line 5: down here, when detailing Stream, it does not look a class, but an abstract term.

docs/no_installation_usage.rst
yoavkatz: We should improve the standard template for nil and then update this.

docs/concepts.rst
yoavkatz: This page will go through a bit revision.

Issue #297, opened 2023-11-12
New operator to return most common values from a given field


yoavkatz:
Often we need to list a set of choices for the model, which can be derived from the labels of train data.

Today, people need to write them manually, use metadata (if available) or find some other way to generate them.

The idea is to add a new StreamOperators which returns field values

ExtractFieldValues(stream="train", field="label",  to_field="choices",  min_precent (optional)=1 , min_count (optional)= 10)

- Fetch all instance from stream ("train")
- Count distribution of values of field ("label")
- Remove from dictionary all labels  below min _precent or below  min_count
- Sort the list of value  by count
- Add the list field named **to_field** to all streams in input multi stream  ("choices"). [ "negative" , "positive"] 

Before:
train:
    { "text" : ... , "label: positive}
    { "text" : ... , "label: positive}
    { "text" : ... , "label: neutral}
    { "text" : ... , "label: neutral}
    { "text" : ... , "label: negative}
    
test:
    { "text" : ... , "label: positive}
    { "text" : ... , "label: negative}
    
After:
    
train:
    { "text" : ... , "label: positive" , choices : ["positive","neutral"}
    { "text" : ... , "label: positive,   choices : ["positive","neutral"}}
    { "text" : ... , "label: negative,  choices : ["positive","neutral"}}
    
test:
    { "text" : ... , "label: positive,  choices : ["positive","neutral"}}
    { "text" : ... , "label: negative, choices : ["positive","neutral"}}
    
Similar to merge (MultiStreamOperator)

yoavkatz:
Implemented. Thanks Dafna.

PR #316, opened 2023-11-16
added new MultiStream operator: ExtractFieldValues

dafnapension:
… FilterByCondsOnValues
per issue 297, and a demand from the field..


dafnapension:
Hi @yoavkatz , I added, at last, the Class that we talked about and you summarized in issue #297.
this is FindAndStoreMostCommonValues.
'on the way' I added yet another Class that extends FilterByValue to a more general condition for filtering: FilterByCondsOnValues.  A need for this came from some application where the language of the document is hiding inside a field, rather than being the whole field, and yet, we need to filter by that language.
The tests show expected results for some simple examples.
please review.

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    Then, for all streams of the multistream, for all rows of each stream, add a new field, named
+    'new_field' whose value will be the sored list of the remaining values.
+    (that is each and every row in the multistream will carry that very same field-value pair.)
+    """
+
+    def process(self, multi_stream: MultiStream) -> MultiStream:
+        stream = multi_stream[self.stream_name]
+        all_values = []
+        for instance in stream:
+            all_values.append(instance[self.field_name])
```
yoavkatz, line 722: Need to support list elements

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    stream_name: Optional[str] = "train"
+    top_percent: Optional[int] = 80
+    min_relative_freq_percent: Optional[int] = 5
+    new_field: Optional[str] = None
+    """
+    Given a MultiStream, a stream_name, and a field name, and two margins of commonality.
+    Do nothing (exception thrown) if the multistream does not contain a stream named stream_name, or if the multistream does not
+    contain a field (feature) named field_name.
+    Count the frequency of each value, over the stream named stream_name. Sort the values by frequency.
+    Discard each value that is not at the top_percent of that sorted list.
```
yoavkatz, line 711: Not sure if there is a reason to take first 80% of the list. One can always cut it later. Maybe I'm missing something.

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+        all_values = []
+        for instance in stream:
+            all_values.append(instance[self.field_name])
+        counter = Counter(all_values)
+        how_many_in_top = math.ceil(self.top_percent * len(counter.most_common()) / 100)
+        top_values = counter.most_common(how_many_in_top)
+        min_frequency = math.floor(self.min_relative_freq_percent * len(all_values) / 100)
+        while top_values[-1][1] < min_frequency:
+            top_values.pop()
+        values_to_remain = [ele[0] for ele in top_values]
```
yoavkatz, line 729: values_to_keep?

yoavkatz (changes requested):
Thanks Dafna - added comments in the PR.  If you want to can also talk. Not sure all my comments were clear.

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
                 yield instance
 
 
+class FilterByCondsOnValues(SingleStreamOperator):
```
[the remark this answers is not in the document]
dafnapension: And where will the value computed in the first stage be kept for the second stage? extend the record with yet another field?

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+                    )
+                if not filtlambda(instance[key]):
+                    filter = True
+            if not filter:
+                yield instance
+
+
+class FindAndStoreMostCommonValues(MultiStreamOperator):
+    field_name: str
+    stream_name: Optional[str] = "train"
```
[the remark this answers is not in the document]
dafnapension: I accept.

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    top_percent: Optional[int] = 80
```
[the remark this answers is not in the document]
dafnapension: I accept

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    min_relative_freq_percent: Optional[int] = 5
+    new_field: Optional[str] = None
+    """
+    Given a MultiStream, a stream_name, and a field name, and two margins of commonality.
```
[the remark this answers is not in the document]
dafnapension: Very nice. I accept. I now understand what you had in mind: both measures deal with the relative frequency of the unique values (this, I now understand, is what you had in mind) and not their relative rank in the sorted list of relative frequencies (as I first implemented).

src/unitxt/operators.py
dafnapension: As I wrote above: not sure if to support them individually: count the frequency of each individual, and then which records are to stay? those whose lists consist of all frequent values? whose lists contain at least one frequent value? 
Or rather: measure frequencies of whole lists, and then the same rule used for individual values would apply for whole lists, and the issue is just to allow, programmatically, lists of str, and not only   str.

src/unitxt/operators.py
dafnapension: I accept. I now understand what you had in mind: measure only by frequency of coverage.

src/unitxt/operators.py
dafnapension: yes, better. Thanks!

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+                    raise ValueError(
+                        f"Required filter field ('{key}') in FilterByCondsOnValues is not found in {instance}"
+                    )
+                if not filtlambda(instance[key]):
+                    filter = True
+            if not filter:
+                yield instance
+
+
+class FindAndStoreMostCommonValues(MultiStreamOperator):
```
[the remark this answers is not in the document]
dafnapension: Yes, better. I accept.

src/unitxt/operators.py
```diff
@@ -671,6 +672,69 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    field_name: str
+    stream_name: Optional[str] = "train"
+    top_percent: Optional[int] = 80
+    min_relative_freq_percent: Optional[int] = 5
+    new_field: Optional[str] = None
+    """
+    Given a MultiStream, a stream_name, and a field name, and two margins of commonality.
+    Do nothing (exception thrown) if the multistream does not contain a stream named stream_name, or if the multistream does not
```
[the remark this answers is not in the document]
dafnapension: I replaced my description with your much nicer one.  Thanks!

dafnapension:
Thank you very much, @yoavkatz , for your thorough review. I got it all, and replied above accordingly, 
except two things:  
(1) I did not understand the semantics of  `process_every_value=True` in `ExtractFieldValues`,
(2) And was not sure where to write the computed value, from the first stage that you suggested, of `FilterByCondsOnValues`.

dafnapension:
Thank you very much, @yoavkatz , for your thorough review. I implemented, and ran tests, all your comments above (which I also replied to), except two things:
(1) I did not understand the semantics of `process_every_value=True` in `ExtractFieldValues`,
(2) And I was not sure where to write the computed value, from the first stage that you suggested, of `FilterByCondsOnValues`.

dafnapension:
Hi @yoavkatz , 
I think I got it my question (1) above:  here is how I am going to implement:
When the value of 'field' is list, and not a single value:
if `process_every_value=False`,  I treat the list as one element, and like before, find the most popular lists, counting the number of occurrences of each and every unique list in column 'field'. At the end, I also report in the 'to_field'  -- a list of lists, being the most popular lists.
if `process_every_value=True`, I count the occurrences of each and every individual element in that list.  e.g., the list ['a', 'b', 'c'], showing in field 'field' of a row in 'train' (say), contributes 1 to the number of occurrences of 'a',  1 to the number of occurrences of 'b' and 1 to the number of occurrences of 'c'.  At the end I report the most popular individual **values**.

Later, down the stream of implementation, people will do with the values they find in field 'to_field' -- as they want. Either filter out any row hose list contains no popular element, or filter out any row whose list is not wholly made of popular elements.

dafnapension:
Hi @yoavkatz ,
I implemented as I understood and described above.
If this is in the direction of what you had in mind - then I am only left with the question: where to write the value that you suggested [here](https://github.com/IBM/unitxt/pull/316#discussion_r1398409893).

src/unitxt/operators.py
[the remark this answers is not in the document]
yoavkatz: Apply can receive any function, but I think it needs to be named function. It stores the results in a new field or the same field.

I'll update the doc, but here is a copy:

  """A class used to apply a python function and store the result in a field.

    Args:
        function (str): name of function. 
        to_field (str): the field to store the result
        additional arguments are field names passed to the function

    Examples:

    Store in field  "b" the uppercase string of the value in field "a"
    Apply("a", function=str.upper, to_field="b")   

    Dump the json representation of field "t" and store back in the same field.
    Apply("t", function=json.dumps, to_field="t")

    Set the time in a field 'b'.
    Apply(function=time.time, to_field="b")

tests/test_operators.py
```diff
@@ -375,6 +399,336 @@ def test_merge(self):
         expected_merged = [{"field": "test1"}, {"field": "train1"}]
         self.compare_streams(merged, expected_merged)
 
+    def test_extract_values(self):
+        input_multi_stream = MultiStream(
+            {
+                "test": [{"field": "test1"}],
+                "validation": [{"field": "validation1"}],
+                "train": [
+                    {"field": "train1"},
```
yoavkatz, line 408: I suggeste make the test more interesting (and test more of the behavior). Add 10 examples with more meaningful names, not in order of frequency from the start.

fish
dog
dog
cat
dog
cat
sheep
cat
fish
shark





tain

src/unitxt/operators.py
```diff
@@ -671,6 +674,100 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+                sum_counts += p[1]
+                if sum_counts >= top_frequency:
+                    break
+            values_and_counts = counter.most_common(i + 1)
+        if self.min_frequency_percent > 0:
+            min_frequency = math.floor(self.min_frequency_percent * len(all_values) / 100)
+            while values_and_counts[-1][1] < min_frequency:
+                values_and_counts.pop()
+        values_to_keep = [[*ele[0]] if isinstance(ele[0], tuple) else ele[0] for ele in values_and_counts]
+        if self.to_field is None:
```
yoavkatz, line 763: I suggest we just throw an error if to_field is not set.  It's more consistent with other operators.

src/unitxt/operators.py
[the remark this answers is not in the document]
dafnapension: Thanks, @yoavkatz ,
If the argument to the passed function also needs to be passed upon creation of Apply, then you need to create one Apply object per record, and not one Apply per stream?

tests/test_operators.py
dafnapension: Good idea,  will do.

src/unitxt/operators.py
dafnapension: I accept.

dafnapension:
Thank you, @yoavkatz , for your very good suggestions. I will implement them.

dafnapension:
Hi @yoavkatz, I implemented changes that you asked.
Thank you for your thorough and educating review.

dafnapension:
Hi @yoavkatz ,
My joy about Apply was too early.. I think the method is bugged.
As you can see in PR https://github.com/IBM/unitxt/pull/321, I tried to use it with a simple function I defined in the module almostEvilML_qa_by_lang.  That function is translated to a string by Apply, and stored in the cards that show in src/catalog/cards/almostEvilML_qa_by_lang.
However, the translation to string is miserable and erroneous, and hence on loading time, the function can not be restored, and hence the tests fail when I push.

This is irrelevant to this PR, but since we talked about Apply in this PR, I am writing this comment..

yoavkatz:
> Hi @yoavkatz , My joy about Apply was too early.. I think the method is bugged. As you can see in PR #321, I tried to use it with a simple function I defined in the module `almostEvilML_qa_by_lang`. That function is translated to a string , `"__main__.extract_lang_from_METADATA"`, by Apply, and stored in the cards that show in src/catalog/cards/almostEvilML_qa_by_lang.
> 
> It seems, though, that global lists, `__builtins__` and `globals()` are not updated as necessary, and the restoring of the function from its string format - fails.
> 
> This is irrelevant to this PR, but since we talked about Apply in this PR, I am writing this comment here

Right.   We can not use functions which are not in the main code base of unitxt or fm-eval, because it will have no way to find them in the prepare cards.

src/unitxt/operators.py
```diff
@@ -677,6 +680,74 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+        values_and_counts = counter.most_common()
+        if self.overall_top_frequency_percent < 100:
+            top_frequency = math.ceil(len(all_values) * self.overall_top_frequency_percent / 100)
+            sum_counts = 0
+            for i, p in enumerate(values_and_counts):
+                sum_counts += p[1]
+                if sum_counts >= top_frequency:
+                    break
+            values_and_counts = counter.most_common(i + 1)
+        if self.min_frequency_percent > 0:
```
yoavkatz, line 740: Thinking about this more.  I think we should not support both top_frequency_percent and min frequency_precent together, since if you take min_frequency=20, you might not be able to get top_frequency_percent.  

I suggest that for now, we check this and throw an error if both are set.  
  This can be done in the verify method, which is called before prepare.

src/unitxt/operators.py
```diff
@@ -677,6 +680,74 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+    ExtractFieldValues(stream_name="train", field="label", to_field="classes",min_frequency_percent=5) -
+    extracts all possible values of field 'label' that cover, each, at least 5% of the instances.
+    Stores these values, sorted by decreasing order of frequency, in field 'classes' of each record in all streams.
+    """
+
+    def process(self, multi_stream: MultiStream) -> MultiStream:
+        stream = multi_stream[self.stream_name]
+        all_values = []
+        for instance in stream:
+            if (not isinstance(instance[self.field], list)) or (self.process_every_value == False):
```
yoavkatz, line 724: IS this the logic?
  if (self.process_every_value):
     if (not isinstance(instance[self.field], list)) 
         raise ValueError(...)
     all_values.extend(instance[self.field])
  else
      all_values.append(instance[self.field])
      
      What Am I missing?  Why do this " (*instance[self.field],) if isinstance(instance[self.field]," for lists?

src/unitxt/operators.py
yoavkatz: We should also verify both numbers are between 0 and 100.

tests/test_operators.py
```diff
@@ -375,6 +377,282 @@ def test_merge(self):
...
+            + json.dumps(expected_output2)
+            + "\n but instead, received: \n"
+            + json.dumps(output_multi_stream),
+        )
+        # with lists, treated as single elements
+        input_multi_stream = MultiStream(
+            {
+                "test": [{"field": ["a", "b", "c"]}],
+                "validation": [{"field": ["d", "e", "f"]}],
+                "train": [
```
yoavkatz, line 463:
```suggestion

 # The list has 
 # h: 6
  # m: 6
 # j: 3
 # i: 3 
 # k: 3
 # o: 3 
 # p: 3 
  # q: 2
 # r: 2
 # s: 2
 # t: 1
 # u:1 
 # v:1
 
# Tuples:

# ["h", "I", "j"] : 3
# ["k", "h", "m"]: 3
# ["m", "n", "o"] : 3
 # ["q", "r", "s"] : 2
 # ["t","u","v"] : 1

  
                "train": [
```

src/unitxt/operators.py
yoavkatz, line 724: I think that if we process_every_value = True and the value is not a list, an error should be generated.

src/unitxt/operators.py
```diff
@@ -677,6 +680,74 @@ def process(self, stream: Stream, stream_name: str = None) -> Generator:
...
+                all_values.append(
```
yoavkatz, line 725: Add a comment why we convert the list to a tuple (took me a while to understand).

I think this is the logic:
1. If process_every_value and value is not a list, throw an error.
2. If not process_every_value and is not a list, append element to the all_values list.
3. If not process_every_value and list,  convert to tuple, and append tuple to all_values list.
4. If process_every_value and list, add each element in list to the all_values list.

yoavkatz (changes requested):
Thank Dafna.  I had another review - hopefully the final one , with a few comments, mostly on documentation and error checking.

src/unitxt/operators.py
dafnapension: Done. Very nice catch!

src/unitxt/operators.py
dafnapension: Thanks! indeed important!  Done.

src/unitxt/operators.py
dafnapension: Exactly! conversion is done only to allow the simple use of Counter, which would not accept lists hashables - for its entries.

dafnapension:
Thank you, @yoavkatz , I addressed your comments. Please do not hesitate to further review and write comments. They help me improve my contribution, and I learn a lot from them.

src/unitxt/operators.py
```diff
@@ -695,39 +695,62 @@ class ExtractFieldValues(MultiStreamOperator):
...
     ExtractFieldValues(stream_name="train", field="labels", to_field="classes", process_every_value=True) -
     in case that field 'labels' contains a list of values (and not a single value) - track the occurrences of all the possible
-    values in these lists, and report the most frequent values.
+    value members in these lists, and report the most frequent values.
     if process_every_value=False, track the most frequent whole lists, and report those (as a list of lists) in field 'to_field'
 
     ExtractFieldValues(stream_name="train", field="label", to_field="classes",overall_top_frequency_percent=80) -
-    extracts the most frequent possible values of field 'label' that cover at least 80% of the samples,
-    and stores them in field 'classes' of all streams.
+    extracts the most frequent possible values of field 'label' that cover at least 80% of the rows of stream_name,
```
yoavkatz, line 714:
```suggestion
    extracts the most frequent possible values of field 'label' that cover at least 80% of the instances of stream_name,
```

src/unitxt/operators.py
```diff
@@ -695,39 +695,62 @@ class ExtractFieldValues(MultiStreamOperator):
...
+    and stores them in field 'classes' of each record of all streams.
```
yoavkatz, line 715:
```suggestion
    and stores them in field 'classes' of each instance in all the streams.
```

src/unitxt/operators.py
```diff
@@ -695,39 +695,62 @@ class ExtractFieldValues(MultiStreamOperator):
...
+    def verify(self):
+        assert (
+            self.overall_top_frequency_percent <= 100 and self.overall_top_frequency_percent >= 0
+        ), "'overall_top_frequency_percent' must be between 0 and 100"
+        assert (
+            self.min_frequency_percent <= 100 and self.min_frequency_percent >= 0
+        ), "'min_frequency_percent' must be between 0 and 100"
+        assert not (
+            self.overall_top_frequency_percent < 100 and self.min_frequency_percent > 0
+        ), "At most one of 'overall_top_frequency_percent' and 'min_frequency_percent' is allowed to move from their default value"
```
yoavkatz, line 731:
```suggestion
        ), "At most one of 'overall_top_frequency_percent' and 'min_frequency_percent' is allowed to be set "
```

yoavkatz (approved):
Minor two suggestions  regarding teminology. We have "streams" that contain "instances" (not rows or records)

dafnapension:
Hi @yoavkatz ,
I thought about your question regarding the complex condition around what to append/extend to all_values in `process(..)`, and thought that it is not too efficient to check these for each and every instance, as all the instances have the same 'answers' to these condition.
So I moved these condition-evaluation to out of the loop.
Could you please review this as well?  (pardonez moi)
(Like in the old joke of my grandmother: when at last the old maid is invited to a dance, the little orchestra goes out for a pee recess..)

yoavkatz:
> Hi @yoavkatz , I thought about your question regarding the complex condition around what to append/extend to all_values in `process(..)`, and thought that it is not too efficient to check these for each and every instance, as all the instances have the same 'answers' to these condition. So I moved these condition-evaluation to out of the loop. Could you please review this as well? (pardonez moi) (Like in the old joke of my grandmother: when at last the old maid is invited to a dance, the little orchestra goes out for a pee recess..)

I think what was before is safer, because it could be that fields are of different values (likely to be a bug, but still better to identify it).  Also in terms of runtime - we iterato over the whole stream, so adding another check is minimal cost.

dafnapension:
Thank you @yoavkatz , nice catch. Did not occur to me.   I am the little orchestra (KleiZmer) that went out to a pee recess.  Returning to the dancing hall right away.

PR #321, opened 2023-11-20
process almost evil lang by lang

dafnapension:
using StreamOperator  Apply


dafnapension:
Hi @gitMichal and @yoavkatz ,
This automatic test run fails, because the generation of the function that is passed to Apply fails. It seems that finding the containing module - almostEvil_by_lang - fails. When I only run this almostEvil_by_lang, which ends with test_card, the run completes successfully, as almostEvil_by_lang is found in the list of imports. But when test_card is invoked from other module -- almostEvil_by_lang is found no where, when Apply looks for instantiating it from the string read from the card.

Please advise..

prepare/cards/almostEvilML_qa_by_lang.py
```diff
@@ -0,0 +1,38 @@
...
+    md_as_json = json.loads(md_as_string)
+    return md_as_json["language"]
+
+
+for lang in langs:
+
+    card = TaskCard(
+        loader=LoadHF(path="0x22almostEvil/multilingual-wikihow-qa-16k"),
+        preprocess_steps=[
+            Apply("METADATA", function=extract_lang_from_METADATA, to_field="extracted_language"),
```
yoavkatz, line 22: I think this should work:
Apply("METADATA", function=json.load, to_field="metadata")
CopyFields(field_to_field=[("metadata/language", "extracted_language")], use_query=True),

yoavkatz: Note the dateset has a legal disclaimer, so we will not want to use it, until approved.  At the minimum, if we add the card, we should add a link to the disclaimer.

dafnapension: Thank you so much, @yoavkatz , for yet another lesson I learned from you!
Indeed all automatic tests now pass !!

what disclaimer did you have in mind?  this one?
https://www.wikihow.com/wikiHow:Creative-Commons

and how to add it to the card? manually?

dafnapension: Hi @yoavkatz , Thanks again for the Apply trick and education. Indeed all automatic tests now pass !!

Please advise re. the disclaimer:
(1) What disclaimer did you have in mind? this one? https://www.wikihow.com/wikiHow:Creative-Commons
(2) and how to add it to the card? manually?

yoavkatz (approved):

PR #323, opened 2023-11-21
Added error message when failed to process

yoavkatz:
value in instrance operator and augmentors

### DIFF
--- a/src/unitxt/test_utils/operators.py
+++ b/src/unitxt/test_utils/operators.py
@@ -33,7 +33,7 @@ def test_operator_exception(
         apply_operator(operator, inputs)
     except Exception as e:
         if tester is not None:
-            tester.assertEquals(str(e), exception_text)
+            tester.assertEqual(str(e), exception_text)
         elif str(e) != exception_text:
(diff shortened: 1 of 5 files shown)


matanor (changes requested):
please see  a few comments

src/unitxt/test_utils/operators.py
```diff
@@ -21,8 +21,33 @@ def apply_operator(operator: StreamingOperator, inputs: List[dict], return_multi
...
+
+
 def test_operator(
-    operator: StreamingOperator, inputs: List[dict], targets: List[dict], tester=None, sort_outputs_by=None
+    operator: StreamingOperator,
+    inputs: List[dict],
+    targets: List[dict],
+    tester=None,
+    sort_outputs_by=None,
+    exception_text=None,
```
matanor, line 50: why is exception_text added? i don't see its used?

matanor (approved):
LGTM, left one small comment

src/unitxt/test_utils/operators.py
yoavkatz: I'll remove.  It was added by mistake.

PR #324, opened 2023-11-21
Added initial implementation of  Loader that loads existing cards

yoavkatz:
that allows using output of other cards in new cards


yoavkatz:
Not sure there is a strong need now.  Closing for now.

PR #327, opened 2023-11-22
Added option to add instruction only before last user turn

yoavkatz:


gitMichal:
1. please remove the print(source)
2. also maybe a better naming should be used?

yoavkatz:
Matan proposed a much better way to handle formatting by providing the textual format with placeholders.
If we want to approve this in the intern .

Change to 'add_instruction_after_demos' and added tests.

gitMichal (approved):

PR #328, opened 2023-11-22
A new format that does not have an input output seperator

matanor:
This is required for inference in 0-shot mode, when it is expected that the model completes the input, and does not answer in a new line.


yoavkatz (approved):

matanor:
> Do we want want to remove the space that is added?
> 
> The default: target_prefix: str = " "

Not sure, i think it does not affect the current problem, since the translation is 0-shot.

I can remove however, in that case, i should probably rename the format, no?
Also, if removing, what about demo_separator? that one should be kept, i think.

PR #330, opened 2023-11-22
Better control of input output seperator

yoavkatz:
Each template now defines the input output seperator.

This allows some templates to avoid adding newline between the input and outputs.

Also removed target fix which unneeded, and
by default added a space.


yoavkatz:
Closing this one.  We will implement a different behavior.

yoavkatz:
Closing as we plan a better solution.

Issue #326, opened 2023-11-22
Add a ability to map specific values in List


yoavkatz:
Currently MapInstanceValues - maps complete valus in fields

MapInstanceValues(mappers={"labels": {"[]": ["none"]}}, strict=False),

However, in multi label classification, we would like to map individual elements within a field.

We would like an option 'process_every_value' to be added to MapInstanceValues to allow working on list.

MapInstanceValues(mappers={"labels": {"a": "apple", "b": "bandana"}}, strict=False, process_every_value=True),

dafnapension:
Hi @yoavkatz , please see [pull request](https://github.com/IBM/unitxt/pull/331) that addresses this issue.

PR #331, opened 2023-11-22
first attempt for issue #326

dafnapension:


yoavkatz (commented):
Why do we need a new Operator, can't we extend the existing MapIstanceValues operator, to just have the "process_every_value" field?

src/unitxt/operators.py
```diff
@@ -100,6 +100,72 @@ def process(self, instance: Dict[str, Any], stream_name: str = None) -> Dict[str
...
+        strict (bool): If True, the mapping is applied strictly. That means if a value
+            does not exist in the mapper, it will raise a KeyError. If False, values
+            that are not present in the mapper are kept as they are.
+        process_every_value (bool): Let 'm' denote the mapper associated with field 'f', and let 'v' be a key in 'm'.
+            When process_every_value=False: apply the mapping only if field 'f' content equals 'v'. In that case, replace
+            'v' by 'm[v]'.
+            When process_every_value=True: apply also if the content of 'f' is a list, one of whose members is 'v'.
+            In such a case, replace just member 'v' in the list, by 'm[v]', leaving the other members of the list as they are
+            (potentially, to be replaced by other values of 'm').
+    """
```
yoavkatz, line 122: I think the most important point is to add examples.

src/unitxt/operators.py
```diff
@@ -100,6 +100,72 @@ def process(self, instance: Dict[str, Any], stream_name: str = None) -> Dict[str
...
+    it maps values of instances in a stream using predefined mappers.
+
+    Attributes:
+        mappers (Dict[str, Dict[str, str]]): The mappers to use for mapping instance values.
+            Keys are the names of the fields to be mapped, and values are dictionaries
+            that define the mapping from old values to new values.
+        strict (bool): If True, the mapping is applied strictly. That means if a value
+            does not exist in the mapper, it will raise a KeyError. If False, values
+            that are not present in the mapper are kept as they are.
+        process_every_value (bool): Let 'm' denote the mapper associated with field 'f', and let 'v' be a key in 'm'.
```
yoavkatz, line 116: Better to avoid formal notations and simplify for the user.  .  All this means, is that if prices_every_value =True is that, each field mist be a  list, and the mapping is done to individual elements in the list.

dafnapension:
Hi @yoavkatz ,
You are absolutely right in all your comments. 
Such an elegant solution of just adding a new parameter did not occur to me at all..
Simplified now implementation and documentation. Added examples.

src/unitxt/operators.py
```diff
@@ -71,32 +71,64 @@ class MapInstanceValues(StreamInstanceOperator):
         strict (bool): If True, the mapping is applied strictly. That means if a value
             does not exist in the mapper, it will raise a KeyError. If False, values
             that are not present in the mapper are kept as they are.
+        process_every_value (bool): If True, all fields to be mapped should be lists, and the mapping
+            is to be applied to their individual elements. If False, mapping is only applied to a field
+            containing a single value.
+
+    Examples: MapInstanceValues with mappers={"a": {"1": "hi", "2": "bye"}}, and process_every_value = True,
```
yoavkatz, line 78: @dafnapension @elronbandel - I think we should agree on the example format for all operators.  I think we should aim that each example will be self contained, and one could cut and past it.  Fro example:

Replace '1' with 'hi' and '2' with 'bye' in field 'a' of all streams:
`MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}})`
So {"a":1 , "b": 2} becomes "a":"hi" , "b": 2} 

Assuming 'a' is field with a list of values, replace  all 1s with 'hi' and all 2s with 'bye'
`MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}, process_every_element=True`
 So {"a": [1, 2], "b": 2} becomes {"a": ["hi", "bye"], "b": 2}.

 Ensure all values are mapped in the instance
 `MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}, strict=True)`
 So if the input is  {"a":3 , "b": 2} it will throw an exception, since the mapping of 3 in not in the mapper.

tests/test_operators.py
```diff
@@ -45,23 +46,86 @@ def compare_streams(self, all, expected_all):
             self.assertDictEqual(input_dict, output_dict)
 
     def test_map_instance_values(self):
+        mappers = {"a": {"1": "hi", "2": "bye"}}
+
         inputs = [
-            {"a": 1, "b": 2},
-            {"a": 2, "b": 3},
+            {"a": "1", "b": "2"},
```
yoavkatz, line 52: I thinking keeping the input  as numbers is a more realistic example.

tests/test_operators.py
```diff
@@ -45,23 +46,86 @@ def compare_streams(self, all, expected_all):
...
         check_operator(
-            operator=MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}),
+            operator=MapInstanceValues(mappers=mappers),
             inputs=inputs,
             targets=targets,
             tester=self,
         )
 
+        # process_every_value=True would not accept non-list inputs
+        with self.assertRaises(ValueError):
```
yoavkatz, line 70: Please check exception with the new `check_operator_exception` that also checks the exception text.

    exception_text = "Required filter field ('c') in FilterByValues is not found in {'a': 1, 'b': 2}"
        check_operator_exception(
            operator=FilterByValues(required_values={"c": "5"}),
            inputs=inputs,
            exception_text=exception_text,
            tester=self,
        )

tests/test_operators.py
```diff
@@ -45,23 +46,86 @@ def compare_streams(self, all, expected_all):
...
+                targets=targets,
+                tester=self,
+            )
+
+        # strict is True by default, input value "3" in field "a" is missing from the mapper of "a"
+        with self.assertRaises(KeyError) as ke:
+            operator = MapInstanceValues(mappers=mappers)
+            operator.process(instance={"a": "3", "b": "4"})
+
+        inputs_p_e_v = [
```
yoavkatz, line 83: Why name is p_e_v?

yoavkatz (changes requested):
Hi @dafnapension . Please see a few comments below.

tests/test_operators.py
dafnapension: to be used with example case with process_every_value  is True.  Will extend to full name..

src/unitxt/operators.py
dafnapension: I accept.

tests/test_operators.py
dafnapension: Hi @yoavkatz ,
We have a hidden problem here: the mappers are defined as  str->str dictionaries, and when processing, the input values are first transformed to strings, then looked up at the dictionary, and the output is str again. Not back to int.  So any mapped value is always a str.  Could this make a problem?

src/unitxt/operators.py
```diff
@@ -71,32 +71,64 @@ class MapInstanceValues(StreamInstanceOperator):
...
+        when processing instance {"a": ["1", "2"], "b": 2}, will map the individual elements of field "a",
+        and output {"a": ["hi", "bye"], "b": 2}.
+        If strict = True, input {"a": ["1", "2", "3"], "b": 2} will raise a KeyError, because "3" is missing from the
+        mapper of "a".  When process_every_value = False, then the mapping would not be applied on "a", and, if Strict = True,
+        a KeyError will be raised, because the value of "a", which is a list, is not a key in the mapper of "a".
+
     """
 
     mappers: Dict[str, Dict[str, str]]
```
yoavkatz, line 87: I think this should be Dict[str, Dict[any, str]]

tests/test_operators.py
yoavkatz: I think the use of string internally is fine.  but people use it to map inputs that are not string, so we should keep this example.

Usually from int to string - as above.

src/unitxt/operators.py
```diff
@@ -75,13 +75,19 @@ class MapInstanceValues(StreamInstanceOperator):
...
-    Examples: MapInstanceValues with mappers={"a": {"1": "hi", "2": "bye"}}, and process_every_value = True,
-        when processing instance {"a": ["1", "2"], "b": 2}, will map the individual elements of field "a",
-        and output {"a": ["hi", "bye"], "b": 2}.
-        If strict = True, input {"a": ["1", "2", "3"], "b": 2} will raise a KeyError, because "3" is missing from the
-        mapper of "a".  When process_every_value = False, then the mapping would not be applied on "a", and, if Strict = True,
-        a KeyError will be raised, because the value of "a", which is a list, is not a key in the mapper of "a".
+    Examples:
 
+        Replace '1' with 'hi' and '2' with 'bye' in field 'a' in all instances of all streams:
+        MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}) will modify, for example,
```
yoavkatz, line 81:
```suggestion
        MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}) 
```

src/unitxt/operators.py
```diff
@@ -75,13 +75,19 @@ class MapInstanceValues(StreamInstanceOperator):
...
+        instance {"a":"1", "b": 2} to become {"a":"hi", "b": 2}.
```
yoavkatz, line 82:
```suggestion
        instance {"a":"1", "b": 2} to become {"a":"hi", "b": 2}.
        
```

src/unitxt/operators.py
```diff
@@ -75,13 +75,19 @@ class MapInstanceValues(StreamInstanceOperator):
...
+        Assuming field 'a' is a list of values, potentially including "1"-s and "2"-s, replacing each
+        such "1" with "hi" and "2" -- with "bye" can be achieved by employing
+        MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}, process_every_element=True). This will modify,
```
yoavkatz, line 85:
```suggestion
        MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}, process_every_element=True)
```

src/unitxt/operators.py
```diff
@@ -75,13 +75,19 @@ class MapInstanceValues(StreamInstanceOperator):
...
+        for example, instance {"a": ["1", "2"], "b": 2} to become {"a": ["hi", "bye"], "b": 2}.
+
+        To ensure that all values of field 'a' are mapped in every instance, use
+        MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}, strict=True). Here, input instance
```
yoavkatz, line 89:
```suggestion
        MapInstanceValues(mappers={"a": {"1": "hi", "2": "bye"}}, strict=True
```

dafnapension:
Thanks, @yoavkatz , 
I accept your good comments, and implemented.

PR #338, opened 2023-11-28
Stop using disk cache for SpreadSplit

elronbandel:


yoavkatz:
Can you explain the rational, why caching is not needed? Why was it originally added?

elronbandel:
@yoavkatz the cache on the disk is not needed because it already used cache on the memory `self.local_cache`.

yoavkatz (approved):
Remove comments.

PR #340, opened 2023-11-28
Fix templates for multi reference qa

elronbandel:


yoavkatz (changes requested):
As part of the change, we need to validate the field is a list of strings.
 
class MultiReferenceTemplate(InputOutputTemplate):
    references_field: str = "references"
    is_multi_reference = True

    def process_outputs(self, outputs: Dict[str, object]) -> List[str]:
        return outputs[self.references_field]

prepare/templates/qa/contextual.py
```diff
@@ -1,15 +1,19 @@
 from src.unitxt.catalog import add_to_catalog
-from src.unitxt.templates import InputOutputTemplate, TemplatesList
+from src.unitxt.templates import MultiReferenceTemplate, TemplatesList
 
 add_to_catalog(
-    InputOutputTemplate(input_format="Context: {context}\nQuestion: {question}", output_format="{answer}"),
+    MultiReferenceTemplate(
+        input_format="Context: {context}\nQuestion: {question}", output_format="{answers}", references_field="answers"
```
yoavkatz, line 6: We should not require output format in MultiReferenceTemplate.  This is because it is not used.

elronbandel: done

yoavkatz (approved):

PR #344, opened 2023-11-29
Improve error message for instance operator

elronbandel:


yoavkatz:
Decided not to change, to allow all information to be displayed in a single error message.

yoavkatz:
Decided to close , so error messages will be in a single line.

PR #347, opened 2023-11-29
augment suffix

dafnapension:


src/unitxt/operators.py
```diff
@@ -376,6 +376,67 @@ def process_value(self, value: Any) -> Any:
         return new_value
 
 
+class AugmentSuffix(Augmentor):
+    """
+    Augments the input by appending to it a randonly selected (typically, whitespace) pattern.
```
yoavkatz, line 381:
```suggestion
    Augments the input by appending to it a randomly selected (typically, whitespace) pattern.
```

src/unitxt/operators.py
```diff
@@ -376,6 +376,67 @@ def process_value(self, value: Any) -> Any:
...
+    Any existing trailig whitespace (if any), of the input is first discarded.
```
yoavkatz, line 382: Why you think we need this behavior?

src/unitxt/operators.py
```diff
@@ -376,6 +376,67 @@ def process_value(self, value: Any) -> Any:
...
+    Finally, the selected pattern is appended to the (potentially trimmed at its end) input.
+
+    Args:
+     suffixes : the potential white-space patterns to select from.
+        the dictionary version allowes to specify relative weights of the different patterns.
+
+    Examples:
+        to append a '\n' or a '\t' to the end of the input, employ
+        AugmentSuffix(augment_model_input=True, suffixes=['\n','\t'])
+
```
yoavkatz, line 393: Add example with dictionary.

src/unitxt/operators.py
```diff
@@ -376,6 +376,67 @@ def process_value(self, value: Any) -> Any:
...
+        super().verify()
+        self.next_pats = self.select_next_100_patterns()
+
+    def process_value(self, value: Any) -> Any:
+        assert value is not None, "input value should not be None"
+        new_value = str(value).rstrip()
+        assert len(new_value) > 0, f"nothing left of input value {str(value)} after trimming trailing whitespaces"
+        if len(self.next_pats) == 0:
+            self.next_pats = self.select_next_100_patterns()
```
yoavkatz, line 421: Why not select one element at a time?

tests/test_operators.py
```diff
@@ -1030,6 +1033,95 @@ def test_augment_whitespace_with_none_text_error(self):
...
+        outputs = apply_operator(operator, inputs)
+        output0 = str(outputs[0]["inputs"]["text"]).rstrip("".join(suffixes))
+        assert (
+            output0 == text[: len(output0)]
+        ), f"the prefix of {str(outputs[0]['inputs']['text'])} is not equal to the prefix of {text}"
+
+    def test_augment_suffix_with_non_string_suffixes_error(self):
+        text = "She is riding a black horse  \t\t  "
+        suffixes = [1, 2, "S", "T"]
+        with self.assertRaises(AssertionError) as ae:
```
yoavkatz, line 1086: Maybe we should take this code (which repeats) and create a utility function.  check_operator_verification_execption.  That looks like

  exception_text = "Error processing instance '0' from stream 'test' in AugmentSuffix due to: Error augmenting value 'None' from 'inputs/text' in instance: {'inputs': {'text': None}}"
        check_operator_verification_exception(
            operator,
            tester=self,
            exception_text=exception_text,
        )

yoavkatz (changes requested):
A few comments.  Main one regarding the default trimming of existing whitespace.
If we want to add it, it probably should be an explicit flag which is off by default.

src/unitxt/operators.py
dafnapension: Thanks, @yoavkatz , 
Not sure.. thought to clean the plate before laying food on it. But you are right that I shouldnt impose it, and that the user can pre-clean through Apply. Removing now.

src/unitxt/operators.py
dafnapension: I am not sure when the seed for the random is set. If it is set to same value each time this method is entered, then the same random result is output.

src/unitxt/operators.py
dafnapension: Fixed. Thanks!

src/unitxt/operators.py
dafnapension: Done. and also added the new flag, per your good idea.

tests/test_operators.py
dafnapension: Frankly, I do not like these test_operator-s. They swallow the original exception, and change to ValueError, and add more words to the original error message.

src/unitxt/operators.py
yoavkatz: The random set is set for every value that is being processed:


     with nested_seed(str(hash(old_value))):
                try:
                    new_value = self.process_value(old_value)
                except Exception as e:
                
 So it's ok.    
 
 You can calculate the weights only once.

Seems like simpler code.

yoavkatz: Hi.  You don't need to add the nested_seed wrapper, as it is handled in the parent class.

yoavkatz (approved):

PR #361, opened 2023-12-04
first draft for expand code coverage and documentation

dafnapension:


src/unitxt/operators.py
```diff
@@ -221,11 +234,11 @@ def process(
...
 
 class FieldOperator(StreamInstanceOperator):
-    """A general stream that processes the values of a field (or multiple ones.
+    """A general stream instance operator that processes the values of a field (or multiple ones).
 
     Args:
         field (Optional[str]): The field to process, if only a single one is passed Defaults to None
-        to_field (Optional[str]): Field name to save, if only one field is to be saved, if None is passed the operator would happen in-place and replace "field" Defaults to None
+        to_field (Optional[str]): Field name to save, if only one field is to be saved, if None is passed the operation would happen in-place and replace "field". Defaults to None
         field_to_field (Optional[Union[List[Tuple[str, str]], Dict[str, str]]]): Mapping from fields to process to their names after this process, duplicates are allowed. Defaults to None
```
dafnapension, line 242: Tuple - fails,  since the artifact is saved through json sumps, and the tuples are forgotten. The json string remembers lists, not tuples. and the test save_load throws an exception.

elronbandel: It sounds like we should allow only `Dict[str, str]` would you agree?

elronbandel (approved):

src/unitxt/operators.py
dafnapension: Yes, agree. I think it is sufficiently expressive.

PR #367, opened 2023-12-06
Add support for mapping full list in MapInstanceValues

yoavkatz:
In recent change, mapping of list was not supported .  This prevented valid uses cases such as mapping [] to ['None']/

This is a fix for this.

Also ensured that mapper values are deep copied to prevent overwritten .

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -134,31 +134,29 @@ class MapInstanceValues(StreamInstanceOperator):
                     )
                 if isinstance(value, list) and self.process_every_value:
                     for i, val in enumerate(value):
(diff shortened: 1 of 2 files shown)


elronbandel:
It's not critical but I think we should reuse code in this class. (as it is now is hard to follow)
I would split it to two functions:
```python
def process_value(val):
       val = str(val)  # make sure the value is a string
       if self.strict and (val not in self.mapper):
                  raise KeyError(
                       f"value '{val}' in instance '{instance}' is not found in mapper '{mapper}', associated with field '{key}'."
                 )
         if val in mapper:
               res = mapper[val]
              return res
```

also I think `dict_set` should be called only once at the end of process

PR #372, opened 2023-12-06
Improve API and coverage of CastFields and fix problem with casting list fields

dafnapension:


src/unitxt/operators.py
```diff
@@ -811,7 +814,7 @@ def _cast_single(self, value, type, field):
             return self.failure_defaults[field]
 
     def _cast_multiple(self, values, type, field):
-        values = [self._cast_single(value, type, field) for value in values]
+        return [self._cast_single(value, type, field) for value in values]
```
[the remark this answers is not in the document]
elronbandel: Amazing!

src/unitxt/operators.py
```diff
@@ -758,7 +758,7 @@ def process(
 
 
 class CopyFields(FieldOperator):
-    """Copies specified fields from one field to another.
+    """Copies values from specified fields to specified fields.
```
yoavkatz, line 761: Hi. I think it will be helpful to add examples, as this is a common operator.  Include the use of advanced options like use_query=True.

src/unitxt/operators.py
```diff
@@ -770,6 +770,8 @@ def process_value(self, value: Any) -> Any:
 
 
 class AddID(StreamInstanceOperator):
+    """Stores an id value in the designated 'id_field_name' field of the given instance."""
```
yoavkatz, line 773: What kind of id is added?  Worth adding, and showing an example.

src/unitxt/operators.py
```diff
@@ -783,10 +785,11 @@ class CastFields(StreamInstanceOperator):
     """Casts specified fields to specified types.
 
     Args:
-        types (Dict[str, str]): A dictionary mapping fields to their new types.
+        types (Dict[str, str]): A dictionary mapping type names to types.
```
yoavkatz, line 788: Not sure if this is right.  It maps between between field names and their new types.

src/unitxt/operators.py
```diff
@@ -783,10 +785,11 @@ class CastFields(StreamInstanceOperator):
     """Casts specified fields to specified types.
 
     Args:
-        types (Dict[str, str]): A dictionary mapping fields to their new types.
```
yoavkatz, line 786: I don't think this field should be visible to users.

src/unitxt/operators.py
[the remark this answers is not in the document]
yoavkatz: I think should rename cast_multiple to process_every_value, like in other operators.

Also, rename use_nested_query to use_query (like in other operators).  [Although we may want to change everything to use_nested_query)

yoavkatz (approved):
Added some comments - many based on fresh look at this operator.

yoavkatz (changes requested):
Hi Dafna.  See some suggested changes.

src/unitxt/operators.py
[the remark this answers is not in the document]
dafnapension: Would you suggest, then, to 'upgrade' argument `process_every_value`  from `FieldOperator` (where it currently sits) to the parent class `StreamInstanceOperator`, which is also the parent class of this ,`CastFields`, class?

src/unitxt/operators.py
dafnapension: Operator `Addit` is only used once, in [prepare/metrics/squad.py](https://github.com/IBM/unitxt/blob/5df775d70b0cbb65a1552d4ddf3367475364d145/prepare/metrics/squad.py#L2). with no effect at all. Seems like no effect is expected from it. I suggest to remove class `AddIt` altogether.

elronbandel (approved):

src/unitxt/operators.py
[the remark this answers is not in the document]
yoavkatz: I don't suggest to move the the process_every_value to FieldOperator.  Rather, I suggest using the same name in both classes for consistency.

A larger change would be to make this into a FieldOperator (not sure I would go for it now)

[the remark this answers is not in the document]
dafnapension: OK. Done. changed locally  cast_multiple -> process_every_value

src/unitxt/operators.py
yoavkatz: I think it should not be documented, as users are not supposed to change it.

yoavkatz (approved):
Minor point about 'type' in documentation of CastFields.

PR #375, opened 2023-12-08
expand augment suffix to become augment prefix suffix

dafnapension:


src/unitxt/operators.py
```diff
@@ -508,75 +508,121 @@ def process_value(self, value: Any) -> Any:
...
         new_value = str(value)
-        if self.remove_existing_trailing_whitespaces:
-            new_value = new_value.rstrip()
-        new_value += get_random().choices(self.pats, self.weights, k=1)[0]
+        if self.remove_existing_whitespaces:
+            new_value = new_value.strip()
+        prefix = ""
+        suffix = ""
+        if self.prefixes:
+            prefix = "".join(
```
yoavkatz, line 611: I suggest extracting this common code to a function.

src/unitxt/operators.py
```diff
@@ -508,75 +508,121 @@ def process_value(self, value: Any) -> Any:
...
+                get_random().choices(
+                    self.pats["prefixes"], self.weights["prefixes"], k=1
+                )
+                * get_random().randint(1, 5)
```
yoavkatz, line 615: Let's have the number of repetition also be a parameter of the augmentor.

src/unitxt/operators.py
```diff
@@ -508,75 +508,121 @@ def process_value(self, value: Any) -> Any:
...
-            [1.0 / total_weight] * len(self.pats)
-            if isinstance(self.suffixes, list)
-            else [float(self.suffixes[p]) / total_weight for p in self.pats]
-        )
+        # Being an artifact, prepare is invoked before verify. Here we need verify before the actions
+        self.verify()
+        self.pats = {"prefixes": None, "suffixes": None}
+        self.weights = {"prefixes": None, "suffixes": None}
+
+        for end, end_name in zip(
```
yoavkatz, line 582: I think it would be better to use functions for this

(self.prefix_patterns, self.prefix_probabilities) = calculate_distributions(self.prefixes)
(self.suffix_patterns, self.suffix_probabilities) = calculate_distributions(self.suffixes)

src/unitxt/operators.py
```diff
@@ -508,75 +508,121 @@ def process_value(self, value: Any) -> Any:
...
-                ), f"suffixes should map to ints, whereas value {v!s} is of type {type(v)}"
-        else:
-            for k in self.suffixes:
-                assert isinstance(
-                    k, str
-                ), f"suffixes should be a list of strings, whereas member {k!s} is of type {type(k)}"
+            self.prefixes or self.suffixes
+        ), "At least one of prefixes/suffixes should be not None."
+        for arg, arg_name in zip(
+            [self.prefixes, self.suffixes], ["prefixes", "suffixes"]
```
yoavkatz, line 553: Elron has a better way for type checking.  @elronbandel - can you please advise?

yoavkatz (changes requested):
Hi Dafna.  Thanks.  Please the suggested feedback.

src/unitxt/operators.py
dafnapension: OK. Done.

src/unitxt/operators.py
dafnapension: Done. Indeed cleaner. Thanks!

src/unitxt/operators.py
```diff
@@ -508,75 +508,124 @@ def process_value(self, value: Any) -> Any:
...
+            self.suffix_len > 0
+        ), f"suffix_len must be positive, got {self.suffix_len}"
         super().verify()
 
-    def prepare(self):
-        self.pats = (
-            self.suffixes
-            if isinstance(self.suffixes, list)
-            else [k for k, v in self.suffixes.items()]
+    def _calculate_distribution(self, end, end_name):
```
yoavkatz, line 584: Hi Dafna, I think this kind of implicit assigned through names is harder to read.  Better to have functions that receive input  and not modify them and then return outputs.  Then the caller can assign the value to the right field.

src/unitxt/operators.py
```diff
@@ -508,75 +508,124 @@ def process_value(self, value: Any) -> Any:
...
         assert value is not None, "input value should not be None"
         new_value = str(value)
-        if self.remove_existing_trailing_whitespaces:
-            new_value = new_value.rstrip()
-        new_value += get_random().choices(self.pats, self.weights, k=1)[0]
-
-        return new_value
+        if self.remove_existing_whitespaces:
+            new_value = new_value.strip()
+        prefix = self._get_random_pattern("prefixes")
```
yoavkatz, line 626: Here also - best to pass in  pats["suffixes"], and weights["suffixes"] and not extract it in the functions.

Perhaps also call pats -> patterns

elronbandel (changes requested):
Complex type checking should be done with `isoftype` in `unitxt.type_utils`. If it does not work for this case you might want to either augment the function or add new one to type_utils, preferably with appropriate unittests.

src/unitxt/operators.py
elronbandel: added relevant comment in the main thread

src/unitxt/operators.py
```diff
@@ -508,75 +509,122 @@ def process_value(self, value: Any) -> Any:
...
+        ), "At least one of prefixes/suffixes should be not None."
+        for arg, arg_name in zip(
+            [self.prefixes, self.suffixes], ["prefixes", "suffixes"]
+        ):
+            assert (
+                arg is None
+                or isinstance(arg, list)
+                and isoftype(arg, List[str])
+                or isinstance(arg, dict)
+                and isoftype(arg, Dict[str, int])
```
elronbandel, line 563: maybe simpler:
```suggestion
                or isoftype(arg, List[str])
                or isoftype(arg, Dict[str, int])
```

src/unitxt/operators.py
dafnapension: Done. Now I finally got it what you said in the first place..

src/unitxt/operators.py
dafnapension: Thanks!  Done.

src/unitxt/operators.py
dafnapension: Hi @elronbandel ,
ahh!  I tried it first. But punkt the test I used, uncovers a problem:  `isoftype(arg, Dict[...])` rushes into `arg.items()` before verifying that this `arg` has `items()` at all!
So I added these.
I think that `isoftype` should better be wrapped with cautions alike, but was not sure how to..

src/unitxt/operators.py
```diff
@@ -508,75 +509,122 @@ def process_value(self, value: Any) -> Any:
...
+            self._prefixes["patterns"],
+            self._prefixes["weights"],
+        ) = self._calculate_distributions(self.prefixes)
+        (
+            self._suffixes["patterns"],
+            self._suffixes["weights"],
+        ) = self._calculate_distributions(self.suffixes)
         super().prepare()
 
+    def _get_random_pattern(self, _prefs_or_suffs) -> str:
```
yoavkatz, line 609: I think function params should not started with _ (reserved for private members)

src/unitxt/operators.py
dafnapension: Hi @elronbandel , [I tried my best](https://github.com/IBM/unitxt/pull/375/files#diff-42f0a8d39d64c93bedee0b3ec80505ef104b89e81064dcdd31588ac2cbf93b36), please review..

dafnapension: Hi @elronbandel , please see if the change I made to typeof is OK.

elronbandel (approved):

dafnapension:
> Hi Dafna. Thanks. Please the suggested feedback.

Hi @yoavkatz , please see if you can approve the PR now.

PR #376, opened 2023-12-08
The loop business

dafnapension:
catch fields participating in the field_to_field input list of FieldOperator in both roles of from_field and to_field, and not in the same pair.


yoavkatz:
Very nice catch of a subtle bug.

elronbandel (commented):
I think it might be better to handle the assertion by raising a ValueError as it is more explicit and cannot be skipped by python interpeters

elronbandel (approved):

PR #377, opened 2023-12-08
Add caching to LoadFromIBMCos,

yoavkatz:
Location is determined by IBM_COS_CACHE env var,
or by cwd if env var is not set.

Cache is enabled by default.

### DIFF
--- a/src/unitxt/loaders.py
+++ b/src/unitxt/loaders.py
@@ -152,7 +152,7 @@ class LoadFromIBMCloud(Loader):
         self.endpoint_url = os.getenv(self.endpoint_url_env)
         self.aws_access_key_id = os.getenv(self.aws_access_key_id_env)
         self.aws_secret_access_key = os.getenv(self.aws_secret_access_key_env)
-        root_dir = os.getenv("IBM_COS_CACHE", None) or os.getcwd()
+        root_dir = os.getenv("UNITXT_IBM_COS_CACHE", None) or os.getcwd()
         self.cache_dir = os.path.join(root_dir, "ibmcos_datasets")
 
         if not os.path.exists(self.cache_dir):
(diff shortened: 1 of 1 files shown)


src/unitxt/loaders.py
```diff
@@ -106,6 +106,7 @@ class LoadFromIBMCloud(Loader):
     bucket_name: str
     data_dir: str = None
     data_files: Sequence[str]
+    caching: str = True
```
dafnapension, line 109: str or bool?

src/unitxt/loaders.py
```diff
@@ -152,6 +152,11 @@ def prepare(self):
         self.endpoint_url = os.getenv(self.endpoint_url_env)
         self.aws_access_key_id = os.getenv(self.aws_access_key_id_env)
         self.aws_secret_access_key = os.getenv(self.aws_secret_access_key_env)
+        root_dir = os.getenv("IBM_COS_CACHE", None) or os.getcwd()
```
elronbandel, line 155:
```suggestion
        root_dir = os.getenv("UNITXT_IBM_COS_CACHE", None) or os.getcwd()
```

elronbandel (approved):
I approve it for now so we can enjoy the performance gain. Yet I want to raise a concern regarding a new separate cache mechanism without transparency or control. 

Ideally we would like to have a separated caching mechanism with proper unitests loggings and documented control methods.

yoavkatz:
> I approve it for now so we can enjoy the performance gain. Yet I want to raise a concern regarding a new separate cache mechanism without transparency or control.
> 
> Ideally we would like to have a separated caching mechanism with proper unitests loggings and documented control methods.

I agree.  As caching is post important in IO operations, then loaders are the most important ones.  The point is that most loaders including HF and Lakehouse already have support for caches, so the question is whether we want to use these native caches or use our own.

Issue #378, opened 2023-12-08
Subtle issue in ExtractFieldValues


yoavkatz:
The tests of ExtractFieldValues work well but when I use it, I saw the the field was not added.

The reason , I think is that process method modified the stream in place and does not create a new stream generator that modifies the input streams.

```
   for name in multi_stream:
            for instance in multi_stream[name]:
                instance[self.to_field] = values_to_keep
        return multi_stream

```        
  
So if you add, before the return.
```
 for name in multi_stream:
         for instance in multi_stream[name]:
            print(instance)
```
            
 You will see that  the instance (which is a actually new instance that is fetched again from the multi-stream[name], does not include the change ) 

I think the implementation should be similar to :

`class SpreadSplit(InstanceOperatorWithMultiStreamAccess):
`
which has access to the multi stream, but then adds a single value to the instance.

```
class ExtractFieldValues(InstanceOperatorWithMultiStreamAccess):
    field: str
    stream_name: str
    overall_top_frequency_percent: Optional[int] = 100
    min_frequency_percent: Optional[int] = 0
    to_field: str
    process_every_value: Optional[bool] = False

    def prepare(self):
        self.local_cache = None

    def verify(self):
            return super().verify()

    def process(
        self, instance: Dict[str, object], multi_stream: MultiStream
    ) -> Dict[str, object]:
        try:
            if self.local_cache is None:
                self.local_cache =  calculate_extracted_values(multi stream)
         
            instance[self.to_field] =  self.local_cache
            return instance
        except Exception as e:
            raise Exception(
                f"Unable to fetch instances from '{self.source_stream}' to '{self.target_field}'"
            ) from e
        

```

dafnapension:
Hi Yoav,  very nice and important catch!
I managed to reconstruct, and generated a test for it.
And solved with deep copies.
Please see [PR](https://github.com/IBM/unitxt/pull/379)

PR #379, opened 2023-12-08
deep copy at extract_field_values, in order to allow the same input m…

dafnapension:
…s to feed different runs of operator extract_field_values, while maintaining all the different outpus.  For [issue 378](https://github.com/IBM/unitxt/issues/378)

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -971,21 +971,6 @@ class FilterByValues(SingleStreamOperator):
                 yield instance
 
 
-class AddFieldValues(SingleStreamOperator):
-    """Adds fields and their values to all instances of a stream.
-
-    Args: fields_values : the fields and their respective vaulues to add to each instance in the stream
-    """
-
-    fields_values: Dict[str, Any]
-
-    def process(self, stream: Stream, stream_name: Optional[str] = None):
-        for instance in stream:
-            for f, v in self.fields_values.items():
-                instance[f] = v
-            yield instance
-
-
 class ExtractMostCommonFieldValues(MultiStreamOperator):
     field: str
     stream_name: str
@@ -1083,7 +1068,7 @@ class ExtractMostCommonFieldValues(MultiStreamOperator):
             for ele in values_and_counts
         ]
 
-        addmostcommons = AddFieldValues(fields_values={self.to_field: values_to_keep})
+        addmostcommons = AddFields(fields={self.to_field: values_to_keep})
         return addmostcommons(multi_stream)
 
 
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1087,6 +1087,15 @@ class ExtractMostCommonFieldValues(MultiStreamOperator):
         return addmostcommons(multi_stream)
 
 
+class ExtractFieldValues(ExtractMostCommonFieldValues):
+    def verify(self):
+        super().verify()
+
+    def prepare(self):
+        self.overall_top_frequency_percent = 100
+        self.min_frequency_percent = 0
+
+
 class FilterByListsOfValues(SingleStreamOperator):
     """Filters a stream, yielding only instances that  whose field values are included in the specified value lists.
 
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1048,13 +1048,13 @@ class ExtractFieldValues(MultiStreamOperator):
             [*ele[0]] if isinstance(ele[0], tuple) else ele[0]
             for ele in values_and_counts
         ]
-        multi_stream_to_return = {}
+        add_field_operator = AddFields(
+            fields={self.to_field: values_to_keep}, use_query=False, use_deepcopy=True
+        )
         for name in multi_stream:
-            multi_stream_to_return[name] = []
             for instance in multi_stream[name]:
-                instance[self.to_field] = values_to_keep
-                multi_stream_to_return[name].append(deepcopy(instance))
-        return multi_stream_to_return
+                add_field_operator.process(instance)
+        return multi_stream
 
 
 class FilterByListsOfValues(SingleStreamOperator):
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1029,10 +1029,13 @@ class ExtractFieldValues(MultiStreamOperator):
             [*ele[0]] if isinstance(ele[0], tuple) else ele[0]
             for ele in values_and_counts
         ]
+        multi_stream_to_return = {}
         for name in multi_stream:
+            multi_stream_to_return[name] = []
             for instance in multi_stream[name]:
                 instance[self.to_field] = values_to_keep
-        return multi_stream
+                multi_stream_to_return[name].append(deepcopy(instance))
+        return multi_stream_to_return
 
 
 class FilterByListsOfValues(SingleStreamOperator):
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1083,17 +1083,8 @@ class ExtractFieldValues(MultiStreamOperator):
             for ele in values_and_counts
         ]
 
-        multi_stream_to_return = {}
-        addmostcommon_stream_operator = AddFieldValues(
-            fields_values={self.to_field: values_to_keep}
-        )
-        for stream_name, stream in multi_stream.items():
-            updated_stream = addmostcommon_stream_operator._process_single_stream(
-                stream
-            )
-            multi_stream_to_return[stream_name] = updated_stream
-
-        return MultiStream(multi_stream_to_return)
+        addmostcommons = AddFieldValues(fields_values={self.to_field: values_to_keep})
+        return addmostcommons(multi_stream)
 
 
 class FilterByListsOfValues(SingleStreamOperator):
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1007,21 +1007,21 @@ class ExtractMostCommonFieldValues(MultiStreamOperator):
 
     Examples:
 
-    ExtractFieldValues(stream_name="train", field="label", to_field="classes") - extracts all the unique values of
+    ExtractMostCommonFieldValues(stream_name="train", field="label", to_field="classes") - extracts all the unique values of
     field 'label', sorts them by decreasing frequency, and stores the resulting list in field 'classes' of each and
     every instance in all streams.
 
-    ExtractFieldValues(stream_name="train", field="labels", to_field="classes", process_every_value=True) -
+    ExtractMostCommonFieldValues(stream_name="train", field="labels", to_field="classes", process_every_value=True) -
     in case that field 'labels' contains a list of values (and not a single value) - track the occurrences of all the possible
     value members in these lists, and report the most frequent values.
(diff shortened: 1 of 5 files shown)


src/unitxt/operators.py
```diff
@@ -1029,10 +1029,13 @@ def process(self, multi_stream: MultiStream) -> MultiStream:
             [*ele[0]] if isinstance(ele[0], tuple) else ele[0]
             for ele in values_and_counts
         ]
+        multi_stream_to_return = {}
         for name in multi_stream:
```
elronbandel, line 1033: First time seeing this, I think this might should be done in a lazy way like other MultiStreamOpertors?

dafnapension: Hi @elronbandel ,
It seems to me that all MultiStreamOperator-s in operators.py return actual multi-stream-s, not a dictionary of generators. So I think that also `process` of this class has to un-lazy the individual streams.  And this is actually equivalent to what we currently have.

yoavkatz: I suggested a larger change in the issue by using a different base class:

```
class ExtractFieldValues(InstanceOperatorWithMultiStreamAccess):
    field: str
    stream_name: str
    overall_top_frequency_percent: Optional[int] = 100
    min_frequency_percent: Optional[int] = 0
    to_field: str
    process_every_value: Optional[bool] = False

    def prepare(self):
        self.local_cache = None

    def verify(self):
            return super().verify()

    def process(
        self, instance: Dict[str, object], multi_stream: MultiStream
    ) -> Dict[str, object]:
        try:
            if self.local_cache is None:
                self.local_cache =  calculate_extracted_values(multi stream)
         
            instance[self.to_field] =  self.local_cache
            return instance
        except Exception as e:
            raise Exception(
                f"Unable to fetch instances from '{self.source_stream}' to '{self.target_field}'"
            ) from e
  ````          
            
            
  The parent class uses the lazy scheme:
  ```

  class InstanceOperatorWithMultiStreamAccess(StreamingOperator):
    """A class representing an instance operator with global access in the streaming system.

    An instance operator with global access is a type of `StreamingOperator` that operates on individual instances within a `Stream` and can also access other streams.
    It uses the `accessible_streams` attribute to determine which other streams it has access to.
    In order to make this efficient and to avoid qudratic complexity, it caches the accessible streams by default.
    """

    def __call__(self, multi_stream: Optional[MultiStream] = None) -> MultiStream:
        result = {}

        for stream_name, stream in multi_stream.items():
            stream = Stream(
                self.generator,
                gen_kwargs={"stream": stream, "multi_stream": multi_stream},
            )
            result[stream_name] = stream

        return MultiStream(result)

    def generator(self, stream, multi_stream):
        yield from (self.process(instance, multi_stream) for instance in stream)

    @abstractmethod
    def process(self, instance: dict, multi_stream: MultiStream) -> dict:
        pass
```

dafnapension: Thanks, @yoavkatz , still:  how did you realize there was a problem with the previous (now in main) implementation of ExtractFieldValue?
Please share so I can make it into an important unit test

yoavkatz: I happened during the use in a real use case.   I saw that the field which was presumably extracted was not there  there.  I'm not 100% sure how to recreate the scenario in unitest.

dafnapension: This is the only reason to shake the current implementation (now in main). @elronbandel says it is OK streaming-lazyness-wise.
The whole field was missing? or the name was still there, but no value (being the list of most common values) in that field?

yoavkatz: The whole field was missing.   Maybe you can try to use it in a card, and actually make it work.  For example, in /Users/yoavkatz/migrate/v10/unitxt/prepare/cards/sst2.py replace 
```
AddFields(
            fields={
                "classes": ["negative", "positive"],
                ```
                
                With the ExtractFieldValues.

Comment (dafnapension) [reply src/unitxt/operators.py:1054]: @yoavkatz , @elronbandel , I committed a new version, per Elron's idea, which employs AddFields operator.  With it, it is easy to set deep_copy, which can help in the cases that Yoav ran into.

Comment (dafnapension) [reply src/unitxt/operators.py]: @yoavkatz , @elronbandel , I committed yet another very lazy implementation, shamelessly adopted from the only relevant multistream->multistream sister class: `SplitByValue`.  I also commit a version of `sst2.py`, and the yielded card `sst2.json`, just for you to see: `sst2.py` runs all the way through (successfully completing test_card). That is the only sign that I can see from here that everything is alright. 
I suggest **not** to merge sst2 into main. This is just for you to see that the fields are not gone.

Comment (dafnapension) [reply src/unitxt/operators.py]: well, test_recipe.py caught the change of order of `positive`, `negative` inside field `classes`.  I need to learn how the template is generated from classes. For now, I just allowed the test to pass.

File: src/unitxt/operators.py [hunk 2]
```diff
@@ -1048,10 +1063,18 @@ def process(self, multi_stream: MultiStream) -> MultiStream:
...
+        addmostcommon_stream_operator = AddFieldValues(
+            fields_values={self.to_field: values_to_keep}
+        )
+        for stream_name, stream in multi_stream.items():
+            updated_stream = addmostcommon_stream_operator._process_single_stream(
+                stream
+            )
+            multi_stream_to_return[stream_name] = updated_stream
+
+        return MultiStream(multi_stream_to_return)
```
elronbandel, line 1077:
```suggestion
        result_multi_stream = AddFieldValues(
            fields_values={self.to_field: values_to_keep}
        )
        return result_multi_stream
```
I think this could work

elronbandel (changes requested):
Look at my suggestion. Also maybe it should be called `ExtractFieldUniqueValues`?

dafnapension:
Thanks, @elronbandel , will soon check your shortcut.
`ExtractMostCommonValues` is closest to the operation it performs.

src/unitxt/operators.py
yoavkatz: Thanks Dafna. I think we should keep the new sst2, because it's a good example.

src/unitxt/operators.py
dafnapension: Did not work:

```
  File "/home/dafna/workspaces/unitxt/tests/test_operators.py", line 691, in test_extract_values

    for k, v in output_multi_stream1.items():

AttributeError: 'AddFieldValues' object has no attribute 'items'
```

dafnapension: Myself tried to copy all functions/methods from AddFieldValues and the functions invoked from them, copy all into ExtractFieldValues, by the word. And this did not work for me.  Looks deeper into Python objects and their properties that I could follow..

elronbandel: sorry should be :
```python 
add_values = AddFieldValues(
            fields_values={self.to_field: values_to_keep}
        )
        result_multi_stream = add_values(original_multi_stream)
        return result_multi_stream
```
or similar

dafnapension:
@yoavkatz , we already have `Unique` which returns the list of unique values encountered in a given field. This is what you would like `ExtractFieldValues` (in the designed incarnation) to do. Isnt it?

dafnapension:
> Look at my suggestion. Also maybe it should be called `ExtractFieldUniqueValues` or even more fluent `ExtractUniqueFieldValues`?

Hi @elronbandel , I renamed the class to `ExtractMostCommonValues`

src/unitxt/operators.py
```diff
@@ -971,7 +971,22 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
                 yield instance
 
 
-class ExtractFieldValues(MultiStreamOperator):
+class AddFieldValues(SingleStreamOperator):
```
yoavkatz, line 974: We already have a class AddFields(StreamInstanceOperator), why do we need this one?

yoavkatz:
I think we should have ExtractMostCommonFieldValues, but also have
ExtractFieldValues like before, where all params are verified in verify() such  that all the values are returned.

This is because it's unnatural (e.g. in sst2) to call  ExtractMostCommonFieldValues to get all values.

src/unitxt/operators.py
dafnapension: I am aware of this `AddFields`. But, could very well be due to my ignorance, I needed a Stream processor, and not an instance processor. I looked for an example with generators and with streams, and found `SplitByValue`, and adapted from it. Perhaps I can learn something now.  E.g. if I knew how to produce a StreamProcessor  (whose .process() is fed with stream, and it outputs a generator of a stream) from a StreamInstance processor which like `AddFields`, simply returns a dict.

dafnapension:
> I think we should have ExtractMostCommonFieldValues, but also have ExtractFieldValues like before, where all params are verified in verify() such that all the values are returned.
> 
> This is because it's unnatural and confusing (e.g. in sst2) to call ExtractMostCommonFieldValues to get all values.

I think that DAVKA here, the case that motivated you to initiate this operator (classify among a huge set of potential classes, and you wanted to reduce the choice to a reasonable set of candidates), it is a good example.  Although here, for only two classes, indeed no hassle is needed; these two classes better and safer be written by hand.  I changed sst2 **only** in order to show to you that `to_field` is not gone. I even recommended to **not** merge the new sst2 into main.

dafnapension:
> I think we should have ExtractMostCommonFieldValues, but also have ExtractFieldValues like before, where all params are verified in verify() such that all the values are returned.
> 
> This is because it's unnatural and confusing (e.g. in sst2) to call ExtractMostCommonFieldValues to get all values.

So perhaps I better extend `verify()` of the existing `Unique`? This `Unique` does return a list of all the different values encountered in a given field (or unique tuples, in case of two or more given fields)

yoavkatz:
Unique is more of a utility operator:

def test_unique_on_single_field(self):
        inputs = [
            {"a": 1, "b": 2},
            {"a": 2, "b": 3},
            {"a": 2, "b": 4},
        ]

        targets = {(1,), (2,)}

        outputs = apply_operator(
            operator=Unique(fields=["a"]),
            inputs=inputs,
        )

        self.assertSetEqual(set(outputs), targets)


It totally changes the output stream.    IT is used in internally SplitByValue. It also does not deal in the expected way with lists of values (it considers them as unique tuples).

dafnapension:
> Unique is more of a utility operator:
> 
> def test_unique_on_single_field(self): inputs = [ {"a": 1, "b": 2}, {"a": 2, "b": 3}, {"a": 2, "b": 4}, ]
> 
> ```
>     targets = {(1,), (2,)}
> 
>     outputs = apply_operator(
>         operator=Unique(fields=["a"]),
>         inputs=inputs,
>     )
> 
> It totally changes the output stream. IT is used in internally SplitByValue. It also does not deal in the expected way with lists of values (it considers them as unique tuples).

I see. and you aim at the behavior of `ExtractMostCommonFieldValues`, that pushes the unique values into every instance of the input multi-stream, right?  And the only difference from `ExtractMostCommonFieldValues`, is that the behavior you need pushes **all** the different unique values, and not only the popular (most common) ones?

dafnapension:
Hi @elronbandel , please see whether your request is well addressed.
And thanks so much for the lesson you taught me with your demonstration how to produce a generator-based result.

yoavkatz (approved):
LGTM

elronbandel (approved):

PR #388, opened 2023-12-11
cover recursive divide. Really needed?

dafnapension:
Nowhere is recursive divide invoked. It does not look a sophisticated piece of work that is hard to generate only when needed.  Do we need it at all?

### DIFF
--- a/tests/test_operators.py
+++ b/tests/test_operators.py
@@ -41,6 +41,7 @@ from src.unitxt.operators import (
     TakeByField,
     Unique,
     ZipFieldValues,
+    recursive_divide,
 )
 from src.unitxt.stream import MultiStream
 from src.unitxt.test_utils.operators import (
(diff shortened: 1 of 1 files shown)


elronbandel:
it seems like its being used by 

```
class DivideAllFieldsBy(StreamInstanceOperator):
    divisor: float = 1.0
    strict: bool = False
    recursive: bool = True

    def process(
        self, instance: Dict[str, Any], stream_name: Optional[str] = None
    ) -> Dict[str, Any]:
        return recursive_divide(instance, self.divisor, strict=self.strict)
```

which has many issues, like it does not use the same schema as other field to field classes. and can be applied to containers and not only integers and floats. this can create ambiguity and lack of control.

dafnapension:
Thanks, @elronbandel , I missed this use..

yoavkatz:
As far as I know DivideAllFieldsBy is not used.

As you mentioned, it does not follow the field operator.

In my view this code should be deleted.

The other alternative is to rewrite it as a FieldOperator, but we can delay it until is actually needed.

elronbandel:
Then it should be in a different PR. 

CC: @dafnapension

PR #390, opened 2023-12-12
Fixed error check in multi label F1

yoavkatz:


src/unitxt/metrics.py
```diff
@@ -723,7 +723,7 @@ def compute(
 
         for reference in references:
             assert isinstance(
-                references, list
+                reference, list
             ), f"Each reference is expected to list of strings in F1 multi label metric. Received reference: {reference}"
```
elronbandel, line 727: I think it should be clearer with
```python
if not isoftype(List[List[str]]):
    rasie ValueError("...")
```

elronbandel: @yoavkatz this should work now due to https://github.com/IBM/unitxt/pull/395

src/unitxt/metrics.py
```diff
@@ -777,6 +766,23 @@ def compute(
             final_result = {self.main_score: result["f1"]}
         return final_result
 
+    def _validate_references_and_prediction(self, references, predictions):
+        for reference in references:
+            if not len(reference) == 1:
+                raise ValueError(
+                    f"Only a single reference per prediction is allowed in F1 multi label metric. Received reference: {reference}"
+                )
+            if not isoftype(reference[0], List[str]):
```
dafnapension, line 775: reference[0]? or reference?

elronbandel (approved):

PR #391, opened 2023-12-12
Model diff sig tests

sam-data-guy-iam:
@matanor 
merging of statistical tests on LLM metric comparisons, and visualizations.  Also included a notebook showing how to use them


requirements/base.rqr
```diff
@@ -8,3 +8,8 @@ scikit-learn
 dpath
 jiwer
 editdistance
+statsmodels
+scipy
+seaborn
+matplotlib
```
yoavkatz, line 14: Do we need all these requirements?  

Perhaps some can be moved to tests.rqr? Because they are only relevant if you use metric_paried_significance.

src/unitxt/operators.py
```diff
@@ -35,7 +35,6 @@
 from .random_utils import get_random, nested_seed
```
yoavkatz, line 35: These changes seem like a problematic merge.

tests/test_operators.py
```diff
@@ -10,7 +10,7 @@
     ApplyMetric,
```
yoavkatz, line 10: These changes seem like a problematic merge.

requirements/base.rqr
sam-data-guy-iam: No problem, I can move them there.  They are not only for testing applications, though, but if you think that's ok I will do it

src/unitxt/operators.py
sam-data-guy-iam: the changes to operators and the test is a mistake, I am removing them

sam-data-guy-iam:
Have decided it may be better to make this its own repo outside of unitxt since it is of general use.

PR #393, opened 2023-12-12
not exposing types dictionary. back to backward compatibility

dafnapension:
not exposing types dictionary, rather - generating it for the needed type names

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -817,7 +817,7 @@ class CastFields(StreamInstanceOperator):
     """Casts specified fields to specified types.
 
     Args:
-        use_query (bool): Whether to cast nested fields, expressed in dpath. Defaults to False.
+        use_nested_query (bool): Whether to cast nested fields, expressed in dpath. Defaults to False.
         fields (Dict[str, str]): A dictionary mapping field names to the names of the types to cast the fields to.
             e.g: "int", "str", "float", "bool". Basic names of types
         defaults (Dict[str, object]): A dictionary mapping field names to default values for cases of casting failure.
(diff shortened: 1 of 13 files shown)


dafnapension:
Hi @yoavkatz , I rushed to merge https://github.com/IBM/unitxt/pull/372 before it accumulates conflicts, and overlooked your comment at the, when you approved.
Perhaps this #393  completes the mission you had in mind?

src/unitxt/operators.py
```diff
@@ -817,24 +817,35 @@ class CastFields(StreamInstanceOperator):
...
-    }
     fields: Dict[str, str] = field(default_factory=dict)
     failure_defaults: Dict[str, object] = field(default_factory=dict)
-    use_nested_query: bool = False
+    use_query: bool = False
     process_every_value: bool = False
 
+    def prepare(self):
+        self.types = {}
+        apply_operator = Apply(function=str, to_field="__")
```
yoavkatz, line 845: Is this just a  dummy operator so you can call  str_to_function? 
If so you can just change the function to a @classmethod (it does not use @self) 

However, I also don't want to user to be able to call any function here (for security reasons -  Apply is a risky operator because it allows running arbitrary code).   Hence, I prefer you store original fixed map.

yoavkatz (commented):
1. This is breaking change (there are data cards that use CastField  "use_query").  We need to note in the next release note  that and fix downstream projects
2. Please see my comment about the use of Apply (we should try to avoid)

src/unitxt/operators.py
dafnapension: I see. I try another trick now, that I learned from Google, perhaps you will like better.

dafnapension:
> 1. This is breaking change (there are data cards that use CastField  "use_query").  We need to note in the next release note  that and fix downstream projects
> 2. Please see my comment about the use of Apply (we should try to avoid)

Then I would not push for a revolution now. The cause is not that dramatic.  I will return back to the good old name, `use_nested_query`, and withdraw from all the implied changes of name, over the several files.  No big deal at all.

dafnapension:
Hi @yoavkatz , the PR shrank to natural size: only hiding the setting of the type-map.  No longer change of flag-names that I thought I could handle, forgetting that I only see a limited picture.
Please see if you can now approve.

yoavkatz (approved):
LGTM

PR #394, opened 2023-12-13
metric and processor related to robustness evaluation

ellarabi:


src/unitxt/metrics.py
```diff
@@ -474,6 +474,24 @@ def compute(
...
 
 
+class StringContainment(InstanceMetric):
+    reduction_map = {"mean": ["string_containment"]}
+    main_score = "string_containment"
+
+    def compute(
+        self, references: List[Any], prediction: Any, additional_inputs: List[Dict]
+    ) -> dict:
+        references = [r.strip() for r in references[0].split("|")]
```
matanor, line 484: IMO it would be nice if we could avoid this split step in the metric, since a metric already has a way of receiving multiple references. We just need to figure out how to pass multiple references to the metric, from how they are represented in your dataset (where they are in a single string separated by a '|'). I think most suitable would be to have some pre-processing step, that takes your input the single references string, and does the splitting, converting it to a list of strings.
I will try to lookup an example for that.

prepare/metrics/string_containment.py
```diff
@@ -0,0 +1,33 @@
+from src.unitxt import add_to_catalog
+from src.unitxt.metrics import StringContainment
+from src.unitxt.test_utils.metrics import test_metric
+
+metric = StringContainment()
+
+predictions = ["barak obama is a politician", "David Gilmour is an English guitarist"]
+references = [["politician | politic | pol"], ["artist"]]
```
matanor, line 8:
```suggestion
references = [["politician", "politic, "pol"], ["artist"]]
```

matanor: I think it will be better to use the option of providing multiple references to the metric. Then we are not limited to this specific encoding with the '|' separator, and the metric is more general.
This though requires some other changes, pls see my other comments.

src/unitxt/processors.py
```diff
@@ -75,6 +75,15 @@ def process(self, instance):
         return splitted[0].strip()
 
 
+class ConvertToBoolean(BaseFieldOperator):
+    def process(self, instance):
```
matanor, line 79: where is this new operator used? could you please point me to an example of it? is it in your other PR in FM-Eval?

matanor: Also, please add tests for the new operator..

matanor (changes requested):
@ellarabi looks good! left a few comments, please have a look

src/unitxt/metrics.py
ellarabi: done, we have some extent of control on how the *_robustness datasets are generated, so I refined the field to have multiple references, and adjusted the metric code and the card code

prepare/metrics/string_containment.py
ellarabi: done, please see my comment here https://github.com/IBM/unitxt/pull/394#discussion_r1429117281

src/unitxt/processors.py
ellarabi: the new operator is used in the BoolQ_robustness dataset card in prepare/cards/boolq_robustness.py in fm-eval

added a test for the new processor in tests/test_postprocessors.py

ellarabi:
please note some commits added (not strictly related to my work) due to rebasing the branch to sync with the master

src/unitxt/processors.py
matanor: > the new operator is used in the BoolQ_robustness dataset card in prepare/cards/boolq_robustness.py in fm-eval

[Here](https://github.ibm.com/IBM-Research-AI/fm-eval/blob/robustness-datasets-local/prepare/cards/boolq_robustness.py)?

matanor: oh ok, found it!

src/unitxt/processors.py
```diff
@@ -75,6 +75,15 @@ def process(self, instance):
...
+        if any(w in str(instance).strip().lower() for w in ["yes", "right", "correct"]):
```
matanor, line 80: small thing: i suggest to assign `str(instance).strip().lower()` once into some variable and then use that in the tests. to avoid that being done 6 times (in the worst case)..

src/unitxt/processors.py
```diff
@@ -75,6 +75,15 @@ def process(self, instance):
...
+            return "TRUE"
+        if any(w in str(instance).strip().lower() for w in ["no", "not", "wrong"]):
+            return "FALSE"
+        return instance.upper()
```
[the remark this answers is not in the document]
matanor: oh IIUC its to catch the "true" and "false" lowercase outputs.. maybe a more explicit way would be to add "true" and "false" to the respective lists defined just above ..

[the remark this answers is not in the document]
matanor: Also, a question: values that are not true or false, like "interesting" in your tests, do we want them perhaps converted to some fixed value like "unknown"?

matanor (changes requested):
hi @ellarabi , tks for making the changes, left a few small comments ..

matanor:
> please note some commits added (not strictly related to my work) due to rebasing the branch to sync with the master

I merged from main, so the diff will only show your changes

src/unitxt/processors.py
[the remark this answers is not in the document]
ellarabi: right, the task is essentially a binary classification task, where models answer questions based on passage; the references (the "answer" column) in the attached screenshot, are treated as strings in the (existing) accuracy metric in this case, that's why I'm simply aiming at precisely this set: {"TRUE", "FALSE"}

added "true" and "false" to the list, as you suggested, and changed any other return value to "OTHER", since it obviously won't match one of the {"TRUE", "FALSE"}, and is better readable

image

src/unitxt/processors.py
ellarabi: done

matanor (approved):
LGTM!

PR #396, opened 2023-12-15
expand code coverage and documentation of refiner classes

dafnapension:


elronbandel:
Originally the ides behind stream refiner was to have common base class for all operators that are working on the final stream. meaning doing some refining operation on the resulting stream once everything else was done. 

The stream refiner api was created in order to make it clear for future developers what class they should inherit from if they want to create new operations on the resulting stream that does not change it drastically but just slightly refine the stream. Not entirely sure how to define it better. 

From this perspective the refiner should be an abstract class with no operation. Then we should might have another class called StreamLimiter who has the argument max_instances. Not sure about it. what do you think @yoavkatz @dafnapension?

dafnapension:
I like the idea. Do I understand correctly that at any rate, the refinement is just by discarding instances, and the question is - which ones?
And you would like to see a clear wrap of that indicator, identifier of the ones to be discarded?

In my limited imagination, I see a class that has two apis:  (1) update [the internals of the class] given the instance now examined, (2) give an indication whether or not to discard that instance.

Will that do?

elronbandel (approved):

PR #402, opened 2023-12-17
Enhance FilterByValue to have disallowed_values

elronbandel:


src/unitxt/operators.py
```diff
@@ -1028,27 +1028,70 @@ def process(
...
 class FilterByValues(SingleStreamOperator):
-    """Filters a stream, yielding only instances that match specified values in the provided fields.
+    """Filters a stream, yielding only instances that match specified required values in the provided fields and do not match specified disallowed values.
+
+    Raises an error if a required key is missing.
 
     Args:
-        values (Dict[str, Any]): For each field, the values that instances should match to be included in the output.
+        required_values (Dict[str, Any]): Values that instances must match to be included in the output.
+        disallowed_values (Dict[str, Any]): Values that instances must not match to be excluded from the output.
```
dafnapension, line 1037: must not match to be **included in** the output

Perhaps emphasize with stars, in both lines *must match*  and *must not match*

ilyashnil (approved):

src/unitxt/operators.py
```diff
@@ -1028,27 +1028,70 @@ def process(
...
+            )
+
+    def _is_required(self, instance: dict) -> bool:
+        if not self.required_values:  # If required_values is empty, return True
+            return True
+
+        for key, value in self.required_values.items():
+            if key not in instance:
+                raise ValueError(
+                    f"Required filter field ('{key}') in FilterByValues is not found in {instance}"
```
dafnapension, line 1080: in **required_values** is not found

src/unitxt/operators.py
```diff
@@ -1028,27 +1028,70 @@ def process(
...
+                )
+            if instance[key] != value:
+                return False
+        return True
+
+    def _is_disallowed(self, instance: dict) -> bool:
+        if not self.disallowed_values:  # If disallowed_values is empty, return False
+            return False
+
```
dafnapension, line 1089: no check here whether all field names mentioned in `disallowed_values` also present in `instance`?

dafnapension, line 1089: no `ValueError` here in case a field name from `disallowed_values` is not present in `instance`?

PR #403, opened 2023-12-18
Fix to allow override empty_label of multi_label template.

yoavkatz:
Turns out that if you don't specify the type, you can not override the parameter.


elronbandel (approved):
This is a good fix for now.

In general it looks like a serious issue that should have not happen. 

The unitests for Dataclass should definitely test such cases.

elronbandel:
> Turns out that if you don't specify the type, you can not override the parameter.
Opened:
https://github.com/IBM/unitxt/issues/406

PR #407, opened 2023-12-18
Fix suffix prefix not return same prefix/suffix repeated

yoavkatz:
I missed it in the review, but the behavior was that the same suffix/prefix was repeat multiple times.


elronbandel (approved):

dafnapension:
Hi @yoavkatz ,  and you needed them to be different from one another? each selected randomly separately?

yoavkatz:
@dafnapension  - Sorry if the specification was unclear.  It was intended for have random repetition of each suffix/prefix option.

PR #410, opened 2023-12-18
Adding simple datasets

ilyashnil:


ilyashnil:
There are few open  issues:
1. There is no multi_label template (fix required to unfair_tos and reuters)
2. Can I use text_type : argument?
3. I wonder if dbpedia_14 is of type text or paragraph
4. I do not remember why I need to add ListFieldValues(fields=["label"], to_field="label") , maybe it's not relevant anymore.
5. There are some datasets where we played with splits. E.g. law_stack_exchange.py in we swapped train and test. Is it ok to do it? Or in financial_tweets validation went to test.

elronbandel (approved):

PR #414, opened 2023-12-19
Independent random generators

matanor:
Remove existing randomization utilities.
Replace them with independent random generators, that depend either on a default fixed seed, and a parametrized sub seed. The sub seed is determined per use case, in some cases it is set to a fixed value, and in other cases it depends on an input value. For example, an augmented text may be used as a seed, before making a random choice required for the actual augmentation.


matanor:
hi @elronbandel  @yoavkatz , please see the changes for removing the current randomization mechanism from unitxt. I replaced the randomization with independent random generators that depend either on a fixed sub seed, or a sub seed that depends on the input.

Some other things that i think may be removed, wasn't 100% sure about them"

[RandomPicker](https://github.com/IBM/unitxt/pull/414/files#diff-e34f365eb90b5d90ea06cc6c72c392defef92ac86775846adec94b782fcacfb1L54) - haven't seen any use of it, can it be removed?

[test_thread_safety_sanity](https://github.com/IBM/unitxt/pull/414/files#diff-60e4afdf255b8ba453bb1db436dc007517d395f33a57ba2294d483ecc4f206e6R46), and [test_thread_safety](https://github.com/IBM/unitxt/pull/414/files#diff-60e4afdf255b8ba453bb1db436dc007517d395f33a57ba2294d483ecc4f206e6R64) - can be removed? since we have no special threading handling of a per-thread seed.

src/unitxt/collections.py
```diff
@@ -52,9 +53,13 @@ def __call__(self, collection: Collection):
 
 
 class RandomPicker(Artifact):
+    random_generator: random.Random = field(
+        default_factory=lambda: new_random_generator(sub_seed="random_picker")
+    )
+
```
elronbandel, line 59: Nice! I like the readability of this!

src/unitxt/fusion.py
```diff
@@ -89,10 +89,13 @@ def fusion_generator(self, split) -> Generator:
         weights = copy.deepcopy(self.weights)
         iterators = [iter(origin()[split]) for origin in self.origins]
         total_examples = 0
+        random_generator = new_random_generator(sub_seed="weighted_fusion")
```
elronbandel, line 92:
```suggestion
        random_generator = new_random_generator(sub_seed="weighted_fusion_" + split)
```
I think its better to have different seeds for different splits

src/unitxt/operators.py
```diff
@@ -518,11 +515,12 @@ def process_value(self, value: Any) -> Any:
         words = re.split(r"(\s+)", value)
         new_value = ""
 
+        random_generator = new_random_generator(sub_seed=str(hash(value)))
```
elronbandel, line 518: This means every value will have exactly the same augmentation. is that what we want?

src/unitxt/operators.py
```diff
@@ -648,7 +653,8 @@ class ShuffleFieldValues(FieldOperator):
 
     def process_value(self, value: Any) -> Any:
         res = list(value)
-        get_random().shuffle(res)
+        random_generator = new_random_generator(sub_seed=str(hash(str(res))))
```
elronbandel, line 656: Same comment here it will have exactly the same augmentation per given value.

src/unitxt/operators.py
elronbandel: Another option for example, is that we will use the entire instance to determine the seed. Then it give some more freedom.

src/unitxt/splitters.py
```diff
@@ -108,9 +108,7 @@ class RandomSampler(Sampler):
 
     def prepare(self):
         super().prepare()
-        self.random_generator = get_sub_default_random_generator(
-            sub_seed="random_sample_seed"
-        )
+        self.random_generator = new_random_generator(sub_seed="random_sample_seed")
```
elronbandel, line 111: This is in the general score of the operator which means:
1) Calling the same operator twice in a row will affect each other and won't be identical
2) Calling different streams from the operators in different order will affect its behavior per that stream

src/unitxt/splitters.py
```diff
@@ -151,10 +149,12 @@ class DiverseLabelsSampler(Sampler):
 
     choices: str = "choices"
     labels: str = "labels"
+    random_generator: Random = None
 
     def prepare(self):
         super().prepare()
         self.labels_cache = None
+        self.random_generator = new_random_generator(sub_seed="diverse_labels_sampler")
```
elronbandel, line 157: Same issue here with having the generator in the class scope.

src/unitxt/templates.py
```diff
@@ -360,7 +360,8 @@ def outputs_to_target_and_references(self, outputs: Dict[str, object]) -> List[s
             )
 
         if self.random_reference:
-            target = get_random().choice(references)
+            random_generator = new_random_generator(str(references))
```
elronbandel, line 363: Do we want dependency on the references or the entire instance?

elronbandel (changes requested):
Overall great job.  I think we should always notice what is the scope of the random generator and make sure it is not affected by consecutive calls of different split orders (e.g calling the train then the test vs calling the test then train).

src/unitxt/fusion.py
matanor: fixed

src/unitxt/splitters.py
matanor: oh I thought there was some level of independence here with the generator is reinitialized in `prepare()`.
when is `prepare` called?

the alternative is having the generator initialized in sample(), right? then it can depend on the `instances_pool`, but isn't that pool fixed?

src/unitxt/templates.py
matanor: updated so it depends on the entire `outputs`. is this what you meant?

src/unitxt/operators.py
matanor: So i can create a generator based on the entire instance in `Augmentor.process()`, and pass it here to `process_value`. sounds good?

elronbandel: that's great

src/unitxt/splitters.py
elronbandel: The prepare is called in the instantiation which means once in a class life time. 

The rest lets discuss

src/unitxt/operators.py
yoavkatz: There is a tradeoff - I think we should augment per value.  Indeed it means the same text will always augmented the same (given the global seed is the side). However, if we don't do this  adding a single element/removing to the list, will change the augmentation of all values.

src/unitxt/operators.py
yoavkatz: Same comment as above for me too.  Do we want an additional element in the list to change how all elements are shuffled?

src/unitxt/operators.py
matanor: @elronbandel and i discussed, will keep as is (seed depends on value)

src/unitxt/operators.py
matanor: same decision, keep dependence on the value
@yoavkatz was your intention that even depending on the entire list is too much?

src/unitxt/splitters.py
matanor: fixed

src/unitxt/splitters.py
matanor: fixed

matanor:
@yoavkatz @elronbandel thanks for your reviews. i think its ready for another round. 

@elronbandel I will soon try to add a test for what we discussed.

Also, please see these two questions:

1. [RandomPicker](https://github.com/IBM/unitxt/pull/414/files#diff-e34f365eb90b5d90ea06cc6c72c392defef92ac86775846adec94b782fcacfb1L54) - haven't seen any use of it, can it be removed?

2. [test_thread_safety_sanity](https://github.com/IBM/unitxt/pull/414/files#diff-60e4afdf255b8ba453bb1db436dc007517d395f33a57ba2294d483ecc4f206e6R46), and [test_thread_safety](https://github.com/IBM/unitxt/pull/414/files#diff-60e4afdf255b8ba453bb1db436dc007517d395f33a57ba2294d483ecc4f206e6R64) - can be removed? since we have no special threading handling of a per-thread seed.

elronbandel (approved):

elronbandel:
You can delete those tests as they are no longer relevant 👍

Issue #334, opened 2023-11-27
A formatter based on format strings


matanor:
The current formatter [ICLFormat](https://github.com/IBM/unitxt/blob/main/src/unitxt/formats.py#L12) expresses the used format implicitly, using several sepsrators added in between the template and the demos:

```
class ICLFormat(SizeLimitingFormat):
    prefix: str = ""
    input_prefix: str = ""
    output_prefix: str = ""
    target_prefix: str = " "
    instruction_prefix: str = ""
    input_output_separator: str = "\n"
    demo_separator: str = "\n\n"
    suffix: str = ""
```

A more explicit formatter could be based on two fields only, one for the instructions format, and another for the demos format.
For example,
instructions_format = 
"|system message|

{demos}

{instruction}
"
where the demos are created in a loop, each demo formatted using a `demo_format`, or using the template, and the instruction is also produced by the template.

Generally, there will be several fixed field names that may be integrated into the formats, and one or two format fields that control how the output looks.

matanor:
Suggestion to define two patterns:

full_pattern:
"
<inst>

{example_pattern X num_demos}

{example_pattern}

</inst>
"
------------
example_pattern:
"
user
<template input>
agent
<template output>
"

PR #423, opened 2023-12-22
create instance operator SystemFormat and make BasicRecipe all operators, no renderers

dafnapension:
Closes https://github.com/IBM/unitxt/issues/334


src/unitxt/operators.py
```diff
@@ -179,6 +180,38 @@ def process(
         return flatten_dict(instance, parent_key=self.parent_key, sep=self.sep)
 
 
+class WholeInputFormatter(StreamInstanceOperator):
+    """Generates the whole input to the model, from constant strings that are given as args, and from values found in specified fields of the instance.
```
elronbandel, line 184: the general_format and input_output_format will be given as args. 
`demos`,` instruction`, `system_prompt` `source` and `target` will be in the instance

tests/test_formats.py
```diff
@@ -35,7 +35,7 @@ def test_icl_format_with_demonstrations(self):
         self.assertEqual(result, target)
 
     def test_icl_format_with_demonstrations_and_instruction_after_demos(self):
```
elronbandel, line 37: once changed to operator it will be better if icl format will be tested with the operator testing function

elronbandel (commented):
Great start! see the comments which are really minor changes. so far in the right direction!

src/unitxt/operators.py
```diff
@@ -179,6 +179,127 @@ def process(
...
+        overall_output_size_limit: int = None
+    """
+
+    to_field: str
+    demos_field: str = None
+    system_intro: str = None
+    system_closing: str = None
+    input_prefix: str = ""
+    output_prefix: str = ""
+    target_prefix: str = " "  # only for compatibility with ICLFormat.  surely sure?
```
dafnapension, line 221: @elronbandel , please note the exceptional one space here

tests/test_formats.py
dafnapension: Done, Comparing both results.

src/unitxt/operators.py
dafnapension: I interpreted `system_prompt` as `prefix` and `suffix` of ICLformat, and `input_output_format` as all those small pieces. I tried to combine some, but the flow of the code in ICLformat does not allow to reduce the **number** of formatting args.

yoavkatz:
@dafnapension @elronbandel - I think the idea by @matanor in #334 , is that this collection of parameters is unwieldy for the formatter.  

    to_field: str
    demos_field: str = None
    system_intro: str = None
    system_closing: str = None
    input_prefix: str = ""
    output_prefix: str = ""
    target_prefix: str = " "  # the single space here is only for compatibility with ICLFormat.  surely sure?
    instruction_prefix: str = ""
    input_output_separator: str = "\n"
    demo_separator: str = "\n\n"
    add_instruction_at_start: bool = True
    add_instruction_after_demos: bool = False
    overall_output_size_limit: int = None
    
    
Instead the idea is to have a template where do user puts the separators explicitly.

src/unitxt/operators.py
```diff
@@ -213,6 +213,97 @@ def process(
         return flatten_dict(instance, parent_key=self.parent_key, sep=self.sep)
 
 
+class WholeInputFormatter(StreamInstanceOperator):
```
yoavkatz, line 216: Maybe call it -  ModelInputFormatter?  This is inlined in other places in the code, where we call the "whole" input "model input".  E.g. in augmenters.

src/unitxt/operators.py
```diff
@@ -213,6 +213,97 @@ def process(
...
+
+    Args:
+        to_field (str): the name of the field into which the formatted model's input is to be stored
+        demos_field (str): the name of the field that contains the demos, being dicts with "source" and "target" keys
+        system_prompt (str): the name of the field containing an introductory text and or xml tags
+        demo_format (str): formatting string for a single demo, combining fields source and target
+        general_format (str) overall output format, combinig system_prompt, instruction, demos, and source
+    """
+
+    to_field: str
```
yoavkatz, line 241: default to "model_input"?

src/unitxt/operators.py
```diff
@@ -213,6 +213,97 @@ def process(
...
+    formatting arguments.
+
+    Args:
+        to_field (str): the name of the field into which the formatted model's input is to be stored
+        demos_field (str): the name of the field that contains the demos, being dicts with "source" and "target" keys
+        system_prompt (str): the name of the field containing an introductory text and or xml tags
+        demo_format (str): formatting string for a single demo, combining fields source and target
+        general_format (str) overall output format, combinig system_prompt, instruction, demos, and source
+    """
+
```
yoavkatz, line 240: Important to add examples...

src/unitxt/operators.py
```diff
@@ -213,6 +213,97 @@ def process(
...
+    to_field: str
+    demos_field: str = None
+    system_prompt: str = None
+    demo_format: str = (
+        "{source}\n{target}\n\n"  #  example: "User: {source}\nAgent: {target}\n\n"
+    )
+    general_format: str = (
+        "<SYS>{system_prompt}</SYS>\n{instruction}\n{demos}\n{source}\n"
+    )
+    instruction_prefix: str = ""  # potential prepend for non "" instruction. necessary for backward compatibility with iclformat
```
yoavkatz, line 250: Not needed, in my view.

src/unitxt/operators.py
```diff
@@ -213,6 +213,97 @@ def process(
...
+    Args:
+        to_field (str): the name of the field into which the formatted model's input is to be stored
+        demos_field (str): the name of the field that contains the demos, being dicts with "source" and "target" keys
+        system_prompt (str): the name of the field containing an introductory text and or xml tags
+        demo_format (str): formatting string for a single demo, combining fields source and target
+        general_format (str) overall output format, combinig system_prompt, instruction, demos, and source
+    """
+
+    to_field: str
+    demos_field: str = None
```
yoavkatz, line 242: Default to "demos".

yoavkatz:
Thanks @dafnapension .  Once it stabilizes, can you try to move it so that ICLFormat is deprecated, and all existing formats are moved to this new form?

It will also make sure, we covered all the requirements.

src/unitxt/operators.py
dafnapension: I had a problem with ICLFormat, with which I wanted to keep backward compatibility.
ICLFormat does prepend arg `instruction_prefix` (, e.g. "Instruction: ") only when: (1) the instruction comes before the demos, not after, **and** (2) the instruction is not "".   Now, if I move this "Instruction: " to be part of `general_format`,   "Instrucion: " will show also when an instance does not have instruction, namely: in cases where instruction is "".

src/unitxt/operators.py
dafnapension: Yes, renamed.

src/unitxt/operators.py
dafnapension: renamed `general_format `-> `model_input_format`.

src/unitxt/operators.py
dafnapension: they are very long.. please see in `test_operators` . Shall I copy over such long examples?

src/unitxt/operators.py
dafnapension: changed.

dafnapension:
Yes.  I suggest to make that replacement in a different PR.

src/unitxt/operators.py
```diff
@@ -213,6 +213,103 @@ def process(
         return flatten_dict(instance, parent_key=self.parent_key, sep=self.sep)
 
 
+class ModelInputFormatter(StreamInstanceOperator):
+    r"""Generates the whole input to the model, from constant strings that are given as args, and from values found in specified fields of the instance.
+
+    WholeInputFormatter expects its input instance to have been processed by a Template StreamInstanceOperator
```
matanor, line 219:
```suggestion
    ModelInputFormatter expects its input instance to have been processed by a Template StreamInstanceOperator
```

src/unitxt/operators.py
```diff
@@ -213,6 +213,103 @@ def process(
...
+
+    Args:
+        demos_field (str): the name of the field that contains the demos, being dicts with "source" and "target" keys
+        system_prompt (str): the name of the field containing an introductory text and or xml tags
+        demo_format (str): formatting string for a single demo, combining fields source and target
+        model_input_format (str) overall output format, combining system_prompt, instruction, demos, and source
+    """
+
+    demos_field: str = "demos"
+    system_prompt: str = None
```
matanor, line 243:
```suggestion
    system_prompt_field: str = None
```

matanor: since its a field name, AFAIU

src/unitxt/operators.py
```diff
@@ -213,6 +213,103 @@ def process(
         return flatten_dict(instance, parent_key=self.parent_key, sep=self.sep)
 
 
+class ModelInputFormatter(StreamInstanceOperator):
+    r"""Generates the whole input to the model, from constant strings that are given as args, and from values found in specified fields of the instance.
```
matanor, line 217: I think the docstring should be simplified and focused on the required inputs and produces outputs. Other things such as description of the works of other parts of unitxt should be ommited as much as possible, IMO. Here is my sketchy attempt, please consider something along these lines:

ModelInputFormatter expects the input instance to contain: 
    1. A "source" field with the source dataset instance values represented by a single string, in the context of the underlying task. 
    2. A "instruction" field containing a non-None value.
    3. A field named with the value in `self.demos_field`, containing a list of dicts, each dict with "source" and "target" field representing a single demo.
    4. A field named with the value in `self.system_prompt_field`
    
ModelInputFormatter formats the above fields into a single string inputted to the model. This string overwrites the "source" field of the instance. 
ModelInputFormatter also pops fields "instruction" and "demos" from the input instance.

Args:
    demos_field (str): the name of the field that contains the demos, being dicts with "source" and "target" keys
    system_prompt_field (str): the name of the field containing an introductory text and or xml tags
    demo_format (str): formatting string for a single demo, combining fields source and target
    model_input_format (str) overall output format, combining system_prompt, instruction, demos, and source

src/unitxt/operators.py
```diff
@@ -213,6 +213,103 @@ def process(
...
+        demo_format (str): formatting string for a single demo, combining fields source and target
+        model_input_format (str) overall output format, combining system_prompt, instruction, demos, and source
+    """
+
+    demos_field: str = "demos"
+    system_prompt: str = None
+    demo_format: str = (
+        "{source}\n{target}\n\n"  #  example: "User: {source}\nAgent: {target}\n\n"
+    )
+    model_input_format: str = (
```
matanor, line 247: Why are the `system_prompt` and `instruction` fields of the instance and not explicit parts of this parameter?

And then the formatter could be used by setting `model_input_format` to (for example):

"<SYS>whaterver the system prompt is</SYS>\n whatever additional instructions there are \n{demos}\n{source}\n"

In other words, i would opt for minimizing the things that are passed by the instance to a bare minimum.

src/unitxt/operators.py
```diff
@@ -213,6 +213,103 @@ def process(
...
+        self, instance: Dict[str, Any], stream_name: Optional[str] = None
+    ) -> Dict[str, Any]:
+        assert (
+            "source" in instance
+        ), f"field 'source' is expected to be in the input instance. Received instance: {instance}"
+        source = self._retrieve_field_and_assert_not_none(
+            instance=instance, field_name="source"
+        )
+
+        system_prompt = self._retrieve_field_and_assert_not_none(
```
matanor, line 272: Repeating my comment above, why are `system_prompt` and `instruction` fields of the instance and not explicit parts of the `model_input_format` parameter?

matanor (dismissed):
hi @dafnapension @yoavkatz 
thanks Dafna for creating this new formatter! looks good, added a few comments.
My main question pertains to why are the `system_prompt` and `instruction` fields passed in the instance?

src/unitxt/operators.py
dafnapension: This is, as far as I understand, because this operator is expected to follow other operators, one of which prepares the instruction, another one prepares the template (btw, Elron plans to combine these two into one).

src/unitxt/operators.py
dafnapension: Checking and testing now, `system_prompt` seems an orphan field. No one uses it, and if @yoavkatz agrees that indeed orphan, I will remove that arg altogether.

src/unitxt/operators.py
yoavkatz: I mean an example like this:

"demo_format": "Human: {source}\nAssistant:  {target}\n\n",
    "model_input_format": "{system_prompt}{instruction}\n{demos}\nHuman: {source}\nAssistant: "
    
    Will yield output like:
    
    
    ...

src/unitxt/operators.py
yoavkatz: There is some terminology mismatch now.   

System prompt - is what us today called instruction.

The "Instruction" which should be coming from the template (like source and target) , is not yet implemented.

dafnapension: I see. Removing `system_prompt`    והיה כלא היה

src/unitxt/operators.py
dafnapension: Yes. Starting with removing `system_prompt` altogether.
The real things: `source` and `instruction`, and `template` (the last two separated or joint), and the `demos`, are expected to live in the instance from their birth to the creation of this `model input`, at which time they are popped from the instance. Part of the evolution of the stream.

src/unitxt/operators.py
dafnapension: Added example. Indeed not terribly long.

src/unitxt/operators.py
matanor: @dafnapension @yoavkatz I would separate between things that depend on the data and things that don't, hence IMO having `demos` on the instance makes sense.

However for the `instruction`, i am not sure. Since its a fixed value for all instances, i think it should be passed directly to the formatter (or mega operator) as a parameter. As for `template`, haven't seen that in your code, i think? Is it another field set on the instance?

src/unitxt/operators.py
yoavkatz: This is an error and indeed needs to be removed.

src/unitxt/operators.py
yoavkatz: I see.  Thanks for clarifying.  I think today "Instruction" is really used for system prompt, and in this case, there is now need to add "Instruction:".  It can be explicitly done.

src/unitxt/operators.py
```diff
@@ -213,6 +213,109 @@ def process(
...
+
+        instruction = self._retrieve_field_and_assert_not_none(
+            instance=instance, field_name="instruction"
+        )
+        # pop from instance, as ICLFormat does
+        if "instruction" in instance:
+            instance.pop("instruction")
+
+        if instruction != "":
+            instruction = instruction + "\n\n"
```
matanor, line 293: Can we remove this if? Having this type of formatting choices, should be avoided i think. If the instruction is to be followed by two new lines, then i think whoever created should be responsible to add it. 

This is regardless of my general comment about instruction coming from the instance. BTW, do you know which part of the code places the `instruction` field on the instance?

src/unitxt/renderers.py
```diff
@@ -34,50 +34,26 @@ def process(
...
 
     steps: List[Operator] = InternalField(default_factory=list)
 
     def prepare(self):
         self.steps = [
             self.template,
             RenderDemonstrations(template=self.template, demos_field=self.demos_field),
-            RenderInstruction(instruction=self.instruction),
-            RenderFormat(format=self.format, demos_field=self.demos_field),
+            AddFields(
```
matanor, line 49: Why set the instruction here on the instance, and not have it as a parameter to ModelInputFormatter?

src/unitxt/standard.py
```diff
@@ -153,19 +158,26 @@ def prepare(self):
...
-            demos_field=self.demos_field,
+        # the 4 constituents of StandardRenderer:
+        self.steps.append(self.template)
+        self.steps.append(
+            RenderDemonstrations(template=self.template, demos_field=self.demos_field)
         )
-
-        self.steps.append(render)
+        self.steps.append(
+            AddFields(
```
matanor, line 167: Same question as above: Why set the instruction here on the instance, and not have it as a parameter to ModelInputFormatter?

prepare/formats/user_agent.py
```diff
@@ -1,9 +1,9 @@
 from src.unitxt.catalog import add_to_catalog
-from src.unitxt.formats import ICLFormat
+from src.unitxt.operators import ModelInputFormatter
 
-format = ICLFormat(
-    input_prefix="User: ",
-    output_prefix="Agent: ",
+format = ModelInputFormatter(
+    demo_format="User: {source}\nAgent:  {target}\n\n",
```
yoavkatz, line 5: Let's remove the extra space before the {target} (what used to be the target_prefix). It's was a mistake.

src/unitxt/operators.py
dafnapension: OK. Now that we all saw that `ModelInputFormatter` can do anything `ICLFormat` could do thus far, in all cases and tests; even to imitate this error, I will remove `ICLFormat` altogether. At this moment, it only lives in tests, just to prove that the same targets it was expecting, also `ModelInputFormatter` achieves. And on this occasion, @yoavkatz , I will fix that said error.

src/unitxt/operators.py
yoavkatz: Thanks. This format is a bit odd because the instruction appears after the demos. 

 model_input_format="{demos}User: {instruction}{source}\nAgent: ",
 
In general, we need to clarify the confusion between the instruction and system prompt.  
Today, it's a mess.  The input parameter to the standard recipe is called instruction, but it is used to pass the system prompt.  In the unit test, instructions are task instructions like "solve the math exercisers".

Let's discuss it with Elron, but for now , let's give this example:
demo_format="Input: {source}\nOutput: {target}\n\n",
 model_input_format="Instruction: {instruction}\n{demos}Input: {source}\nOutput: "

src/unitxt/operators.py
yoavkatz: I agree. No implicit formatting should be done.

src/unitxt/standard.py
yoavkatz: The formatter was already created (usually fetched from the catalog), so we can not pass the instruction in the constructor.   We have two options, add a method update_instruction() to update the object, or pass the instruction in the input stream (as is done now).  Note these "instructions" are really the system prompt.

prepare/formats/user_agent.py
dafnapension: Yes, now as `ICLFormat` gets down the stage, l will remove its heritage as well. Please see my reply [here](https://github.com/IBM/unitxt/pull/423#discussion_r1437459745), @yoavkatz

src/unitxt/operators.py
dafnapension: OK. changed the example.

src/unitxt/operators.py
dafnapension: No problem. Now that we owe nothing to `ICLFormat`, we can do anything we want (please see my reply [here](https://github.com/IBM/unitxt/pull/423#discussion_r1437459745))

src/unitxt/renderers.py
dafnapension: @yoavkatz , I will gladly remove `StandardRenderer`, along with `ICLFormat`. @elronbandel told me that the mission is to clean from renderers, and replace them by operators.

src/unitxt/operators.py
dafnapension: Hi @matanor , `template` is a stream operator, assumed to be applied before our  `ModelInputFormatter`, and, among others, prepare the demo records, with their expected fields `source` and `target`

src/unitxt/standard.py
matanor: The former is more explicit, hence preferred IMO.

Generally, not being able to pass params to artifacts is a limitation .. what if we were just creating the formats with some function, and calling that from the card?

src/unitxt/operators.py
matanor: oh ok, thanks! i was confused, thought its another field on the instance ...

prepare/formats/empty_input_output_separators.py
```diff
@@ -1,6 +1,9 @@
 from src.unitxt.catalog import add_to_catalog
-from src.unitxt.formats import ICLFormat
+from src.unitxt.operators import ModelInputFormatter
 
-format = ICLFormat(input_output_separator="")
+format = ModelInputFormatter(
+    demo_format="{source}{target}\n\n",
+    model_input_format="{instruction}\n{demos}\n{source}",
```
yoavkatz, line 6: This adds 3 \n after the last demo, and two \n after the instruction if we don't have demos.

Maybe it should be :

demo_format="{source}{target}\n\n",
model_input_format="{instruction}\n{demos}{source}",

We need to image the output without demos.

prepare/formats/human_assistant.py
```diff
@@ -1,8 +1,9 @@
 from src.unitxt.catalog import add_to_catalog
-from src.unitxt.formats import ICLFormat
+from src.unitxt.operators import ModelInputFormatter
 
-format = ICLFormat(
-    input_prefix="Human: ", output_prefix="Assistant: ", input_output_separator="\n"
+format = ModelInputFormatter(
+    demo_format="Human: {source}\nAssistant: {target}\n\n",
+    model_input_format="{instruction}\n{demos}\nHuman: {source}\nAssistant: ",
```
yoavkatz, line 6: Same issue as above.  Remove \n after {demos}

prepare/formats/empty_input_output_separators.py
yoavkatz: Now we also need to consider what happens if instruction is empty.   In this case we don't want the \n after it.

We need some to note that \n is only if the instruction is not empty.

{instruction+\n}{demos}{source}

such that if we find {XXX+something} we replace XXX with the value and append the something.

Do you have other ideas?

dafnapension: no three \n any longer, because I removed the automatic appending of a non empty instruction with \n\n.  If we are talking about a fix prefix, like "Instruction: ", then I can certainly prepend with it, and append any non-"" instruction with one or two \n.

prepare/formats/human_assistant.py
dafnapension: removed the \n after demos. But this is a different issue. it does not relate to the instruction.

src/unitxt/operators.py
dafnapension: Hi @matanor ,  `system_prompt` no longer exists. At least from the few examples and tests that I tried, Indeed no need to have the instruction string taking that short ride on the recipe, but it better be passed as an input arg to the constructor of `ModelInputFormatter`.  @elronbandel asked that I treat instruction as a resident of the input instance.  I will speak with him about the reason for that.

src/unitxt/operators.py
dafnapension: Thanks @matanor , I polished the docstring as you elegantly suggested.

prepare/formats/empty_input_output_separators.py
elronbandel: I think this should be like the demo format. so you have an `instruction_format` argument which can help modifying it just like the `demo_format`.

elronbandel (commented):
I want to suggest to changes to the name to reflect (1) it can be any system. (2) to fit the naming convention of the template. 

Hence, the name should simply be imho: `SystemFormat`

prepare/formats/empty_input_output_separators.py
dafnapension: I thought that the aim is to reduce the number of formatting arguments. And since instruction is a simple string, why not, indeed, let the user wrap it with any formatting (prefix and or suffix) as they want?
I think that more important (or that I less understand) is the question of how to push that instruction (in my view: a whole formatted instruction) into the final result: by using `AddFields` a couple of Operation up the recipe (as is done now, but makes @matanor feel uncomfortable) or have an api to update `ModelInputFormatter`. I understood from @yoavkatz that an input argument is not possible once the Operator is a formal, catalogged, artifact.

dafnapension:
> I want to suggest to changes to the name to reflect (1) it can be any system. (2) to fit the naming convention of the template.
> 
> Hence, the name should simply be imho: `SystemFormat`

@elronbandel , you refer to the name of the Operator (i.e. `ModelInputFormatter` -> `SystemFormat`)? or you refer to one of the input arguments?  (`model_input_format`?)

elronbandel:
The name of the operator `ModelInputFormatter` -> `SystemFormat`

dafnapension:
> The name of the operator `ModelInputFormatter` -> `SystemFormat`

Done.

src/unitxt/standard.py
```diff
@@ -29,7 +34,7 @@ class BaseRecipe(Recipe, SourceSequentialOperator):
     card: TaskCard
     template: Template = None
     instruction: Instruction = None
-    format: ICLFormat = ICLFormat()
+    format: SystemFormat = SystemFormat()
```
dafnapension, line 38: @elronbandel , I add the comment here but this refers to `verify` method, that I have not changed, hence Github would not allow to add comments to. a couple of lines down.
It seems that (in some cases) `ValueError` is raised DAVKA when the desired condition, as described in the exception text, **is** satisfied. Shall I swap the condition, or the text?. In most cases, the text seems to me correct (up to the pettiness with the == case), but I wanted to check with you.

elronbandel (approved):

PR #424, opened 2023-12-24
Binary recall and precision and more advanced filtering operators.

lilacheden:
Add operators and metrics needed for kpa evaluation


src/unitxt/operators.py
```diff
@@ -1092,6 +1093,31 @@ def _is_disallowed(self, instance: dict) -> bool:
         return False
 
 
+class FilterByOrder(SingleStreamOperator):
```
elronbandel, line 1096: Can you add a short description for this operator with few examples?

elronbandel: Also if you can add a common base class for this class and FilterByValue

elronbandel (changes requested):
Really great PR thanks for the contribution! refer to the minor comments I made and we can merge once the few little changes are made.

matanor (changes requested):
For the metrics, if viable, i would suggest a different approach:

1. I think the shift from the string output ("yes"/"no"/other wrong stuff) to the metric input (0s or 1s) should be with a post-processor. Meaning, have a post-processor that  changes the outputs to 0s and 1s, and then the metric just works on 0s and 1s.

2. With 1 done, we can reuse `HuggingfaceMetric` - have `BinaryPrecision` and `BinaryRecall` inherit from it directly, without needing `BinaryMetric` (also, should we call them following their HuggingFace equivalents, `Recall` and `Precision`?)

elronbandel (approved):

yoavkatz (dismissed):
Thanks for integrating the changes and adding new capabilities.  However, I think the interface to the binary recall/precision metrics should be same as in the the other metrics of binary , so it can be added to all cards that are binary classification.

Today, binary classification is defined as:

add_to_catalog(
    FormTask(
        inputs=["text", "text_type", "class"],
        outputs=["class", "label"],
        metrics=[
            "metrics.accuracy",
        ],
        augmentable_inputs=["text"],
    ),
    "tasks.classification.binary",
    overwrite=True,
)

These metrics expect a list which is either empty [] , or the name of the class (e.g. ['Termination of convenience']) 

Now one can argue if this is the right representation for binary - but consistency is important.

For now I suggest we stick to this approach.  If we to redefine the inputs/outputs of the binary task, it best done in a separate issue.

src/unitxt/operators.py
```diff
@@ -1092,6 +1104,49 @@ def _is_disallowed(self, instance: dict) -> bool:
         return False
 
 
+class FilterByOrder(BaseFilter):
```
yoavkatz, line 1107: I think our use of filters is becoming complex, and we are creating a new "programming language" .
Still for this PR, I think it's ok for now to add this.
I suggest calling call it FilterByCondition and not FilterByOrder.

 FilterByCondition(required_values = {"a":4}, condition = "gt") will yield only instances where "a">4

yoavkatz: Now that I think of it maybe have
FilterByValues just be a special case , where condition is "operator.eq"

matanor:
> Thanks for integrating the changes and adding new capabilities. However, I think the interface to the binary recall/precision metrics should be same as in the the other metrics of binary , so it can be added to all cards that are binary classification.
> 
> Today, binary classification is defined as:
> 
> add_to_catalog( FormTask( inputs=["text", "text_type", "class"], outputs=["class", "label"], metrics=[ "metrics.f1_micro_multi_label", "metrics.f1_macro_multi_label", "metrics.accuracy", ], augmentable_inputs=["text"], ), "tasks.classification.binary", overwrite=True, )
> 
> These metrics expect a list which is either empty [] , or the name of the class (e.g. ['Termination of convenience'])
> 
> Now one can argue if this is the right representation for binary - but consistency is important.
> 
> For now I suggest we stick to this approach. If we to redefine the inputs/outputs of the binary task, it best done in a separate issue.

Not sure it matters, but KPA will probably be a new task anyway, cause it accepts two input texts. So if defining a new task, we can consider a new output format, meaning, not a class name and an empty list, but rather two fixed values.

src/unitxt/operators.py
lilacheden: Changed to FilterByCondition. 
FilterByValues has both required and disallowed values, so it will require two consecutive FilterByCondition instances.

lilacheden:
> Thanks for integrating the changes and adding new capabilities. However, I think the interface to the binary recall/precision metrics should be same as in the the other metrics of binary , so it can be added to all cards that are binary classification.
> 
> Today, binary classification is defined as:
> 
> add_to_catalog( FormTask( inputs=["text", "text_type", "class"], outputs=["class", "label"], metrics=[ "metrics.f1_micro_multi_label", "metrics.f1_macro_multi_label", "metrics.accuracy", ], augmentable_inputs=["text"], ), "tasks.classification.binary", overwrite=True, )
> 
> These metrics expect a list which is either empty [] , or the name of the class (e.g. ['Termination of convenience'])
> 
> Now one can argue if this is the right representation for binary - but consistency is important.
> 
> For now I suggest we stick to this approach. If we to redefine the inputs/outputs of the binary task, it best done in a separate issue.

I added PrecisionMultiLabel and Recall multi label the inherit from F1MultiLabel (whose metric was changed to a parameter), so now it gives the same global results as in the solution I suggested. However, as Matan wrote,  the kpa task must be different than the binary classification task, even if it has the same metrics

yoavkatz:
@lilacheden - Thanks for making the more general change to the metric. One of the goal of unitxt is to allow more consistent reuse across datasets .

I agree the KPA matching task is not a standard classification task .  It's sort of like entailment.  I would make it different, but as consistent with other binary classification tasks as possible.

```
{
    "type": "form_task",
    "inputs": [
        "text",
        "text_type".  // E.g. "review", "survey"
        "keypoint",
        "class" // ['matched']
    ],
    "outputs": [
        "label". // can be [] or ['matched']
    ],
    "metrics": [
        "metrics.precision_binary"  <--  suggest using this name for consistency
         "metrics.recall_binary"  <-- suggest using this name for consistency
         "metrics.f1_binary"    <-- suggest  adding this one (just as an alias to micro_f1_multi_label)
    ]
}

**Comment by yoavkatz** [comment]
Thinking about it more.  Maybe we should see KPA more of an extraction task like NER.
 


In NER we have the processed predictions  (entity-surface-form, entity-type ) like:

[("Amir", "Person"), ("Yaron", "Person")]
 [("Ran", "Person"), ("Yonatan", "Person")]
 
and output:

   "f1_Person": 0.57,
    "f1_macro": 0.57,
    "in_classes_support": 1.0,
    "f1_micro": 0.57,
    "recall_micro": 0.67,
    "recall_macro": 0.67,
    "precision_micro": 0.5,
    "precision_macro": 0.5,
    "score": 0.57,
    "score_name": "f1_micro",
    "f1_micro_ci_low": 0.57,
    "f1_micro_ci_high": 0.57,
    "score_ci_low": 0.57,
    "score_ci_high": 0.57,

I think this is exactly the same metric used for NER (see unitxt/prepare/metrics/custom_f1.py)

I think all we need to do for KPA, is create a metric that maps the prediction  and keypoint into this format and calls the current metric.

So if the model predicted   "I like my canine friend"  is matched to "I like dogs", we will convert it to:
 
( "matched", "I like dogs" )

And if not, we will convert it to:

( "not matched", "I like dogs")
  

 Then we will get micro, macro f1 and precision recall over the key points - with one call.

**Comment by matanor** [comment]
> Thinking about it more. Maybe we should see KPA more of an extraction task like NER.
> 
> In NER we have the processed predictions (entity-surface-form, entity-type ) like:
> 
> [("Amir", "Person"), ("Yaron", "Person")] [("Ran", "Person"), ("Yonatan", "Person")]
> 
> and output:
> 
> "f1_Person": 0.57, "f1_macro": 0.57, "in_classes_support": 1.0, "f1_micro": 0.57, "recall_micro": 0.67, "recall_macro": 0.67, "precision_micro": 0.5, "precision_macro": 0.5, "score": 0.57, "score_name": "f1_micro", "f1_micro_ci_low": 0.57, "f1_micro_ci_high": 0.57, "score_ci_low": 0.57, "score_ci_high": 0.57,
> 
> (see unitxt/prepare/metrics/custom_f1.py)
> 
> I think all we need to do for KPA, is create a metric that maps the prediction and keypoint into this format and calls the current metric.
> 
> So if the model predicted "I like my canine friend" is matched to "I like dogs", we will convert it to:
> 
> [( "matched", "I like dogs" )]
> 
> And if not, we will convert it to:
> 
> []
> 
> Then we will get micro, macro f1 and precision recall over the key points - with one call.
> 
> Something like:
> 
> class KPA(NER): def compute( self, references: List[Any], predictions: List[Any], additional_inputs: List[Dict], ) -> dict:
> 
> ```
>      predictions = [ (prediction, additional_inputs['keypoint]) for prediction,additional_input in zip(predictions, additional_inputs) ]   
>     
>           
>      references  = [ [(references[0], additional_inputs['keypoint])] if [reference[0] == "matched") else [] for references,additional_inputs in zip(references, additional_inputs) ]   
>      return super.compute(references, predictions, additional_inputs)
> ```

I am not sure about this one.. i think its an advantage to reuse code, but on the other hand the task definition (and output, as part of its definition) should be as intuitive as possible.
IMO making the metric robust to various output types is a better option than defining a task output to match an expected metric input.

Also, a suggestion: maybe we can discuss these options online before moving on to the implementation..

**Comment by yoavkatz** [comment]
@elronbandel @lilacheden - I made a NON backward compatible change in the API, uniting all FilterByOperator to the single FilterByCondition.  Tis simplifies both the API and implementation.  For this, I added "in" and "not in" conditions to support lists.

**Comment by lilacheden** [comment]
> @elronbandel @lilacheden - I made a NON backward compatible change in the API, uniting all FilterByOperator to the single FilterByCondition. Tis simplifies both the API and implementation. For this, I added "in" and "not in" conditions to support lists.

Nice! 
For compatibility, maybe in the meanwhile we better keep FilterByValues/ListOfValues as deprecated and have them call the FilterByCondition?

**Comment by yoavkatz** [comment]
There are not many places where it was used. I find them in unitxt, there are a couple in fm-eval, which I'll change when we migrate to the new version.

**Comment by lilacheden** [comment]
> Thinking about it more. Maybe we should see KPA more of an extraction task like NER.
> 
> In NER we have the processed predictions (entity-surface-form, entity-type ) like:
> 
> [("Amir", "Person"), ("Yaron", "Person")] [("Ran", "Person"), ("Yonatan", "Person")]
> 
> and output:
> 
> "f1_Person": 0.57, "f1_macro": 0.57, "in_classes_support": 1.0, "f1_micro": 0.57, "recall_micro": 0.67, "recall_macro": 0.67, "precision_micro": 0.5, "precision_macro": 0.5, "score": 0.57, "score_name": "f1_micro", "f1_micro_ci_low": 0.57, "f1_micro_ci_high": 0.57, "score_ci_low": 0.57, "score_ci_high": 0.57,
> 
> (see unitxt/prepare/metrics/custom_f1.py)
> 
> I think all we need to do for KPA, is create a metric that maps the prediction and keypoint into this format and calls the current metric.
> 
> So if the model predicted "I like my canine friend" is matched to "I like dogs", we will convert it to:
> 
> [( "matched", "I like dogs" )]
> 
> And if not, we will convert it to:
> 
> []
> 
> Then we will get micro, macro f1 and precision recall over the key points - with one call.
> 
> Something like:
> 
> class KPA(NER): def compute( self, references: List[Any], predictions: List[Any], additional_inputs: List[Dict], ) -> dict:
> 
> ```
>      predictions = [ (prediction, additional_inputs['keypoint]) for prediction,additional_input in zip(predictions, additional_inputs) ]   
>     
>           
>      references  = [ [(references[0], additional_inputs['keypoint])] if [reference[0] == "matched") else [] for references,additional_inputs in zip(references, additional_inputs) ]   
>      return super.compute(references, predictions, additional_inputs)
> ```

Seems like an over-complication - we defined the task as a sentence-pair classification task, which is the most intuitive.

We will also need the binary precision and recall for the stance and kp quality tasks, which are purely classification tasks and fit well with the f1_multi_label implementation, so maybe it's better to reuse the code from there.

File: src/unitxt/metrics.py [hunk 3]
```diff
@@ -1430,3 +1447,27 @@ def _compute(
             for k in self.k_list:
                 result[self.score_name(measure_name, k)] = measure_array[min(k, max_k)]
         return result
+
+
+class KPA(NER):
```
matanor, line 1452: @yoavkatz suggestion: can we modify `F1MultiLabel` such that the grouping functions also have access to the additional_inputs passed to compute? and then we can directly group by an input field, instead of encoding that field in a tuple, to match the NER format?

matanor (changes requested):
hi @lilacheden @yoavkatz i made another pass, added some suggestions.

src/unitxt/metrics.py
```diff
@@ -965,30 +987,46 @@ def f1(self, pn, pd, rn, rd):
         except ZeroDivisionError:
             return self.zero_division
 
+    def get_classes(self, references, additional_inputs):
```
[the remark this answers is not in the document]
matanor: Since now these aren't necessarily classes ..

src/unitxt/metrics.py
```diff
@@ -965,30 +987,46 @@ def f1(self, pn, pd, rn, rd):
...
+        classes = set()
```
matanor, line 991:
```suggestion
        groups = set()
```

matanor: for the same reason ..

src/unitxt/metrics.py
```diff
@@ -965,30 +987,46 @@ def f1(self, pn, pd, rn, rd):
...
+        for sublist, additional_input in zip(references, additional_inputs):
+            for e in sublist:
+                if self.should_ignore_element(e, additional_input):
+                    continue
+                classes.add(self.get_element_group(e, additional_input))
+        return classes
+
     def compute(
         self,
         references: List[Any],
```
matanor, line 1001:
```suggestion
        references: List[List[Any]],
```

matanor: i think that's an error in the existing type hint (unrelated to the changes here). might be worth to fix

src/unitxt/metrics.py
```diff
@@ -965,30 +987,46 @@ def f1(self, pn, pd, rn, rd):
         except ZeroDivisionError:
             return self.zero_division
 
+    def get_classes(self, references, additional_inputs):
+        classes = set()
+        for sublist, additional_input in zip(references, additional_inputs):
+            for e in sublist:
```
matanor, line 993:
```suggestion
            for reference in single_example_references:
```

matanor: essentially, e is just a reference IIUC (here). so perhaps call it for what it is.

matanor (approved):
LGTM, left a few suggestion for renaming some items..

src/unitxt/loaders.py
```diff
@@ -140,7 +140,7 @@ class LoadFromIBMCloud(Loader):
     aws_secret_access_key_env: str
     bucket_name: str
     data_dir: str = None
-    data_files: Sequence[str]
+    data_files: Union[Sequence[str], Dict[str, str]]
```
matanor, line 143: Can you please add a comment explaining what is the use of the dict here?
Also, i suggest adding a test with a dict value would to handle the coverage messages below.

prepare/metrics/precision_recall.py
```diff
@@ -0,0 +1,315 @@
...
+)
+
+outputs = test_metric(
+    metric=precision_macro_metric,
+    predictions=predictions,
+    references=references,
+    instance_targets=instance_targets_precision_macro,
+    global_target=global_target_precision_macro,
+)
+# add_to_catalog(precision_metric, "metrics.precision", overwrite=True)
```
matanor, line 192: remove commented out line?

prepare/metrics/precision_recall.py
```diff
@@ -0,0 +1,315 @@
...
+    global_target=global_target_recall_micro,
+)
+outputs = test_metric(
+    metric=recall_macro_metric,
+    predictions=predictions,
+    references=references,
+    instance_targets=instance_targets_recall_macro,
+    global_target=global_target_recall_macro,
+)
+# add_to_catalog(recall_metric, "metrics.recall", overwrite=True)
```
matanor, line 121: remove commented out line?

prepare/metrics/precision_recall.py
```diff
@@ -0,0 +1,315 @@
...
+    global_target=global_target_precision_micro,
+)
+outputs = test_metric(
+    metric=precision_macro_metric,
+    predictions=predictions,
+    references=references,
+    instance_targets=instance_targets_precision_macro,
+    global_target=global_target_precision_macro,
+)
+# add_to_catalog(precision_metric, "metrics.precision", overwrite=True)
```
matanor, line 70: remove commented out line?

matanor (approved):
LGTM, left a few minor comments..

src/unitxt/loaders.py
lilacheden: @matanor - Done, and also added another optional input type, is the current version ok?

src/unitxt/loaders.py
```diff
@@ -140,7 +140,12 @@ class LoadFromIBMCloud(Loader):
     aws_secret_access_key_env: str
     bucket_name: str
     data_dir: str = None
-    data_files: Sequence[str]
+
+    # Can be either:
+    # 1. a list of file names, the split of each file is determined by the file name pattern
+    # 2. Mapping: split -> file_name, e.g. {"test" : "test.json", "train": "train.json"}
+    # 3. Mapping: split -> file_names, e.g. {"test" : ["test1.json", "test2.json"], "train": ["train.json"]}
```
matanor, line 147: very clear, tks for adding!

matanor (approved):
LGTM

PR #426, opened 2023-12-24
modify to different bleu impl.

gitMichal:
change the translation metric from bleu to normalized_sacrebleu


elronbandel (changes requested):
you need also to add the resulting json to the catalog

elronbandel (approved):

PR #431, opened 2023-12-26
adding ci to scores other than main_score

assaftibm:


src/unitxt/metrics.py
```diff
@@ -63,6 +63,7 @@ class MetricWithConfidenceInterval(Metric):
     # Use None to disable confidence interval computation.
     n_resamples: int = None
     confidence_level: float = 0.95
+    ci_scores: List[str] = None
```
yoavkatz, line 66: I like that there is now a way to define which scores to run on.  I suggest, that by default it will be set to an empty list (so by default, ci will not be calculated).  Then, we will add for each metric the relevant scores or the default score.

yoavkatz:
A couple of tests fail.

prepare/metrics/rag.py
```diff
@@ -39,7 +31,7 @@
     {"f1": 0, "precision": 0, "recall": 0, "score": 0, "score_name": "f1"},
 ]
 
-# Currently CopyFields does not delete the source fields,
+# Currently, CopyFields does not delete the source fields,
```
dafnapension, line 34: will RenameFields  fit better here?

dafnapension (approved):

src/unitxt/metrics.py
matanor: I opened an issue for that: https://github.com/IBM/unitxt/issues/432.

yoavkatz: @matanor @assaftibm .  Do we want to make the change in this PR or in separate one?

matanor: @assaftibm @yoavkatz i tend towards having a separate one

yoavkatz: OK. Let's have a separate one.  There are few failing tests that need to be addressed.

prepare/metrics/rag.py
assaftibm: I'm not familiar with this, @elronbandel - what do you think?

src/unitxt/metrics.py
assaftibm: @yoavkatz I guess I'm not seeing the full picture, but are there metrics for which CI won't be computed? (and if so, why is that?)

prepare/metrics/rag.py
dafnapension: If you need 'move' rather than 'copy', then 'RenameField' will do the job for you. It carefully removes the source, paying attention to dpath subtle situations.

src/unitxt/metrics.py
matanor: The reason is runtime. For example, it was disabled for the default version of rouge ([here](https://github.com/IBM/unitxt/blob/main/prepare/metrics/rouge.py#L5)). 

There are other cases of users that asked to turn it off. 

This issue is mainly (or completely) for global metrics.

PR #433, opened 2023-12-28
Added support for Japanese in sacrebleu

yoavkatz:
1. Added new requirement
2. Passed tokenize ja-mecab to metric
3. Renamed sacrebleu file to avoid name conflict with HF sacrebleu imports.

### DIFF
--- a/prepare/metrics/normalized_sacrebleu.py
+++ b/prepare/metrics/normalized_sacrebleu.py
@@ -4,11 +4,12 @@ from src.unitxt.operators import CopyFields, MapInstanceValues
 from src.unitxt.test_utils.metrics import test_metric
 
 language_to_tokenizer = {
-    "deutch": "",
-    "french": "",
-    "romanian": "",
-    "english": "",
-    "en": "",
+    "deutch": None,
+    "french": None,
+    "romanian": None,
+    "english": None,
+    "en": None,
+    "japanese": "ja-mecab",
     "ja": "ja-mecab",
 }
 
--- a/prepare/metrics/normalized_sacrebleu.py
+++ b/prepare/metrics/normalized_sacrebleu.py
@@ -1,14 +1,33 @@
 from src.unitxt import add_to_catalog
-from src.unitxt.metrics import HuggingfaceMetric
+from src.unitxt.metrics import HuggingfaceMetric, MetricPipeline
+from src.unitxt.operators import CopyFields, MapInstanceValues
 from src.unitxt.test_utils.metrics import test_metric
 
-metric = HuggingfaceMetric(
-    hf_metric_name="sacrebleu",
-    hf_main_score="score",
+metric = MetricPipeline(
     main_score="sacrebleu",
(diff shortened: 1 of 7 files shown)


gitMichal:
thanks, Yoav. Two questions, what should we use for the tasks that used to use sacrebleu?
if we will have more languages that requires tokenizer, how should we do this?

yoavkatz:
I only added this parameter to the existing metrics.  Based on the tests, it did not change the results I English.

We were told: "Here is an example of using Ja word tokenizer with HF bleu, I've also added example for rougeL, these examples works for any language, en, de, es, fr pt and ja"

gitMichal (approved):

yoavkatz:
@matanor - Please see this PR.  Can you explain, why I had to change the confidence interval expected results , when I made a change that does not effect the functionality.   The old and  new expected results for ci are deterministic, it's just that changing the wrapper code, seems to cause them to generate different results.

matanor:
> @matanor - Please see this PR. Can you explain, why I had to change the confidence interval expected results , when I made a change that does not effect the functionality. The old and new expected results for ci are deterministic, it's just that changing the wrapper code, seems to cause them to generate different results.

Not sure, they should have stayed the same, i think. Will try to debug this.

matanor:
> > @matanor - Please see this PR. Can you explain, why I had to change the confidence interval expected results , when I made a change that does not effect the functionality. The old and new expected results for ci are deterministic, it's just that changing the wrapper code, seems to cause them to generate different results.
> 
> Not sure, they should have stayed the same, i think. Will try to debug this.

@yoavkatz 
The change is due to the metric becoming a `MetricPipeline` instead of a `GlobalMetric`.

This is the root cause: https://github.com/IBM/unitxt/blob/main/src/unitxt/test_utils/metrics.py#L75

I suppose we should change that `test_metric` for `MetricPipeline` objects also uses n_resamples=3 when testing the confidence intervals. What do you think?

yoavkatz:
> > > @matanor - Please see this PR. Can you explain, why I had to change the confidence interval expected results , when I made a change that does not effect the functionality. The old and new expected results for ci are deterministic, it's just that changing the wrapper code, seems to cause them to generate different results.
> > 
> > 
> > Not sure, they should have stayed the same, i think. Will try to debug this.
> 
> @yoavkatz The change is due to the metric becoming a `MetricPipeline` instead of a `GlobalMetric`.
> 
> This is the root cause: https://github.com/IBM/unitxt/blob/main/src/unitxt/test_utils/metrics.py#L75
> 
> I suppose we should change that `test_metric` for `MetricPipeline` objects also uses n_resamples=3 when testing the confidence intervals. What do you think?
 
 
Yes. That makes sense.

yoavkatz:
Maybe 3 is too low of a sample size? I see it causes some metrics to have Nan in the confidence interval.

yoavkatz:
I'll create a new issue for resamples change.

PR #434, opened 2023-12-28
Create tasks_taxonomy

gitMichal:
wip.. 
taxonomy definition

### DIFF
--- a/docs/docs/tasks_taxonomy
+++ b/docs/docs/tasks_taxonomy
@@ -91,3 +91,6 @@ Semantic Textual Similarity
 
 Information Retrieval
 -------
+
+Code
+-------
--- a/docs/docs/tasks_taxonomy
+++ b/docs/docs/tasks_taxonomy
@@ -1,0 +1,56 @@
+=================
+Tasks Taxonomy
+=================
+
+Tasks taxonomy allows easy navigation throughout the different tasks. The taxonomy is inspired by `paperswithcode <https://paperswithcode.com>`_.
+
+
+Text classification
+-------
+* Multi-class classification
+
+  * Intent classification
+
+  * Sentiment analysis
+
+* Mutli-label classification
+
+  * Emotion classification
+
+
+Question Answering
+-------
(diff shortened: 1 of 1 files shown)


borgr (commented):
We differentiate open QA from multiple choice but not abstractive and extractive and not regular from multihop? (This makes sense to me, as the templates might be the same. So the distinction just helps finding the task but have little practical value)
Also closed book must have their own sub category. 
What about no answer? It is probably shown in the templates, but should it define the category somehow?
Same for summarization abstractive and extractive?
Summarization as you know is just very ill-defined but a lot could be changed there as well (headline generation simplification etc.).
what unifies the different text generations? What is MT and summarization not under it?

borgr (commented):
Also, what about regression? (ranking and clustering I guess are just very rare tasks in NLP outputs)

gitMichal:
> We differentiate open QA from multiple choice but not abstractive and extractive and not regular from multihop? (This makes sense to me, as the templates might be the same. So the distinction just helps finding the task but have little practical value) Also closed book must have their own sub category. What about no answer? It is probably shown in the templates, but should it define the category somehow? Same for summarization abstractive and extractive? Summarization as you know is just very ill-defined but a lot could be changed there as well (headline generation simplification etc.). what unifies the different text generations? What is MT and summarization not under it?

thanks @borgr , will try to address some of the issues you raised..
- indeed "no answer" is somehow encoded in some of the templates. If you think it deserves a category by itself, we should consider.
- for summarization i think that extractive is (almost) no longer in the game... and thus wonder if we should create such a sub-category
- MT- machine translation 
- about text generation.. i agree that another taxonomy could be NLG, NLU and then to put under NLG summarization.. but again we need to stick to something..

gitMichal:
i did few more changes. @borgr please take a looks and feel free to modify

yoavkatz:
Can we close this?

yoavkatz:
Closing as I assume it's no longer needed.

PR #435, opened 2023-12-31
additional verification toward uniquely determined result

dafnapension:
no two different from_fields feed the same to_field. Otherwise, can not determine which will be the last one, which finalizes the value of that to_field.


yoavkatz:
@dafnapension @elronbandel - Why do we need this complexity?  It's very hard to follow.   Why not just allow a single pair of from/to fields , and just require multiple operators when needed.  This will make the order unambiguous, the API simpler and the use less error prone.  The downside, is that it will be a bit more verbose.  

What do you think?

dafnapension:
Verification is not too complex, but I suggest a Mapai-c Compromise: 
A Disclaimer (to replace the 'why not' in the docstring):
Please note that in the case of using the `field_to_field` input, the order by which its pairs are processed is not deterministic. Therefore, to avoid any ambiguity, in case of more than one occurrence of any field in any position in `field_to_field`, recommended is a partition of `field_to_field` into two or more successive invocations of the operator.

yoavkatz:
I think this behavior complicates the users, with very hard to understand error message.  Unitxt requires learning a new language - we should not make it more complicate than needed.

Let's wait for Elron and decide whether we want to keep this functionality.

dafnapension:
Yes, I, too, suggest to hear Elron's word before merging this PR in (let alone the [huge other PR](https://github.com/IBM/unitxt/pull/423) that I developed per his guidance, as I understood it..) .
Just notice that I only think now to say some semi-warning words in the docstring. Enforcing through an involved functionality is not a must. I agree.

elronbandel:
In the one hand it could be appealing to have every operator do only one operation on one field. It has two issues: (1) some operators we have such as the formatter/template are operating on many fields. (2) Having only single operation operators can result in many redundant operation. 

My suggestion is to simplify by having all the operators to operate on many fields by using field_to_field dictionary. This will align with other operators and will offer simple api. We can eliminate many order issues by verification that a field_to_field dictionary does not contain dependencies. (it happens only once when creating the operator, so could be heavy check).

This will also remove the ambiguity of operators with only `field` parameter that implicitly define the `to_field` parameter. 

Yet im not 100% sure about all of those api simplicity decisions since I mainly see things from the perspective of heavy unitxt user.

yoavkatz:
I don't think that all operators can or should work on a single field.  But most operators can work on a single field, or a pair of fields.

First, all field operators can support only `field` and  `to_field` (by default , to_field=field).  Also MapInstanceValues can become a field operator.

I can say, that many users report it  hard to learn memorize a new language for things that are easily done in python.  We need to make it simple as possible for them.

dafnapension:
The [script where I "caught" the loop](https://github.com/IBM/unitxt/blob/main/prepare/metrics/squad.py) used a series of copies in one invocation of CopyField, and I "forced" them (they did not notice, it was part of my PR) to break the long list into two invocation. 
I now think that the best is:
To announce in docstring that when `field_to_field` is a `[List[List[str]]` (inner lists are of length 2, this is verified), the order of applications of the field operation (the invocation of method `process_value`)) is the order of the pairs in `field_to_field`.  the `field` of the pair (the first component) is fed to `process_value`, and the result is stored in `to_field` of the pair (the second component of the pair).
But when `field_to_field` is a `Dict[str, str]]`, with two or more `items()`, then no order of operation is guaranteed, and hence if the result depends on the order of operation, an AssertionError is raised.
Such a dependency occurs if either same field name plays the role of `to_field` in one pair, and the role of `field` in a different pair, or when two different `field`-s are paired with same `to_field`.

I will commit that change, please see if it looks acceptable.
If OK, then decency demands that I return the long copy list to where I caught it...

elronbandel (approved):

elronbandel:
Until we decide to change the api and remove field_to_field these changes are necessary. Further discussion about the removal of field_to_field should be somewhere else imho.

yoavkatz:
We can the longer term simplification options in a separate issue/PR.

dafnapension:
Thank you, @elronbandel  and @yoavkatz !

PR #442, opened 2024-01-01
Perplexity

assaftibm:


assaftibm:
The metric computes the likelihood of generating the text Y after the text X (i.e. P(Y|X))
There is no clear alignment between `X` and `Y`, and the notions of `reference` and `prediction` as used in Unitxt.
After consulting with Yosi Mass, we think it makes sense to treat them this way:
*  `X` = `prediction`
*  `Y` = `reference`.

yoavkatz (changes requested):
I think you also need to commit the json files.

assaftibm:
> I think you also need to commit the json files.

I just pushed them now

yoavkatz:
Perhaps use this naming convention:

instead of
src/unitxt/catalog/metrics/perplexity/flan/t5/small.json

use
src/unitxt/catalog/metrics/perplexity/flan_t5_small.json


So there will be less directories.

assaftibm:
fixed

yoavkatz (approved):

Issue #445, opened 2024-01-03
Document how to edit the Unitxt docs


matanor:
Add a section to the unitxt contributors guide which explains how to work with the unitxt docs. Where are the docs files, how to start a local docs server that visualized the documentation (`make docs-server`).

PR #446, opened 2024-01-03
Added instructions on how to update the unitxt docs

matanor:
Add a section to the unitxt contributors guide which explains how to edit and test the unitxt docs.

Closes https://github.com/IBM/unitxt/issues/445

### DIFF
--- a/docs/docs/contributors_guide.rst
+++ b/docs/docs/contributors_guide.rst
@@ -4,12 +4,20 @@ Contributors Guide
 
 This guide will assist you in contributing to unitxt.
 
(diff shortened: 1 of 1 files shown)


elronbandel (changes requested):
I think we should add a section describing that half the documentation is from the .rst files and half is generated automatically from the docstrings within the library python files

matanor:
> I think we should add a section describing that half the documentation is from the .rst files and half is generated automatically from the docstrings within the library python files

hi @elronbandel i added a description, please have a look

elronbandel (approved):

Issue #415, opened 2023-12-19
An operator that supports applying lambda functions


matanor:
The [apply](https://github.com/IBM/unitxt/blob/main/src/unitxt/operators.py#L668) operator supports applying functions to specific fields. However it only support regular functions and not lambda functions.

Adding support for lambda functions would allow a more flexible application of custom code.

PR #449, opened 2024-01-05
Add FilterByQuery and ExecuteQuery as simple operators receiving textual python query

dafnapension:
It seems that the simple conditions, listed in `FilterByCondition` , can all be replaced by simple lambda conditions.  We can thus replace, and enjoy from additional, perhaps more involved, conditions that lambda can express. 
This also refers to [an issue](https://github.com/IBM/unitxt/issues/415) by @matanor

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1263,30 +1263,25 @@ class FilterByCondition(SingleStreamOperator):
         return True
 
 
-class FilterByLambdaCondition(SingleStreamOperator):
-    """Filters a stream, yielding only instances for which the input specified fields fulfil the conditions assigned to them.
+class FilterByQuery(SingleStreamOperator):
+    """Filters a stream, yielding only instances which fulfil all the conditions specified as strings to be python's eval-uated.
 
     Raises an error if a field for which a condition is specified is missing from the instance
 
     Args:
-       lambda_conditions (Dict[str, str]): a mapping of fields to lambda conditions encoded as strings
+       queries (List[str]): a list of conditions encoded as strings, that are to be processed by python's eval()
        error_on_filtered_all (bool, optional): If True, raises an error if all instances are filtered out. Defaults to True.
 
     Examples:
-       FilterByCondition(lambda_conditions = {"a": "lambda x: x > 4"}) will yield only instances where "a">4
-       FilterByCondition(lambda_conditions = {"a": "lambda x: x <= 4"}) will yield only instances where "a"<=4
-       FilterByCondition(lambda_conditions = {"a": "lambda x: x in [4, 8]"}) will yield only instances where "a" is 4 or 8
-       FilterByCondition(lambda_conditions = {"a": "lambda x: x not in [4, 8]"}) will yield only instances where "a" is neither 4 nor 8
+       FilterByCondition(lambda_conditions = {"a > 4"}) will yield only instances where "a">4
+       FilterByCondition(lambda_conditions = {"a <= 4"}) will yield only instances where "a"<=4
+       FilterByCondition(lambda_conditions = {"a in [4, 8]"}) will yield only instances where "a" is 4 or 8
+       FilterByCondition(lambda_conditions = {"a not in [4, 8]"}) will yield only instances where "a" is neither 4 nor 8
 
     """
 
-    lambda_conditions: Dict[str, str]
+    queries: List[str]
     error_on_filtered_all: bool = True
-
-    def prepare(self):
-        self.functs = {}
-        for key, val in self.lambda_conditions.items():
-            self.functs[key] = eval(val)
 
     def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generator:
         yielded = False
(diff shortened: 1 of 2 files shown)


src/unitxt/operators.py
```diff
@@ -1475,7 +1532,7 @@ class AddFieldNamePrefix(StreamInstanceOperator):
     """Adds a prefix to each field name in each instance of a stream.
 
     Args:
-        prefix_dict (Dict[str, str]): A dictionary mapping stream names to prefixes.
```
dafnapension, line 1478: Operator `AddFieldNamePrefix` is used nowhere. Shall I remove it?

elronbandel: yes looks unnecessary

dafnapension: Removed.. as part of this branch. If this branch will not be merged into main, I will remember to remove `AddFieldNamePrefix` in another PR..

elronbandel (changes requested):
The direction is really promising. 
I think a better machinism will be to use eval directly:
```
>>> eval("x < y", None, {"x":5, "y":7})
True
>>> eval("x < y", None, {"x":5, "y":3})
False
```
That way we can have complicated conditions consisted of many variables. 

so the final api can be:
`operator = FilterByQuery(query="len(target) > 5 or target == 'yes'")`

Accordingly we can add also general operator:
`operator = ExecuteQuery(query="target + '_end'",  to_field="target")`

dafnapension:
> The direction is really promising. I think a better machinism will be to use eval directly:
> 
> ```
> >>> eval("x < y", None, {"x":5, "y":7})
> True
> >>> eval("x < y", None, {"x":5, "y":3})
> False
> ```
> 
> That way we can have complicated conditions consisted of many variables.
> 
> so the final api can be: `operator = FilterByQuery(query="len(target) > 5 or target == 'yes'")`
> 
> Accordingly we can add also a general operator: `operator = ExecuteQuery(query="target + '_end'", to_field="target")`

A neat idea!!  replaced FilterByLambdaCondition to FilterByQuery, and implemented.  Indeed neat!

dafnapension:
@elronbandel , this is really neat! I now added `ExecuteQuery` per your idea! I think it may be handy for metrics (I am now digging into issue 439)

yoavkatz:
I also encountered this problem that may pose a new requirement.

We have for each instance two fields with llists
   labels = ['person', 'location', 'person']
   entity = ['Jane,     'New York',  'John']
   
   and I want to filter both lists together to and keep only 'persons'.
   
   labels = ['person',  'person']
   entity = ['Jane',  'John']
   
   How could this be handled?

dafnapension:
Hi, @yoavkatz ,  per [your question](https://github.com/IBM/unitxt/pull/449#issuecomment-1881205717), I added a unittest that does just that.
```
operator = ExecuteQuery(
            queries_to_fields=[
                ['[e for l,e in zip(labels, entity) if l=="person"]', "entity"],
                ['[l for l in labels if l == "person"]', "labels"],
            ]
        )
```

matanor:
> The direction is really promising. I think a better machinism will be to use eval directly:
> 
> ```
> >>> eval("x < y", None, {"x":5, "y":7})
> True
> >>> eval("x < y", None, {"x":5, "y":3})
> False
> ```
> 
> That way we can have complicated conditions consisted of many variables.
> 
> so the final api can be: `operator = FilterByQuery(query="len(target) > 5 or target == 'yes'")`
> 
> Accordingly we can add also a general operator: `operator = ExecuteQuery(query="target + '_end'", to_field="target")`

Suggestion: what i was originally thinking of is this:

`operator = Filter(lambda x: len(x[target]) > 5 or x[target] == 'yes')`

Or

`Execute(lambda x: x[target] + '_end', to_field="target")`

I think this avoids the need to express python code in strings. This of for example having a syntax error in the code string.. would be easier IMO if its just a lambda expression.

yoavkatz:
A few points:
1. I agree that adding a lambda expression is better than string. Although at the end, we will store in the json the string representation, and need to maintain this option.
2. Why support multiple queries in same operator call?  I think it will be clearer  to have it one by one, as in Matan's example: Execute(lambda x: x["target"] + '_end', to_field="target")

dafnapension:
I thought that in the spirit of `FieldOperator` , we may want to complete a mission in one call, and a series of executions can do it easier than squashing somehow to one execution. Like with your example of labels and entities.

yoavkatz:
> I thought that in the spirit of `FieldOperator` , we may want to complete a mission in one call, and a series of executions can do it easier than squashing somehow to one execution. Like with your example of labels and entities.

You know by take, that this is complicated API.  I think we should discuss with Elron.

src/unitxt/operators.py
```diff
@@ -1263,6 +1263,87 @@ def _is_required(self, instance: dict) -> bool:
...
+
+    Examples:
+       FilterByCondition(lambda_conditions = {"a > 4"}) will yield only instances where "a">4
+       FilterByCondition(lambda_conditions = {"a <= 4"}) will yield only instances where "a"<=4
+       FilterByCondition(lambda_conditions = {"a in [4, 8]"}) will yield only instances where "a" is 4 or 8
+       FilterByCondition(lambda_conditions = {"a not in [4, 8]"}) will yield only instances where "a" is neither 4 nor 8
+
+    """
+
+    queries: List[str]
```
elronbandel, line 1283:
```suggestion
    query: str # "x > 0 and y < 3"
```

src/unitxt/operators.py
```diff
@@ -1263,6 +1263,87 @@ def _is_required(self, instance: dict) -> bool:
...
+       ExecuteQuery(queries_to_fields =  [["a+b", "c"],["c+10", "d"]])
+       the result is {"a": 2, "b": 3, "c": 5, "d": 15}
+
+       When instance {"a": "hello", "b": "world"} is process-ed by operator
+       ExecuteQuery(queries_to_fields = [["a+' '+b", "c"]])
+       the result is {"a": "hello", "b": "world", "c": "hello world"}
+
+    """
+
+    queries_to_fields: List[List[str]]
```
elronbandel, line 1327:
```suggestion
    query: str
    to_field: str
```

dafnapension:
> > I thought that in the spirit of `FieldOperator` , we may want to complete a mission in one call, and a series of executions can do it easier than squashing somehow to one execution. Like with your example of labels and entities.
> 
> You know by take, that this is complicated API. I think we should discuss with Elron.

@elronbandel agrees with you, @yoavkatz , so I replaced the list of queries by a single query.

elronbandel (approved):
Amazing work!

dafnapension:
Thank you, @elronbandel and @yoavkatz  for suggesting the operators, and for your reviews and improvements

yoavkatz:
Before we release it to uses I suggest a few additional changes (for a new PR)

Rename "FilterByQuery" to "FilterByExpression".
Rename "ExecuteQuery" to "ExecuteExpression".

Issue #389, opened 2023-12-12
In GlobalMetric confidence interval, need check to make sure new generated value is not NaN


sam-data-guy-iam:
@matanor @elronbandel 
In GlobalMetric's compute_global_confidence_intervals, the following lines of code generate new draws to replace any with NaN values.  But there is no check that the resulting samples do not also result in NaNs (as happened to me).  One can debate whether you would want to leave it as is if there are NaNs (i.e. the user may want to know how likely NaNs are, if these are valid values).   I think a better solution is to set some maximum number of tries, say 50, and try one by one to replace the NaNs with a new draw (up to the number of tries), replacing it if not NaN, otherwise continuing, until either the number of tries is reached or there are no NaNs left.  And then finally removing any NaNs (by recalculating error_indices on the final scores vector).   The code as it exists simply tries again without any check on the replacement validity.
On a related note, does a similar type of check need to be done on the score_based_confidence intervals if not global?
````
if n_errors > 0:
           new_scores = random_gen.choice(scores, n_errors, replace=True)
           scores = scores[~error_indices]
           scores = np.concatenate([scores, new_scores])
````

sam-data-guy-iam:
@matanor I think what needs to be done instead is simply when resampling new_scores to sample from scores[~error_indices] so that none will be NaN.  I apologize I was thinking it was generating new resamplings from the original references and predictions, like scores[error_indices] = random_gen.choice(scores[~error_indices], n_errors, replace=True)

matanor:
@sam-data-guy-iam IIUC your suggestion is to draw non-NaN values, since there is no point in resampling values that may be NaN, if we are to remove NaN values in the first place.. This sounds like a good idea to me ...
Do you have the capacity to do this fix?

sam-data-guy-iam:
I made the change for global metrics already in my PR for the InstanceMetric.  I used the same method as in InstanceMetric to simply resample only from the non NaN scores, so we can close this issue once the PR is integrated.

matanor:
nice! tks for fixing this..

PR #452, opened 2024-01-08
Grouped instance metric inherit from InstanceMetrics

sam-data-guy-iam:
Please ignore the earlier PR for aggregation metrics and use this branch instead.  I took @matanor s suggestion to modify InstanceMetric and make the grouped mean type metrics inherit from the InstanceMetric classes that define the instance calculation.

Also: Closes  https://github.com/IBM/unitxt/issues/389


src/unitxt/metrics.py
```diff
@@ -399,31 +541,39 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
 
-        for instance in instances:
-            yield instance
+        group_total_scores = [
+            aggregation_func(scores) for scores in group_to_instance_scores.values()
+        ]
+        group_total_scores = [
+            score for score in group_total_scores if not np.isnan(score)
+        ]
+        # ignore NaNs in aggregation
```
matanor, line 575: The comment is not what the line below does.. should it be removed?

src/unitxt/metrics.py
```diff
@@ -82,43 +83,98 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
+                    arr=arr,
+                )
+
+                return self.resample_from_non_nan(scores)
+
+            # apply bootstrap only on the relevant field
             ci = bootstrap(
-                (scores,),
-                statistic=mean,
+                (identifiers,),
```
matanor, line 149: can you please explain why pass the `identifiers` and not the `instances` here?

src/unitxt/metrics.py
```diff
@@ -82,43 +83,98 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
-            ]
+            # need to redefine the statistic function within the loop because score_name is a loop variable, to avoid ruff errors
+            def statistic(arr, axis, score_name=score_name):
+                # arr is a 2d array where each row is a resampling, so we
+                # iterate over the rows and compute the metric on each resampling
+                def metric(instances):
+                    try:
+                        return aggregation_func(instances, score_name)
+                    except Exception as e:
+                        # this happens in edge cases, for example, when the sampling creates a
```
matanor, line 132: is this something that happens? because the comment talks about bleu which is a GlobalMetric.

matanor (changes requested):
hi @sam-data-guy-iam thanks for this new PR, looks good!
i left a few comments, please have a look..

src/unitxt/metrics.py
```diff
@@ -368,14 +422,102 @@ def compute(
...
+            assert (
+                "agg_func" in fields
+            ), "fields should have a key 'agg_func' consisting of a 2-element list of a function name and function definition"
+            assert callable(
+                fields["agg_func"][1]
+            ), "second item in fields['agg_func'] should be a callable function"
+            if "score_fields" in fields:
+                assert isinstance(fields["score_fields"], list)
+
+            return (
```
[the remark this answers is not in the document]
sam-data-guy-iam: good point, I will move the return outside the verification function

src/unitxt/metrics.py
```diff
@@ -368,14 +422,102 @@ def compute(
...
+                [self.main_score]
+                if "score_fields" not in fields
+                else fields["score_fields"]
+            )
+        return [self.main_score]
+
     def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generator:
+        instances, global_score = self.compute_instance_scores(stream, stream_name)
```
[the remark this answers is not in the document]
sam-data-guy-iam: yes this was done because ruff objected "process" was too complex but it's a good idea anyhow, I think

src/unitxt/metrics.py
sam-data-guy-iam: you're right, I copied this from GlobalMetric confidence intervals.  In this case, since the instance scores are already computed, there is no additional computation on the instances (unless there is a corner case where one designs an aggregation function to do something weird like this) there should be no issue, and the confidence interval already deals with the case where there are NaNs. I am removing it.

src/unitxt/metrics.py
sam-data-guy-iam: You're right, I found a work-around.  I had copied this part of the code from the GlobalMetric (which might not need it either).

src/unitxt/metrics.py
sam-data-guy-iam: added a warnings catch for the RuntimeWarning from nanmean

src/unitxt/version.py
```diff
@@ -1 +1 @@
-version = "1.4.3"
+version = "1.4.3"
```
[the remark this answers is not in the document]
matanor: Its still here, no? just with an updated version:
image

src/unitxt/metrics.py
```diff
@@ -82,43 +83,90 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
-                statistic=mean,
+                (instances,),
+                statistic=statistic,
                 n_resamples=self.n_resamples,
                 confidence_level=self.confidence_level,
                 random_state=self.new_random_generator(),
             ).confidence_interval
-            result[f"{score_name}_ci_low"] = ci.low
-            result[f"{score_name}_ci_high"] = ci.high
+            full_score_name = (
```
matanor, line 147: isn't this just `field_name_full` from https://github.com/IBM/unitxt/pull/452/files#diff-5623e0c8548470ce6339f802312be342978bc20e265c170fab1722acbbd5154eR487 or from https://github.com/IBM/unitxt/pull/452/files#diff-5623e0c8548470ce6339f802312be342978bc20e265c170fab1722acbbd5154eR484 ?

src/unitxt/metrics.py
```diff
@@ -82,43 +83,90 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
 
         for score_name in score_names:
-            scores = [
-                instance["score"]["instance"][score_name] for instance in instances
-            ]
+            # need to redefine the statistic function within the loop because score_name is a loop variable, to avoid ruff errors
+            def statistic(arr, axis, score_name=score_name):
+                # arr is a 2d array where each row is a resampling, so we
+                # iterate over the rows and compute the metric on each resampling
+                # def metric(instances):
```
matanor, line 126: remove the comment of lines 126-7 ?

src/unitxt/metrics.py
```diff
@@ -82,43 +83,90 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
-
         result = {}
 
         if not self._can_compute_confidence_intervals(num_predictions=len(instances)):
             return result
 
-        score_names = (
-            self.ci_scores if self.ci_scores is not None else [self.main_score]
-        )
+        if score_names is None:
```
matanor, line 112: suggest to always explicitly pass this param, i.e. remove the default, and update the relevant calls

src/unitxt/metrics.py
```diff
@@ -82,43 +83,90 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
+            score_names = (
+                self.ci_scores if self.ci_scores is not None else [self.main_score]
+            )
+        if aggregation_func is None:
```
matanor, line 116: same here, suggest to always explicitly pass this param, i.e. remove the default, and update the relevant calls (there is just one call IIRC)

src/unitxt/metrics.py
```diff
@@ -82,43 +83,90 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
+
+            def aggregation_func(instances, field_name):
+                return mean([instance[field_name] for instance in instances])
 
         for score_name in score_names:
```
matanor, line 121: not sure whether this loop should still be inside `score_based_confidence_interval`, because for `InstanceMetric` objects we pass only a single score. It might be easier to update  `BulkInstanceMetric` to run such a loop and call score_based_confidence_interval with one score each time. What do you think?

src/unitxt/metrics.py
```diff
@@ -399,31 +531,48 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+        group_total_scores = [
+            aggregation_func(scores) for scores in group_to_instance_scores.values()
+        ]
+        import warnings
+
+        with warnings.catch_warnings():
+            # final mean should be mean of group_total_score, ignoring NaN, hence nanmean
+            # but if the group function values is NaN for ALL groups, nanmean throws a
+            # RuntimeWarning that it is calculating the mean of an empty slice (with no non-Nans)
+            # this is the desired behavior, but we want to avoid the warning here
```
matanor, line 573: Very clear comment, useful!

src/unitxt/metrics.py
```diff
@@ -399,31 +531,48 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
-                        for instance in instances
-                    ]
-                    global_score[field_name] = mean(scores)
-                    if field_name == self.main_score:
-                        global_score["score"] = global_score[field_name]
-                        global_score["score_name"] = self.main_score
+        with warnings.catch_warnings():
+            # in case instances is empty, return NaN but avoid printing a RuntimeWarning
+            warnings.simplefilter("ignore", category=RuntimeWarning)
+            return np.nanmean(scores)
```
matanor, line 544: what happened before your changes if some of the instance scores were NaN? Is it still the same behaviour?

matanor (changes requested):
hi @sam-data-guy-iam tks for making the changes, i went over the changes in metrics.py again, pls have a look ..

src/unitxt/version.py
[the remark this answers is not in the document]
sam-data-guy-iam: yes, I updated the branch with the merge, but I'm not sure what to do to fix this in particular.  I guess this file can just be removed from the PR?

src/unitxt/metrics.py
```diff
@@ -399,31 +531,48 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
-        for reduction, fields in self.reduction_map.items():
-            assert (
-                reduction in self.implemented_reductions
-            ), f"Reduction {reduction} is not implemented, use one of {self.implemented_reductions}"
+        return instances, global_score
 
-            if reduction == "mean":
-                from statistics import mean
+    @staticmethod
+    def aggregate(instances, field_name):
```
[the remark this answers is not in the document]
sam-data-guy-iam: renaming this and the group function

src/unitxt/metrics.py
```diff
@@ -368,14 +414,100 @@ def compute(
...
+    # Grouped metrics aggregate the instance score per group, and then average over group scores.
+    grouping_field: str = None
 
     @property
     @abstractmethod
     def reduction_map(self) -> dict:
         pass
 
+    def _validate_group_mean_reduction(self):
+        if "group_mean" in self.reduction_map:
```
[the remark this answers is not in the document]
sam-data-guy-iam: no problem, adding it as an assert without an if condition, just in case

src/unitxt/metrics.py
```diff
@@ -368,14 +414,100 @@ def compute(
...
+            assert callable(
+                fields["agg_func"][1]
+            ), "second item in fields['agg_func'] should be a callable function"
+            if "score_fields" in fields:
+                assert isinstance(fields["score_fields"], list)
+
     def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generator:
+        instances, global_score = self.compute_instance_scores(stream, stream_name)
+
+        for reduction, fields in self.reduction_map.items():
```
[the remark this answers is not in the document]
sam-data-guy-iam: accepted

src/unitxt/metrics.py
```diff
@@ -368,14 +414,100 @@ def compute(
...
+                reduction in self.implemented_reductions
+            ), f"Reduction {reduction} is not implemented, use one of {self.implemented_reductions}"
+
+            aggregation_func = None
+            if reduction == "mean":
+                aggregation_func = self.aggregate
+
+            if reduction == "group_mean":
+                self._validate_group_mean_reduction()
+                score_fields = (
```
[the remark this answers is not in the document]
sam-data-guy-iam: accepted

src/unitxt/metrics.py
```diff
@@ -368,14 +414,100 @@ def compute(
...
+        instances, global_score = self.compute_instance_scores(stream, stream_name)
+
+        for reduction, fields in self.reduction_map.items():
+            assert (
+                reduction in self.implemented_reductions
+            ), f"Reduction {reduction} is not implemented, use one of {self.implemented_reductions}"
+
+            aggregation_func = None
+            if reduction == "mean":
+                aggregation_func = self.aggregate
```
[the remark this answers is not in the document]
sam-data-guy-iam: accepted

src/unitxt/metrics.py
```diff
@@ -368,14 +414,100 @@ def compute(
...
+                ):
+                    return self.grouped_aggregate(instances, field_name, field)
+
+            if not aggregation_func:
+                raise ValueError(
+                    f"No aggregation_func was defined for reduction {reduction}. "
+                    f"Please specify a valid reduction method in reduction_map {self.reduction_map}."
+                )
+
+            for field_name in score_fields if reduction == "group_mean" else fields:
```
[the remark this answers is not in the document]
sam-data-guy-iam: accepted

src/unitxt/metrics.py
sam-data-guy-iam: yes but because this is called on the same instance set for both the group_mean and mean reductions, I want to differentiate between the CI for these two reductions, though the score_name operated on is the same for both reductions.  Therefore I pass the aggregation prefix for the group_mean reduction and an empty string if is just the CI of the mean of that score

src/unitxt/metrics.py
sam-data-guy-iam: yes sorry that was left over from when I was trying to remove the need to define metric.

src/unitxt/metrics.py
sam-data-guy-iam: Not necessarily because there is a distinction made (at least in principle) between the ci_scores (the fields for which a CI is calculated, which may be multiple fields such as in the F1-type metrics) and the main_score etc. for which the reduction is done.  That is, possibly one would want to do a reduction but not calculate the CI for a particular field.

matanor (commented):
hi @sam-data-guy-iam , i made another pass.. looks good! still have some not so big comment.. please have a look..

src/unitxt/metrics.py
```diff
@@ -368,14 +429,104 @@ def compute(
...
+                    aggregation_func=aggregation_func,
+                    score_names=list(set(self.ci_scores)),
+                    func_name=field_name_full_prefix,
+                )
+                global_score.update(confidence_interval)
+
+        yield from instances
+
+    def compute_instance_scores(
+        self, stream: Stream, stream_name: Optional[str] = None
```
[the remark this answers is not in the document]
sam-data-guy-iam: yes, agree, I will leave it in the process method

src/unitxt/metrics.py
```diff
@@ -82,43 +83,100 @@ def _can_compute_confidence_intervals(self, num_predictions):
             and num_predictions > 1
         )
 
-    def score_based_confidence_interval(self, instances):
-        """Compute confidence intervals based on existing scores, already computed on the input instances.
+    @staticmethod
+    def aggregate_instance_scores(instances, field_name):
```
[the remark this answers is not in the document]
sam-data-guy-iam: good idea

src/unitxt/metrics.py
```diff
@@ -368,14 +429,104 @@ def compute(
...
+        ), "reduction_map must have a `group_mean' key"
+        fields = self.reduction_map["group_mean"]
+        # for group_mean, expects a dict
+        assert isinstance(fields, dict)
+        assert (
+            "agg_func" in fields
+        ), "fields should have a key 'agg_func' consisting of a 2-element list of a function name and function definition"
+        assert callable(
+            fields["agg_func"][1]
+        ), "second item in fields['agg_func'] should be a callable function"
```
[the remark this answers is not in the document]
sam-data-guy-iam: agree, added a few other checks as well

src/unitxt/metrics.py
```diff
@@ -82,43 +83,100 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
+            field_name: score field names to compute mean for.
         """
-        from statistics import mean
+        scores = [instance["score"]["instance"][field_name] for instance in instances]
+        import warnings
+
+        with warnings.catch_warnings():
+            # in case instances is empty, return NaN but avoid printing a RuntimeWarning
+            warnings.simplefilter("ignore", category=RuntimeWarning)
+            return np.nanmean(scores)
```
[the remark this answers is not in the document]
sam-data-guy-iam: okay, I will return it to what it was before.  Previously, there was no Nanmean here, it was just mean.  If there are any NaN instance scores this will return NaN.  But I don't know any case where there are NaN instance scores.

src/unitxt/metrics.py
```diff
@@ -368,14 +429,104 @@ def compute(
...
+        default_factory=lambda: ["mean", "group_mean"]
+    )
+
+    # for grouped metrics: a field that contains the group id. None to disable grouping.
+    # Grouped metrics aggregate the instance score per group, and then average over group scores.
+    grouping_field: str = None
 
     @property
     @abstractmethod
     def reduction_map(self) -> dict:
```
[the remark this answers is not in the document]
sam-data-guy-iam: good point, I documented this in the _validate_group_mean_reduction method

src/unitxt/metrics.py
```diff
@@ -82,43 +83,100 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
+                score_name if len(func_name) == 0 else func_name + "_" + score_name
+            )
+            result[f"{full_score_name}_ci_low"] = ci.low
+            result[f"{full_score_name}_ci_high"] = ci.high
             if score_name == self.main_score:
                 result["score_ci_low"] = ci.low
                 result["score_ci_high"] = ci.high
         return result
 
+    def resample_from_non_nan(self, values):
```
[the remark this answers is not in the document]
sam-data-guy-iam: yes, I will change this in globalmetric.  It should be the case because the current implementation can still result in NaNs

src/unitxt/metrics.py
```diff
@@ -82,43 +83,100 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
         if not self._can_compute_confidence_intervals(num_predictions=len(instances)):
             return result
-
-        score_names = (
-            self.ci_scores if self.ci_scores is not None else [self.main_score]
-        )
+        func_name = str(func_name)
+        if aggregation_func is None:
+            # by default mean aggregation
+            aggregation_func = self.aggregate_instance_scores
```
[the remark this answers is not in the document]
sam-data-guy-iam: from what I have read this isn't technically possible because of differing evaluation times, so the convention is to have None and then detect that.

src/unitxt/metrics.py
```diff
@@ -151,30 +207,10 @@ def metric(sample_refs, sample_preds, sample_additional_inputs):
                 arr=arr,
             )
 
-            # when running with bca interval (default), the statistic is called twice: with the
-            # original data and with the resamples. here we want to focus only on the latter.
-            if scores.size > 1:
-                # here we deal with samples on which the metric could not be computed. These are
```
matanor, line 157: Maybe maintain this comment in `resample_from_non_nan`? WDYT? it explains the rationale nicely ..

src/unitxt/metrics.py
```diff
@@ -368,14 +407,127 @@ def compute(
...
-    implemented_reductions: List[str] = field(default_factory=lambda: ["mean"])
+    implemented_reductions: List[str] = field(
+        default_factory=lambda: ["mean", "group_mean"]
+    )
+
+    # InstanceMetric currently allows two reductions, 'mean', which calculates the mean of instance scores,'
+    # and 'group_mean', which first applies an aggregation function specified in the reduction_map
+    # to instance scores grouped by the field grouping_field (which must not be None), and returns the mean
+    # of the group scores; if grouping_field is None, grouping is disabled.
+    # see _validate_group_mean_reduction for an example and proper formatting of the reduction_map
```
matanor, line 418: thanks for adding, very clear !

src/unitxt/metrics.py
```diff
@@ -368,14 +407,127 @@ def compute(
...
+            grouping_field = 'group_id'
+            reduction_map = {'group_mean': {'agg_func': ['variance', np.var]}}
+
+        reduction_map must be a dict with
+        - an 'agg_func' field with value being a 2-element list where
+            - 1st element is a string name of the aggregation function (used in naming the CI report)
+            - 2nd element is the callable aggregation function
+        - Optional: 'score_fields' key with list value containing the string names of fields to apply the aggregation to
+            - if not present, the parent class main_score is used.
+        """
```
matanor, line 441: Very clear description, tks for adding it :)

tests/test_metrics.py
```diff
@@ -388,6 +480,130 @@ def test_token_overlap(self):
...
+            outputs = apply_metric(
+                metric=metric,
+                predictions=GROUPED_INSTANCE_PREDICTIONS,
+                references=GROUPED_INSTANCE_REFERENCES,
+                additional_inputs=GROUPED_INSTANCE_ADDL_INPUTS,
+            )
+            self.assertAlmostEqual(
+                target,
+                outputs[0]["score"]["global"]["score"],
+                msg=f"{outputs[0]['score']['global']['score_name']} does not equal the expected value",
```
matanor, line 510:
```suggestion
                msg=f"{outputs[0]['score']['global']['score_name']} does not equal the expected value {target}",
```

matanor: could be nice to have the expected value already in the error message ..

tests/test_metrics.py
```diff
@@ -388,6 +480,130 @@ def test_token_overlap(self):
...
+            outputs = apply_metric(
+                metric=metric,
+                predictions=GROUPED_INSTANCE_PREDICTIONS_SHORT,
+                references=GROUPED_INSTANCE_REFERENCES_SHORT,
+                additional_inputs=GROUPED_INSTANCE_ADDL_INPUTS,
+            )
+            self.assertAlmostEqual(
+                target,
+                outputs[0]["score"]["global"]["score"],
+                msg=f"{outputs[0]['score']['global']['score_name']} does not equal the expected value",
```
matanor, line 525:
```suggestion
                msg=f"{outputs[0]['score']['global']['score_name']} does not equal the expected value {target}",
```

tests/test_metrics.py
```diff
@@ -17,6 +24,91 @@
...
+    ["I am a person", "I AM A DOG", "ABC"],
+    ["AB CD", "AB", "ab"],
+    ["AB 1", "AB1"],
+    [" GMA 123", "GMA"],
+    ["123", 0.12],
+    ["BDE", "BCE", "bdefs"],
+    [" abcdefg", "AB", "abcd"],
+]
+
+GROUPED_INSTANCE_PREDICTIONS_SHORT = [
```
matanor, line 64: a question: why are there two sets of predictions and references (regular an `short`)?

tests/test_metrics.py
```diff
@@ -478,3 +694,129 @@ def _test_confidence_interval(self, metric, expected_ci_low, expected_ci_high):
...
+        if expected_global_result is None:
+            expected_global_result = {
+                f"{group_score_name}_ci_low": expected_ci_low,
+                f"{group_score_name}_ci_high": expected_ci_high,
+                "score_ci_low": expected_ci_low,
+                "score_ci_high": expected_ci_high,
+            }
+
+        global_result = outputs[0]["score"]["global"].copy()
+        logger.info(global_result)
```
matanor, line 796: should remove this logging?

tests/test_metrics.py
```diff
@@ -478,3 +694,129 @@ def _test_confidence_interval(self, metric, expected_ci_low, expected_ci_high):
...
+                    if np.isnan(expected_global_result[score_name]):
+                        assert np.isnan(score_value)
+                    elif np.isnan(score_value):
+                        assert np.isnan(expected_global_result[score_name])
+                    else:
+                        self.assertAlmostEqual(
+                            score_value,
+                            expected_global_result[score_name],
+                            places=5,
+                            msg=f"score mismatch for {group_score_name}",
```
matanor, line 811: suggest to add the value that are mismatched `score_value` and `expected_global_result[score_name]`

tests/test_metrics.py
```diff
@@ -478,3 +694,129 @@ def _test_confidence_interval(self, metric, expected_ci_low, expected_ci_high):
...
+            }
+
+        global_result = outputs[0]["score"]["global"].copy()
+        logger.info(global_result)
+        for score_name, score_value in global_result.items():
+            if score_name in expected_global_result:
+                # Test that the output value is the same as the expected value
+                # allow for cases where value is NaN
+                if not isinstance(score_value, str):
+                    if np.isnan(expected_global_result[score_name]):
```
matanor, line 802: if it isn't expected to have NaN in the tests perhaps this part is not needed? it seems by the coverage tool that these lines are never executed ..

prepare/metrics/grouped_instance_metrics.py
```diff
@@ -0,0 +1,362 @@
...
+    predictions=predictions,
+    references=references,
+    instance_targets=instance_targets_string_containment,
+    global_target=global_target,
+    additional_inputs=additional_inputs,
+)
+
+add_to_catalog(metric, "metrics.group_pdr_string_containment", overwrite=True)
+
+# Try Cohen's h instead of PDR since is symmetric and defined when baseline is 0
```
matanor, line 229: Not sure i understand this comment (?)

matanor (approved):
LGTM! 
left some small items concerning the tests ... please have a look ...

tests/test_metrics.py
sam-data-guy-iam: maybe, but it exists in the other globalmetric confidence interval test so I figured I shouldn't change that

tests/test_metrics.py
sam-data-guy-iam: This was something I did when I tested other precision/recall based scores like F1 which did some kind of averaging over the observed labels, which is not "interesting" if there are not only a few unique labels.  But I just tested it with the original sets, which still yield meaningful values.  So I will change this and the tests to use the same set of predictions.  If we had a F1-type InstanceMetric (as opposed to GlobalMetric), this would be relevant.

src/unitxt/metrics.py
sam-data-guy-iam: I can leave the original comment, you are right.  I don't think we need the conditioning on scores.size > 1 though.

tests/test_metrics.py
sam-data-guy-iam: yes because it doesn't happen in the tests, but it can happed if the distribution is degenerate (a single unique value), which can happen even if there are no NaNs in the original scores.  I mentioned this in the resampling method.  I guess I don't need it in the test if it never happens, but it is something that can happen (since the equality testing doesn't work if there are NaNs).

prepare/metrics/grouped_instance_metrics.py
sam-data-guy-iam: It is just an alternative to the above metric which is undefined in many cases, and the CI and aggregates would ignore these in the computation, which is losing information.

sam-data-guy-iam: I can delete the comment, I just made an additional note in the class documentation.  It doesn't belong here

src/unitxt/metrics.py
matanor: Do you know why is was there in the first place?
I see this comment: 

# when running with bca interval (default), the statistic is called twice: with the
# original data and with the resamples. here we want to focus only on the latter.

seems it is to prevent NaN removal when running on the original data (without resamples). But not sure i understand why. Do you understand this part?

tests/test_metrics.py
matanor: I see.. since its test code i suggest either removing it or, if we want to be very rigorous, add test values that cause NaN values to appear, and then this code will be used..

prepare/metrics/grouped_instance_metrics.py
matanor: ok, sounds good

sam-data-guy-iam:
Add additional reduction to InstanceMetric for PDR and Cohen, and two ways of sampling group scores, one by group unit and one by instance.

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+                    "instance": {
+                        field_name: group_aggregation_func(scores[field_name])
+                        for field_name in field_names
+                    }
+                }
+            }
+            for scores in group_to_instance_scores.values()
+        ]
+
+    def get_group_aggregated_instance_scores_by_subgroups(
```
matanor, line 748:
```suggestion
    def get_group_scores_by_subgroups(
```

matanor: more readable IMO, WDYT?

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
 
-        for reduction, fields in self.reduction_map.items():
-            assert (
-                reduction in self.implemented_reductions
-            ), f"Reduction {reduction} is not implemented, use one of {self.implemented_reductions}"
+        return instances, global_score
 
-            if reduction == "mean":
-                from statistics import mean
+    def get_group_aggregated_instance_scores(
```
matanor, line 704:
```suggestion
    def get_group_scores(
```

matanor: more readable IMO, WDYT?

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+            additional_inputs = instance["additional_inputs"]
+            for cc in [self.grouping_field, subgroup_field]:
+                if cc not in additional_inputs:
+                    raise ValueError(
+                        f"Missing '{cc}' from instance {instance}. "
+                        f"This field is required for group based metric computation."
+                    )
+            group_key = additional_inputs[self.grouping_field]
+            # indicator if is in the baseline group or not
+            is_baseline = str(additional_inputs[subgroup_field]) == baseline_name
```
matanor, line 784: suggestion: maybe have `is_baseline` directly defined on the instance, next to the `group_id`? isn't this already in the data? that way you will only need to define the is_baseline field name. i think you even can just fix it to some value, maybe its name can be `is_group_baseline`?
I think its nice that these things can be customized, but it makes things a bit difficult to follow ..

matanor: Similarly you could say that if one want to do grouping, it should simply set the `group_id` field, without having to specify the group id field name. Meaning, just fix `grouping_field = "group_id"`.. how does that sound?

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+            # convert True (baseline) to 0, and False (others) to 1, store in respective groups
+            idx = int(not is_baseline)
+
+            for field_name in field_names:
+                group_to_instance_scores[group_key][field_name][idx].append(
+                    instance["score"]["instance"][field_name]
+                )
+
+        # now for each group, take the aggregation function, comparing others to baseline
+        return [
```
matanor, line 794: Here, i things it would be best to just return a list of group scores (where each list item is a dict with mapping from  `field_name` to the score value).
I realize that would mean making other changes since the current code expects the score maps to be within `instance["score"]["instance"]`.. but since we are now dealing with groups that convention becomes very confusing..

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+            {
+                "score": {
+                    "instance": {
+                        field_name: group_aggregation_func(
+                            baseline_scores=scores[field_name][0],
```
matanor, line 799: was wondering if perhaps the separation between grouping with subgroups and without subgroups could be avoided: 
1. assume we always have subgroups (baselines and other scores). 
2. The dataset will define what are the baselines and what are the "others" (as coded now). In some cases -- there will be no baselines defined in the dataset. 
3. The `group_aggregation_func`s API will always accept `baseline_scores` and `other_scores`, but it should verify the input is as expected. For example, if we just do mean aggregation for a group, the function should make sure there are no baselines (otherwise assert). If we do PDR, make sure there is a baseline given (otherwise assert), and so on... 

The main advantage of that would be IMO that most of the code can be shared between the two group aggregation options, and it would be more easy to follow.. At the cost of some added verification code for the group aggregation functions.

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+        # if treat groups as units
+        if reduction_params["agg_func"][2]:
+            # pass the group aggregate---not instance---scores to resample as usual
+            aggregation_function = self.average_instance_scores
+            instances_to_resample = self.get_group_aggregated_instance_scores(
+                instances, reduction_fields, reduction_params["agg_func"][1]
+            )
+        else:
+            # pass the instance scores to resample, and calculate the group aggregation on the resamplings
+            instances_to_resample = deepcopy(instances)
```
matanor, line 821: why is the copy needed?

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+                    instances,
+                    reduction_fields,
+                    group_aggregation_func=reduction_params["agg_func"][1],
+                    subgroup_field=reduction_params["subgroups"][0],
+                    baseline_name=reduction_params["subgroups"][1],
+                )
+            )
+        else:
+            # pass the instance scores to resample, and calculate the group aggregation on the resamplings
+            instances_to_resample = deepcopy(instances)
```
matanor, line 855:
```suggestion
            items_to_resample = deepcopy(instances)
```

matanor: at this point its either groups or instances so i suggest to rename..

src/unitxt/metrics.py
```diff
@@ -400,31 +699,180 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+        return instances_to_resample, aggregation_function
+
+    def _set_up_group_mean_subgroup_comparison_aggregation(
+        self, instances, reduction_params, reduction_fields
+    ):
+        # if treat groups as units
+        if reduction_params["agg_func"][2]:
+            # pass the group aggregate---not instance---scores to resample as usual
+            aggregation_function = self.average_instance_scores
+            instances_to_resample = (
```
matanor, line 844: IIUC these are now the `scores_to_resample` right?
since the computation per group is done in `get_group_aggregated_instance_scores_by_subgroups`?

src/unitxt/metrics.py
matanor: or, maybe extract the instance scores here, and call them `scores_to_resample`. 
Since in the upper section of the if statements its group scores.

matanor (changes requested):
hi @sam-data-guy-iam looks very good, impressive.. 
left some comments, mostly i think we should aim to simplify things if possible and make things more readable .. but otherwise looks great!

src/unitxt/metrics.py
```diff
@@ -367,16 +437,245 @@ def compute(
...
+                If False, the instances are resampled individually, and the groups determined
+                (meaning the groups may be of slightly different size or composition from the original
+                depending on the resampling of the instances).
+                For group_mean_subgroup_comparison reduction (see _validate_group_mean_subgroup_comparison_reduction), it's
+                recommended to set it as 'True' to prevent the group score from being NaN too often.
+        - Optional: 'score_fields' key with list value containing the string names of fields to apply the aggregation to
+            - If not present, the parent class main_score is used.
+
+        Verify that instances do not contain a field _index, which is used to sort to make
+        sure that resampling preserves the
```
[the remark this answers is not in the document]
sam-data-guy-iam: Sorry, this should have been deleted

src/unitxt/metrics.py
sam-data-guy-iam: not a bad idea, I guess the user could make these changes in the dataset card, to create this column if it doesn't exist already.

src/unitxt/metrics.py
sam-data-guy-iam: Because in the global calculation I want to simply pass one object, instances_to_resample (or scores_to_resample), which differs based on whether the grouping is applied.  In once case it is simply the instances, in other cases it is the group score applied.

src/unitxt/metrics.py
sam-data-guy-iam: yes: in the _set_up functions, we prepare the inputs to the bootstrapping.  If grouping is requested (reduction_params["agg_func"][2] is True), then the scores are first calculated on the groups, at which point you need to simply average them.  If not, you need to pass the original instance scores (hence the deepcopy) to resampling, and THEN apply the group aggregation function after resampling each time.

src/unitxt/metrics.py
sam-data-guy-iam: Possibly, but the subgroup aggregation functions require additional arguments.  I see that one could check if there is a "is_baseline" column and if not, define it to always be True.  Maybe in these cases it is a good idea to create a copy of the instances so we don't add extra fields to the list that is returned.  But then we have to check the arguments of aggregation functions etc.

src/unitxt/metrics.py
```diff
@@ -367,16 +437,245 @@ def compute(
...
+            reduction_map = {'group_mean': {'agg_func': ['variance', np.var, True]}}
+
+        reduction_map must be a dict with
+        - an 'agg_func' field with value being a 3-element list where
+            - 1st element is a string name of the aggregation function (used in naming the CI report)
+            - 2nd element is the callable aggregation function
+            - 3rd element is a Boolean indicator of whether, during boostrap CI calculation, the groups are to be sampled as single units.
+                If True, the group scores are calculated and then resampled.  This treats the group units as the unit of
+                interest for which the CI is being compared.
+                If False, the instances are resampled individually, and the groups determined
```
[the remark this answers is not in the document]
sam-data-guy-iam: In our specific use case, we have groups that consist of an observation that is the original, and several that are paraphrases of the original sentence.   If the grouping is not enforced (False), then instances are resampled individually and then grouped.  Because the comparison aggregation functions require both baseline and other inputs, this has several effects.  Primarily, if the "original" instance for a given group is not resampled, we can't calculate the function at all.  Therefore to preserve the information, since the unit of interest is "an original sentence and its paraphrase" then we want to resample the units.

src/unitxt/metrics.py
sam-data-guy-iam: I think what I did is okay because it allows the code to be used for both reductions.  Plus the scores are only used internally and not returned in the instance list.

sam-data-guy-iam:
implemented the recommended changes.  One thing that was not changed was to pass the dict of scores without the score:instance keys.  This is because we need the additional_inputs (including is_baseline and group_id columns) as well to aggregate the scores.   Therefore the bootstrapping needs to receive the full instances or the group scores.

sam-data-guy-iam:
implemented recommended changes.  if aggregation function does some comparison or calculation using scores of one or more variant type, checks for a variant_type column.  This allows use of multiple metrics on the same run when there are multiple >2 variant types.

matanor (changes requested):
hi @sam-data-guy-iam please see my comments.
the two main things in my view are:
1. The use of an internal data structure to represent the inputs to the aggregation functions. why not just pass the instances to the aggregation functions?
2. There are many hard coded values for the split between the original texts and the paraphrases, i think it would be nice to have something more generic. if the split is always to two groups, then i suggest having a field with a fixed name and a boolean flag. currently we have hard coded the variant type field name, and the two explicit fields which refer to the original and the paraphrases .. i realize that suits the current use case, but perhaps may be a bit generalized or simplified ..

src/unitxt/metrics.py
```diff
@@ -366,16 +437,201 @@ def compute(
...
+                self._validate_group_mean_reduction(instances=instances)
+                reduction_fields = (
+                    [self.main_score]
+                    if "score_fields" not in reduction_params
+                    else list(set(reduction_params["score_fields"]))
+                )
+                aggregation_function_name = str(reduction_params["agg_func"][0])
+                field_name_full_prefix = "group_" + aggregation_function_name + "_"
+                if reduction_params["agg_func"][2]:
+                    field_name_full_prefix = "fixed_" + field_name_full_prefix
```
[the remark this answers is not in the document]
sam-data-guy-iam: because it indicates the the groups are resampled as fixed units (when reduction_params["agg_func"][2] is True), rather than resampled as instances and then grouped.  I wanted to differentiate the two.

src/unitxt/metrics.py
```diff
@@ -399,31 +655,107 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
-                global_score.update(confidence_interval)
+        Returns:
+            List of dicts, each corresponding to a group of instances (defined by grouping_field),
+                with a group score for each field_name
+        """
+        from collections import defaultdict
+
+        # three-level defaultdict:
+        # first is the grouping, second is the field name, the third is the variant_type (by default 'original')
+        group_to_instance_scores = defaultdict(
```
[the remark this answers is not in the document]
sam-data-guy-iam: because group_aggregation_function operates on each group scores individually.  It is aggregating the scores in a group, not aggregating the overall scores of multiple groups.  My aim is that people do not have to specify the grouping within this aggregation function.

src/unitxt/metrics.py
```diff
@@ -82,43 +96,116 @@ def _can_compute_confidence_intervals(self, num_predictions):
...
         for score_name in score_names:
-            scores = [
-                instance["score"]["instance"][score_name] for instance in instances
-            ]
+            # need to redefine the statistic function within the loop because score_name is a loop variable, to avoid ruff errors
+            def statistic(arr, axis, score_name=score_name):
+                # arr is a 2d array where each row is a resampling, so we
+                # iterate over the rows and compute the metric on each resampling
+
+                # if aggregation_func is None, we simply take the mean of the resampled instance scores
```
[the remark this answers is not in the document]
sam-data-guy-iam: The "arr is a 2d array..." comment is taken from the global metric confidence intervals, from which I adapted the code.  I moved the comment about aggregation_func above to where it is set when it is None.  I think it helps.

src/unitxt/metrics.py
```diff
@@ -366,16 +437,201 @@ def compute(
...
+            assert (
+                reduction_type in self.implemented_reductions
+            ), f"Reduction {reduction_type} is not implemented, use one of {self.implemented_reductions}"
+
+            field_name_full_prefix = ""
+            # used for passing to the bootstrapping, depends on whether the groups are fixed or not
+            aggregation_function = self.average_item_scores
+            if reduction_type == "mean":
+                reduction_fields = list(set(reduction_params))
+                # extract only the dict of instance scores
```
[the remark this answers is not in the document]
sam-data-guy-iam: changed the comment

src/unitxt/metrics.py
```diff
@@ -399,31 +655,107 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+        return [
+            {
+                "score": {
+                    "instance": {
+                        field_name: agg_func(score_dict)
+                        for field_name, score_dict in scores.items()
+                    }
+                }
+            }
+            for scores in group_to_instance_scores.values()
```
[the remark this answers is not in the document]
sam-data-guy-iam: accepted

src/unitxt/metrics.py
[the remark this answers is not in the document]
matanor: i see, sounds good. maybe just a small suggestion, add another word like `fixed_groups_` or `sampled_groups_`. If you agree it improves readability.

sam-data-guy-iam:
made a minor change to Cohen's h, Cohen's d, and PDR metrics to allow a group to consist of multiple labels (e.g. if we want the comparison group to be all variants in a list).

elronbandel:
First time reviewing this and i have a general question, for both @matanor and @sam-data-guy-iam, we currently have grouping mechanism within unitxt, using the group field. why dont we have all metrics in unitxt by definition work on groups with predefined aggregation function per metric? or am i missing something?

sam-data-guy-iam:
> First time reviewing this and i have a general question, for both @matanor and @sam-data-guy-iam, we currently have grouping mechanism within unitxt, using the group field. why dont we have all metrics in unitxt by definition work on groups with predefined aggregation function per metric? or am i missing something?

Thank you for pointing out this field in the schema, I was unaware of it before (perhaps Matan too).  I am looking now at test_examples.test_add_metric_to_catalog and this seems to be doing part of what our modification does.

src/unitxt/metrics.py
```diff
@@ -1790,3 +2120,469 @@ def get_element_representation(self, element, additional_input):
...
+
+    # make sure the expected types appear
+    subgroup_scores_dict.update(
+        {kk: [] for kk in expected_subgroup_types if kk not in subgroup_scores_dict}
+    )
+    return subgroup_scores_dict
+
+
+def performance_drop_rate(
+    subgroup_scores_dict: Dict[str, List], expected_subgroup_types: List[List[str]]
```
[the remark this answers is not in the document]
matanor: since there are always two lists compared, why not have them explicitly in the signature as two params?

matanor (changes requested):
hi @sam-data-guy-iam , maded another pass over metrics.py, left some comments, mostly small things at this point.. pls have a look
one thing which i am not sure i understand is why the `deepcopy` of the instances is done..

matanor:
> First time reviewing this and i have a general question, for both @matanor and @sam-data-guy-iam, we currently have grouping mechanism within unitxt, using the group field. why dont we have all metrics in unitxt by definition work on groups with predefined aggregation function per metric? or am i missing something?

@elronbandel I was not aware of that mechanism.. we also have the `reduction_map` items within the actual metric classes, so we followed that .. regardless of this PR, why do we have these two mechanisms? don't they do the same thing?

i agree with @sam-data-guy-iam that there are features that Sam has contributed that IMO would be difficult to do without directly coding them into one operator (i.e. the metric).. e.g. computing the confidence intervals in various ways, and also doing the aggregation with all kinds of options.. @elronbandel we can perhaps set up some time to discuss, WDYT?

src/unitxt/metrics.py
```diff
@@ -371,16 +441,197 @@ def compute(
...
+                reduction_type in self.implemented_reductions
+            ), f"Reduction {reduction_type} is not implemented, use one of {self.implemented_reductions}"
+
+            field_name_full_prefix = ""
+            # used for passing to the bootstrapping, depends on whether the groups are fixed or not
+            aggregation_function = self.average_item_scores
+            if reduction_type == "mean":
+                reduction_fields = list(set(reduction_params))
+                # no group reduction, so resample instances individually
+                scores_to_resample = deepcopy(instances)
```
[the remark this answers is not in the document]
sam-data-guy-iam: to avoid linking the two objects in case of modification, not sure if it has an effect or not

src/unitxt/metrics.py
```diff
@@ -1790,3 +2120,469 @@ def get_element_representation(self, element, additional_input):
...
+    else:
+        # expected_subgroup_types could be a list of lists; now take unique values
+        from itertools import chain
+
+        expected_subgroup_types = list(
+            set(chain.from_iterable(expected_subgroup_types))
+        )
+
+    # make sure the expected types appear
+    subgroup_scores_dict.update(
```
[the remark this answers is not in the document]
sam-data-guy-iam: you are right, it already is so this check doesn't need to be done

src/unitxt/metrics.py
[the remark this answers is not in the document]
sam-data-guy-iam: good idea, I changed it

elronbandel (approved):

src/unitxt/metrics.py
```diff
@@ -1790,3 +2120,469 @@ def get_element_representation(self, element, additional_input):
 
     def should_ignore_element(self, element, additional_input):
         return element == "none"
+
+
+# define metrics that return means of an aggregation function applied across levels of a grouping variable
```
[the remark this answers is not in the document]
sam-data-guy-iam: I was trying to separate the material that has to do with the group_mean reduction from other classes, but I can remove it.

src/unitxt/metrics.py
```diff
@@ -1790,3 +2120,469 @@ def get_element_representation(self, element, additional_input):
...
+def validate_subgroup_types(
+    subgroup_scores_dict: Dict[str, List], expected_subgroup_types: List
```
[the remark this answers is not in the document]
sam-data-guy-iam: accepted

src/unitxt/metrics.py
```diff
@@ -404,31 +655,110 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+        )
 
-                confidence_interval = self.score_based_confidence_interval(
-                    instances=instances
-                )
-                global_score.update(confidence_interval)
+        # check if function has fields for subgroup_column
+        uses_subgroups = self.subgroup_column is not None
+        if uses_subgroups:
+            assert all(
```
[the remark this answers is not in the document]
sam-data-guy-iam: yes, I can remove it

sam-data-guy-iam:
> hi @sam-data-guy-iam , maded another pass over metrics.py, left some comments, mostly small things at this point.. pls have a look one thing which i am not sure i understand is why the `deepcopy` of the instances is done..

I used deepcopy to copy the instances because I didn't want to tie the resamples list to instances by assignment because we later modify the instances, and I didn't want there to be unexpected results (have had this happen in the past).

src/unitxt/metrics.py
```diff
@@ -404,31 +655,110 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+        return [
+            {
+                "score": {
+                    "instance": {
+                        field_name: agg_func(score_dict)
+                        for field_name, score_dict in group_scores.items()
+                    }
+                }
+            }
+            for group_scores in group_to_instance_scores.values()
```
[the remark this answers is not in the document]
matanor: I was originally thinking of two nested regular loops, but your suggestion above is better..

matanor:
> > hi @sam-data-guy-iam , maded another pass over metrics.py, left some comments, mostly small things at this point.. pls have a look one thing which i am not sure i understand is why the `deepcopy` of the instances is done..
> 
> I used deepcopy to copy the instances because I didn't want to tie the resamples list to instances by assignment because we later modify the instances, and I didn't want there to be unexpected results (have had this happen in the past).

@sam-data-guy-iam 
I see.. i think this can go both ways: you could make accidental changes to the instances, and then the deepcopy prevents that. But at some point others might be making intentional changes to the instances, and then the deepcopy would prevent these from taking effect on the original instances.
IMO, if you know now that you are making updates to the instances, and you don't want them reflected on the original instances, use deepcopy. but otherwise, i am not in favor of doing so, you might be preventing future issues, but you might also be causing future issues..

sam-data-guy-iam:
> > > hi @sam-data-guy-iam , maded another pass over metrics.py, left some comments, mostly small things at this point.. pls have a look one thing which i am not sure i understand is why the `deepcopy` of the instances is done..
> > 
> > 
> > I used deepcopy to copy the instances because I didn't want to tie the resamples list to instances by assignment because we later modify the instances, and I didn't want there to be unexpected results (have had this happen in the past).
> 
> @sam-data-guy-iam I see.. i think this can go both ways: you could make accidental changes to the instances, and then the deepcopy prevents that. But at some point others might be making intentional changes to the instances, and then the deepcopy would prevent these from taking effect on the original instances. IMO, if you know now that you are making updates to the instances, and you don't want them reflected on the original instances, use deepcopy. but otherwise, i am not in favor of doing so, you might be preventing future issues, but you might also be causing future issues..

fair enough.  I will remove them.

src/unitxt/metrics.py
```diff
@@ -372,18 +441,200 @@ def compute(
...
+        of the group scores; if grouping_field is None, grouping is disabled.
+        See _validate_group_mean_reduction for formatting instructions.
+    """
+
     n_resamples: int = OptionalField(
         default_factory=lambda: settings.num_resamples_for_instance_metrics
     )
 
-    implemented_reductions: List[str] = field(default_factory=lambda: ["mean"])
+    # column required to be in additional_inputs if group_mean aggregation function requires a dict input of labels and their lists of scores
```
matanor, line 458: can this comments be clarified? simplified? its a bit unclear ..

prepare/metrics/grouped_instance_metrics.py
```diff
@@ -0,0 +1,553 @@
...
+# these groupings correspond in length to the group identifiers above
+variant_type = np.concatenate(
+    [
+        np.repeat(a=["original", "paraphrase"], repeats=reps)
+        for reps in [[1, 4], [1, 4], [1, 3], [1, 0]]
+    ]
+).tolist()
+
+# construct grouping_field by combining two other fields (and ignoring one); mimics what you would do in cards
+group_by_fields = ["group", "id"]
```
matanor, line 79: generally i think the tests should be as simple and readable as possible..
so i suggest to directly set the `group_id` from the values above, to simplify the tests.. and the same for the `variant_type` field..

prepare/metrics/grouped_instance_metrics.py
```diff
@@ -0,0 +1,553 @@
...
+for ai, vt in zip(additional_inputs, variant_type):
+    ai.update(
+        {
+            "group_id": "_".join([str(ai[ff]) for ff in group_by_fields]),
+            "variant_type": vt,
+        }
+    )
+
+
+instance_targets_string_containment = [
```
matanor, line 90: why is this a dict and not just a list?

prepare/metrics/grouped_instance_metrics.py
```diff
@@ -0,0 +1,553 @@
...
+    {"score": 0.0},
+    {"score": 0.0},
+]
+
+for instance in instance_targets_string_containment:
+    instance.update(
+        {"string_containment": instance["score"], "score_name": "string_containment"}
+    )
+
+instance_targets_accuracy = [
```
matanor, line 113: same here, isn't a list of scores enough?

prepare/metrics/grouped_instance_metrics.py
```diff
@@ -0,0 +1,553 @@
...
+    {"precision": 0, "recall": 0, "f1": 0, "score": 0, "score_name": "f1"},
+    {"precision": 0, "recall": 0, "f1": 0, "score": 0, "score_name": "f1"},
+]
+
+
+metric = GroupMeanTokenOverlap()
+
+outputs = test_metric(
+    metric=metric,
+    predictions=[str(vv) for vv in predictions],
```
matanor, line 546: why the list comprehension loop?

prepare/metrics/grouped_instance_metrics.py
```diff
@@ -0,0 +1,553 @@
...
+    references=[[str(vvv) for vvv in vv] for vv in references],
```
matanor, line 547: why the list comprehension loop?

tests/library/test_metrics.py
```diff
@@ -388,6 +471,115 @@ def test_token_overlap(self):
         for target, value in global_targets.items():
             self.assertAlmostEqual(value, outputs[0]["score"]["global"][target])
 
+    def test_grouped_instance_metrics(self):
```
matanor, line 474: if these are the same tests as in [grouped_instance_metrics.py](https://github.com/IBM/unitxt/pull/452/files?file-filters%5B%5D=.py&show-viewed-files=true#diff-a3f7cbec6d34ae3240ec3ad5e6198dc6f2ccc1ab9edb5fb47bcda4595e246285), then its not needed to do this twice... the grouped_instance_metrics.py runs as part of the tests, so its ok to assume that repeating what it does is not needed..

matanor: they are the same tests, right?

tests/library/test_metrics.py
```diff
@@ -478,3 +670,163 @@ def _test_confidence_interval(self, metric, expected_ci_low, expected_ci_high):
...
+                "group_mean_f1_ci_low": 0.22302503471948287,
+                "group_mean_f1_ci_high": 0.6805555555555555,
+                "score_ci_low": 0.22302503471948287,
+                "score_ci_high": 0.6805555555555555,
+                "group_mean_precision_ci_low": 0.20949399775845196,
+                "group_mean_precision_ci_high": 0.6666666666666666,
+            },
+        )
+
+    def _test_grouped_instance_confidence_interval(
```
matanor, line 782: Same comment as above:

if these are the same tests as in [grouped_instance_metrics.py](https://github.com/IBM/unitxt/pull/452/files?file-filters%5B%5D=.py&show-viewed-files=true#diff-a3f7cbec6d34ae3240ec3ad5e6198dc6f2ccc1ab9edb5fb47bcda4595e246285), then its not needed to do this twice... the grouped_instance_metrics.py runs as part of the tests, so its ok to assume that repeating what it does is not needed..

matanor (changes requested):
hi @sam-data-guy-iam , pls see my comments regarding the tests. 
also left some very minor suggestions for the non-test code

tests/library/test_metrics.py
sam-data-guy-iam: Elron said this is ok

tests/library/test_metrics.py
sam-data-guy-iam: Elron said this is ok

prepare/metrics/grouped_instance_metrics.py
sam-data-guy-iam: To make them strings which TokenOverlap expects, but I'm going to do the string conversion within TokenOverlap instead

prepare/metrics/grouped_instance_metrics.py
sam-data-guy-iam: No problem.  I was trying to illustrate an example where there is a group and id field and the desired group_id field is constructed as a combination of these. Maybe it looks to complex.

prepare/metrics/grouped_instance_metrics.py
sam-data-guy-iam: It is like this for all the other metrics tested (see e.g. accuracy.py), since the receives the whole instance score dict and not just the list of scores.  They check that the key names (i.e., the metric name) match as well.

prepare/metrics/grouped_instance_metrics.py
sam-data-guy-iam: It is like this for all the other metrics tested (see e.g. accuracy.py), since the receives the whole instance score dict and not just the list of scores. They check that the key names (i.e., the metric name) match as well.

prepare/metrics/grouped_instance_metrics.py
matanor: got it, thanks!

matanor (approved):
LGTM! thanks @sam-data-guy-iam for this new feature!

PR #456, opened 2024-01-10
Add a new Perturbate operator that is handy for faking prediction as a perturbated version of target or for any other perturbation use cases

dafnapension:
to be employed just one step before metric evaluation


elronbandel (approved):
Now it's very neat and useful operator!

PR #457, opened 2024-01-10
Fix HuggingFace integration by changing logging.py to logging_utils.py to comply with hf dataset conventions

elronbandel:
hugginface datasets dont work with files called logging as it interfere with their internal logging file


yoavkatz (approved):
Not clear why "comply with hf dataset conventions " needs this, but seems ok.

PR #458, opened 2024-01-10
Delete old formats and make SystemFormat the only format exist

elronbandel:


yoavkatz:
Note the test fail with import error on SystemFormat.

Issue #439, opened 2024-01-01
matthews_correlation  returning 0 on perfect correlation


yoavkatz:
Why is this the accepted behavior (strict=False was set a long time ago)?

********************************************************************************
The results of running the main metric in used in the card (matthews_correlation) over simulated predictions that are equal to the references returns a different score than expected.
One would expect a perfect score of 1.0 in this case, but returned metric score was 0.0.
This is flagged as only as a warning because strict=False was set in the call to test_card().The predictions passed to the metrics were:
 ['acceptable', 'acceptable', 'acceptable']

********************************************************************************

dafnapension:
reproducible via prepare.card.cola.py

dafnapension:
It turns out that matthews_correlation, when fed with all predictions and references being the same single class, suspects something, and returns 0.0.  So it is not us.. it is the huggingface implementation.. See some short proof by example:
[MCC.pdf](https://github.com/IBM/unitxt/files/13866533/MCC.pdf)

yoavkatz:
The HF metric calls scikit-learn:



https://huggingface.co/spaces/evaluate-metric/matthews_correlation/blame/0da51560adeb410656ba31b4cd1807c990898398/matthews_correlation.py

from sklearn.metrics import matthews_corrcoef

def _compute(self, predictions, references, sample_weight=None):
        return {
            "matthews_correlation": float(matthews_corrcoef(references, predictions, sample_weight=sample_weight)),
        }
        
[[ Dot it ](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.matthews_corrcoef.html)](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.matthews_corrcoef.html)

yoavkatz:
I think the issue that in v=[0,0,0] or v=[1,1,1]  - there is only a single class.  This is a special case not treated in the implementation.

yoavkatz:
This seems to be a known issue that has a PR , but was not fixed.

https://github.com/scikit-learn/scikit-learn/issues/25258

dafnapension:
scikit's implementation faithfully follows the definition (as there is only TN or only TP, and all other **three** components are 0, hence the result, by definition of matthew_coef, is 0).
The question is whether for our case, when for testing a metric, we 'fake' full hit, or full miss, we should tweak the example..

yoavkatz:
[...]

PR #470, opened 2024-01-12
override Matthews correlation in cases of perfect correlation

dafnapension:
and perfect anti-correlation, when all prediction components are identical, and their number is > 1.
Addressing issue #439 

@yoavkatz , do we really want this? and not return the true MCC ?
https://bmcgenomics.biomedcentral.com/articles/10.1186/s12864-019-6413-7   (the section with the epsilon..

users might compare our results with their own, or any other implementation of Matthews correlation, and might wonder about our anti-definition result.  See also scikit-learn:

sklearn

and exactly as in the example that started this issue:
mcc2


dafnapension:
Closing for now. I think we should not implement this, but rather stick to the conventional MCC.

PR #471, opened 2024-01-13
Improve ExtractMostCommonFieldValues performance by avoiding the creation of the stream-length long list of values instead building values counter on the fly

dafnapension:
avoid creation of the in-memory  stream-length long list of values before counting number of occurrences for each unique value.

Perhaps need to make similar considerations for metrics.


elronbandel (approved):
Great catch and fix!

PR #478, opened 2024-01-15
Naive recipe and thin global evaluators - draf for now, for future consideration

dafnapension:


dafnapension:
For evaluation, the streams are not so long, and operators must not be thin. They can hold a whole stream, lying in main memory. As done by huggingface.

PR #479, opened 2024-01-15
add a card for HF xsum, a summary dataset

dafnapension:


elronbandel (changes requested):
We should add the task first in `tasks.summerization.abstractive` from a separated preparation file.
Then do the same for the templates. also if possible to add few templates. Even better if at least some use the document type field.

dafnapension:
Hi @elronbandel , done.

elronbandel (approved):
Amazing work like usual!

PR #481, opened 2024-01-15
add new templates

OfirArviv:


prepare/templates/qa/contextual.py
```diff
@@ -10,6 +10,15 @@
     overwrite=True,
 )
 
+add_to_catalog(
+    MultiReferenceTemplate(
+        input_format="Question: {question}\nContext: {context}",
+        references_field="answers",
+    ),
+    "templates.qa.contextual.simple_question_first",
```
matanor, line 18: I changed the [qa tasks](https://github.com/IBM/unitxt/blob/main/prepare/tasks/qa/tasks.py) for unitxt, so the names of the qa templates should be adjusted accordingly.

matanor: note the answer field is `answer` for those tasks (unlike plural `answers` in fm-eval), also the name is `contextualized.abstractive` or `contextualized.extractive`.

prepare/templates/summarization/general.py
```diff
@@ -0,0 +1,66 @@
+from src.unitxt.catalog import add_to_catalog
+from src.unitxt.templates import InputOutputTemplate, TemplatesList
+
+add_to_catalog(
+    InputOutputTemplate(
+        input_format="Produce a succinct summary for the following text, extracting the fundamental concepts and "
+        "crucial information.\n Text: {text}",
+        output_format="{target}",
+    ),
+    "templates.summarization.general.formal",
```
matanor, line 10: i guess here too that the template names should match https://github.com/IBM/unitxt/tree/main/prepare/tasks/summerization ? though there is only one task there `tasks.summarization.abstractive`.

matanor (approved):
LGTM, left some comments about adjusting names of templates etc, for unitxt. pls have a look..

prepare/templates/qa/contextual.py
OfirArviv: @matanor @eladven is there even extractive task fort generative models? Asking it to copy the input?

OfirArviv:
@elronbandel there is a conflict with files of the same names. Did other pr added same templates?

PR #484, opened 2024-01-16
Make postprocessors a general operator that operates on 'prediction' and on 'references' which enables the use of every possible operator as postprocessor

dafnapension:
@elronbandel , please see if this is in the direction of what you had in mind, and I will complete for all the other postprocessors, and the needed unit_tests.

### DIFF
--- a/prepare/processors/processors.py
+++ b/prepare/processors/processors.py
@@ -1,5 +1,6 @@
 from src.unitxt import add_to_catalog
 from src.unitxt.logging_utils import get_logger
+from src.unitxt.operator import SequentialOperator
 from src.unitxt.processors import (
     ConvertToBoolean,
     FirstCharacter,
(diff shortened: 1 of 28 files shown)


elronbandel (commented):
I think it looks really simple and straight forward well done!

I would probably do it like this to make it simpler:

```python
add_to_catalog(
     SequentialOperator(
              steps=[
                       TakeFirstNonEmptyLine(
                                 field="prediction", process_every_value=False
                         ),
                         TakeFirstNonEmptyLine(
                                 field="references", process_every_value=True
                         ),
                         
              ]
       ),
     "processors.take_first_non_empty_line",
     overwrite=True,
 )
```
(no need to add `for_prediction_and_every_references` as it implied by the fact it sits in the processors dir which will be changed in the future to postprocessors)

but overall exactly what we were talking about well done!

elronbandel (approved):
Exactly what we talk about! that's a great contribution to unitxt!!

PR #502, opened 2024-01-21
Add few public datasets cards from IBM internal fmeval project

OfirArviv:


elronbandel (changes requested):
it has to be with relative imports 
from unitxt.blocks -> from src.unitxt.blocks

OfirArviv:
> it has to be with relative imports from unitxt.blocks -> from src.unitxt.blocks

oh sorry. I copied-paste it and didn't notice :)

OfirArviv:
@elronbandel toxin require hugginface token for accepting its term. Can you add it? Otherwise we will get an error in the tests.

`datasets.exceptions.DatasetNotFoundError: Dataset 'skg/toxigen-data' doesn't exist on the Hub. If the repo is private or gated, make sure to log in with `huggingface-cli login`. `

elronbandel:
@OfirArviv Lets get rid of earning calls dataset for now, wdyt?

OfirArviv:
> @OfirArviv Lets get rid of earning calls dataset for now, wdyt?

ahead of you :)

elronbandel (approved):

PR #504, opened 2024-01-21
Add multidoc2dial dataset in its abstractive and extractive forms

arielge:
abstractive and extractive versions of the multidoc2dial data (https://huggingface.co/datasets/multidoc2dial)


elronbandel:
Following @dafnapension methodology this is the preferred way to add atomic well defined code operator:
`ExecuteQuery(query="question.split('[SEP]')[0]",  to_field="question")`
Her idea is that it is still one line and very clear was is done to the data. 
This comment is mainly for future viewers :)

elronbandel (approved):

PR #505, opened 2024-01-21
Add regression task and fix stsb to use it

elronbandel:


prepare/templates/regression/bounded/pair.py
```diff
@@ -0,0 +1,24 @@
+from src.unitxt.catalog import add_to_catalog
+from src.unitxt.templates import OutputQuantizingTemplate, TemplatesList
+
+add_to_catalog(
+    OutputQuantizingTemplate(
+        input_format="""
+                   Given this text: '{text1}' and this text: '{text2}', on a scale of {min_value} to {max_value}, what is the {type_of_value} of this texts?
```
yoavkatz, line 7: Why not keep to existing format?

Given this sentence: '{text1}', on a scale of {min_value} to {max_value}, what is the {type_of_relation} to this text  {text2}

prepare/cards/stsb.py
```diff
@@ -15,23 +14,19 @@
...
         ),
+        RenameFields(
+            field_to_field={
+                "sentence1": "text1",
+                "sentence2": "text2",
+                "label": "value",
+            }
+        ),
+        AddFields(
+            fields={"type_of_value": "similarity", "min_value": "1", "max_value": "5"}
```
yoavkatz, line 25: type_of_value -> type_of_relation

prepare/templates/regression/bounded/pair.py
```diff
@@ -0,0 +1,24 @@
...
+    overwrite=True,
+)
+
+add_to_catalog(
+    TemplatesList(
+        [
+            "templates.regression.bounded.pair.simple",
+        ]
+    ),
+    "templates.regression.bounded.pair.all",
```
yoavkatz, line 22: I think it should be more clear:

templates.regression.bounded.pair -> templates.regression.two_texts

Is there a need for an unbounded regression between text?  
.

yoavkatz (changes requested):
Please see separate comments.

prepare/templates/regression/bounded/single.py
```diff
@@ -0,0 +1,24 @@
+from src.unitxt.catalog import add_to_catalog
+from src.unitxt.templates import OutputQuantizingTemplate, TemplatesList
+
+add_to_catalog(
+    OutputQuantizingTemplate(
+        input_format="""
+                   Given this text: '{text1}', on a scale of {min_value} to {max_value}, what is the {type_of_value} of this text?
```
yoavkatz, line 7: type_of_value -> attribute

[ Maybe a better name could be found?  ]

PR #508, opened 2024-01-22
Rag metrics

assaftibm:
- New RAG metrics (details below)
- A new tool for executing evaluation metrics on an input DataFrame, and setting results as new columns in the DataFrame
- Updates to the Perplexity metric

List of added RAG metrics:

- reference-less:

  - context-relevance:

    - metrics.rag.context_relevance  

  - faithfulness:

    - metrics.rag.faithfulness  

    -  metrics.rag.k_precision 
    -  metrics.rag.bert_k_precision  
  - answer-relevance:
     - metrics.rag.answer_relevance

- reference-based:
   - context-correctness:
       - metrics.rag.mrr
       - metrics.rag.map 
   - answer-correctness:
       - metrics.rag.correctness 
       - metrics.rag.recall
       - metrics.rag.bert_recall


prepare/metrics/rag.py
```diff
@@ -174,4 +176,120 @@
...
+
+answer_relevance = MetricPipeline(
+    main_score="score",
+    preprocess_steps=[
+        CopyFields(field_to_field=[("question", "references")], use_query=True),
+        CopyFields(
+            field_to_field=[("answer", "prediction")],
+            use_query=True,
+        ),
+        ListFieldValues(fields=["references"], to_field="references"),
```
matanor, line 291: why is this needed?

matanor (approved):
LGTM, left just one small question

prepare/metrics/rag.py
```diff
@@ -174,4 +176,120 @@
...
+            use_query=True,
+        ),
+    ],
+    metric="metrics.perplexity_q.flan_t5_small",
+)
+add_to_catalog(context_relevance, "metrics.rag.context_relevance", overwrite=True)
+
+for new_catalog_name, base_catalog_name in [
+    ("metrics.rag.faithfulness", "metrics.token_overlap"),
+    ("metrics.rag.k_precision", "metrics.token_overlap"),
```
matanor, line 231:
```suggestion
    ("metrics.rag.precision_at_k", "metrics.token_overlap"),
```

matanor: a bit clearer IMO. btw, how is the K determined?

prepare/metrics/rag.py
```diff
@@ -174,4 +176,120 @@
...
+    ("metrics.rag.bert_k_precision", "metrics.bert_score.deberta_xlarge_mnli"),
```
matanor, line 232:
```suggestion
    ("metrics.rag.bert_precision_at_k", "metrics.bert_score.deberta_xlarge_mnli"),
```

prepare/metrics/rag.py
matanor: or maybe this isn't standard precision_at_k? i see in the comment above `metrics.rag.precision@K` is also listed

assaftibm: it's "knowledge precision" - so k_precision is the common term for this.

prepare/metrics/rag.py
assaftibm: because this metric compares the answer (as the prediction) to the question (as the reference), so we have to wrap the question by a list, since otherwise it will be a string, and references are expected to be lists

prepare/metrics/rag.py
assaftibm: see above, it's k_precision because the full name is "knowledge precision"

prepare/metrics/rag.py
matanor: got it, i will add this explanation as a comment in the code.. tks!

PR #509, opened 2024-01-22
add human eval dataset

OfirArviv:


dafnapension:
Hi @OfirArviv and @elronbandel ,
I managed somehow to make this work.
I am still puzzled why this form of query worked and not the original of Ofir. This may be explained by the shakyness of eval
More than that I am puzzled why I did not manage to pass the imports_list, as I did manage in the tests here:  https://github.com/IBM/unitxt/pull/542
Something deep in the artifact would not recognize ExecuteQuery with yet another argument.
It failed on line 15 of human_eval
@elronbandel , you will know.

**Be careful to not merge this version of ExecuteQuery.** We can revisit here after @elronbandel approves #542, maybe then the strange phenomenon will go away.

dafnapension (approved):

PR #518, opened 2024-01-28
Update catalog navigation Index on documentation side menu for easier catalog browsing

matanor:
Update catalog directory contents, to enable:
- The TOC panel on the left
- The breadcrumbs on the top

Also:
- This change avoids most of the warnings while creating the docs. 
- Fixed the length of the heading marker under the "Explanation for X" title, now its aligned with the title length.

To make this work, all the catalog RST files are now created in the main doc directory (i.e. not in subdirs), and their name is prefixed with `catalog.`.

image


elronbandel (approved):
Looks amazing!

PR #528, opened 2024-01-29
Tweaking Catalog.py some, and fixed some typos. Also in the .rst fixed pages. Generate explanations also for sub-types in the artifact

dafnapension:


dafnapension:
I think that the html renderer ignores similar references sitting inside  code-block:: json  
@elronbandel , perhaps you have an idea.

docs/catalog.py
```diff
@@ -22,6 +22,31 @@ def custom_walk(top):
             yield entry
 
 
+def all_subtypes_of_artifact(artifact):
```
matanor, line 25: perhaps add type hints? is `artifact` an artifact type or an actual artifact?

docs/catalog.py
```diff
@@ -22,6 +22,31 @@ def custom_walk(top):
...
+    if isinstance(artifact, list):
+        to_return = []
+        for art in artifact:
+            to_return.extend(all_subtypes_of_artifact(art))
+        return to_return
+    # artifact is a dict
+    to_return = []
+    for key, value in artifact.items():
+        if isinstance(value, str):
+            if key == "type":
```
matanor, line 43: just from reading the code i am not sure what is going on .. how can `artifact` become a dict? if its passed a dict in the first call? what does `all_subtypes_of_artifact` mean? maybe a docstring may help ..

docs/catalog.py
```diff
@@ -44,10 +69,23 @@ def make_content(artifact, label, all_labels):
...
+            result += "+" * len(explanation_str) + "\n\n"
+            result += subtype_class.__doc__ + "\n"
+
     references = []
     for label in all_labels:
-        if f'"{label}"' in result:
-            references.append(f":ref:`{label} <{label}>`")
+        label_no_catalog = label[8:]  # skip over the prefix 'catalog.'
+        if f'"{label_no_catalog}"' in result:
+            references.append(f":ref:`{label_no_catalog} <{label}>`")
```
matanor, line 88: good catch!

docs/catalog.py
```diff
@@ -44,10 +69,23 @@ def make_content(artifact, label, all_labels):
...
         result += artifact_class.__doc__ + "\n"
 
+    subtypes = all_subtypes_of_artifact(artifact)
+    subtypes = list(set(subtypes))
+    subtypes.remove(artifact_type)  # this was already documented
+    for subtype in subtypes:
+        subtype_class = Artifact._class_register.get(subtype)
+        subtype_class_name = subtype_class.__name__
+        if subtype_class.__doc__:
+            explanation_str = f"Explanation about `{subtype_class_name}`"
```
matanor, line 79: suggestion: i think this is based on another place where an explanation is added to the rst text? if so, and its the same code in the other place, maybe consider making it a small function ..

src/unitxt/catalog/recipes/wnli_fixed.json
```diff
@@ -47,12 +47,13 @@
             ]
         },
         {
-            "type": "render_format_template",
-            "template": {
-                "type": "input_output_template",
-                "input_format": "Given this sentence: {sentence1}, classify if this sentence: {sentence2} is {choices}.",
-                "output_format": "{label}"
-            }
+            "type": "input_output_template",
```
matanor, line 50: just checking -- this is from a re-run of the corresponding prepare script?

matanor (approved):
LGTM, left a few small comments ..

docs/catalog.py
matanor: This is what i am referring to, just above: https://github.com/IBM/unitxt/pull/528/files#diff-380b97fbfb3b8253f06f95f85d25f212caa0ef7b146e079429c8a5a8057f70fcL42

docs/catalog.py
dafnapension: Thanks!  perhaps you know how to push the reference into the main json block, in the right positions? 
E.g., I wanted to replace  "tasks.summarization.abstractive" in [this card](https://unitxt.readthedocs.io/en/latest/catalog.cards.xlsum.english.html), with a reference [to here](https://unitxt.readthedocs.io/en/latest/catalog.tasks.summarization.abstractive.html) but I do not manage to get the html renderer to look inside json blocks. 

Do you know how?

docs/catalog.py
dafnapension: Yes, I was mysterious more than absolutely needed.. The  idea is to pull out from the main artifact in your `make_content(artifact, label, all_labels):` all the inner 
**"type" : "artifact (typically, operator) name as shows in the catalog"**
and to add explanations also on these.  See [example](https://unitxt.readthedocs.io/en/latest/catalog.cards.winogrande.m.html) where not only the outer "task_card" is now explained.

matanor: nice! very useful..

PR #542, opened 2024-01-31
allow imports list for executequery and filterbyquery and rename to ExecuteExpression and FilterByExpression

dafnapension:


dafnapension:
Hi @elronbandel , please help me understand why the second commit "blew up" the card prepare/cards/multidoc2dial.py
the first commit passed flying colors, and the second, which only employed renaming -- blew up.

elronbandel:
I think AssignExpression might give the impression its just adding the expression as a field. I think that if the name was ExecuteExpression is it was more clear the expression is executed. what do you think @yoavkatz @dafnapension ?

dafnapension:
No problem in renaming. I was not sure why the expression is more 'executed' in this operator that it is by FilterByExpression. 
I thought that the difference between those two is by what you do with the value of the expression:
filtering (by the computed expression) vs assigning (the computed expression) to a field.  
The expression is actually 'executed' (i.e. computed) by the Mixin.

src/unitxt/operators.py
```diff
@@ -1215,30 +1216,48 @@ def _is_required(self, instance: dict) -> bool:
...
+class ComputeExpressionMixin(Artifact):
+    """Computes an expression expressed over fields of an instance.
+
+    Args:
+        expression (str): the expression, in terms of names of fields of an instance
+        imports_list (List[str]): list of names of imports needed for the evaluation of the expression
+    """
+
+    expression: str
+    imports_list: List[str] = []
```
elronbandel, line 1228: This is not good. as the imports_list here is assigned at the class level which means all the instances will forver use the same list

Can be fixed by first: `from .dataclass import OptionalField`

Then `imports_list: List[str] = OptionalField(default_factory=list)`

elronbandel (approved):

PR #544, opened 2024-02-01
Added table serializers operators and add Wikitq table question answering dataset

csrajmohan:
Table serializers convert a given table into a flat sequence with special symbols. This data preparation step is essential for running tabular data tasks with LLMs. As part of this PR,

* Created base class for Table serializers and added 2 main table serializers(Indexed Row major, Markdown) as new Unitxt operators.
* Added a new task card for WikiTableQuestions(wikitq) dataset and validated the serializer operators with table input.


elronbandel (approved):
Well done great contribution!

PR #551, opened 2024-02-03
update cards to the way they are prepared now by  prepare/cards/....

dafnapension:


dafnapension:
@elronbandel , I am not sure what is the correct name to use for templates. I think too many subdirectories sit in  `src/unitxt/catalog/templates/qa`

elronbandel:
Fixed in https://github.com/IBM/unitxt/pull/561

PR #557, opened 2024-02-06
Extend the function that evaluate with unitxt metric on external data to new types of data

assaftibm:


elronbandel (changes requested):
Can you explain here what is the purpose of this? also maybe add comment in the code? I think singledispatch s not very common and wont be clear to everyone what you did

assaftibm:
The idea was to overload the "evaluate" function with two different implementations - one that takes a dataframe and the other takes a list of dicts. This overloading can be implemented in different ways, if you prefer something specific I can change.

assaftibm:
This is the formal doc with example - https://docs.python.org/3/library/functools.html#functools.singledispatch

assaftibm:
I added a comment

elronbandel (approved):

PR #560, opened 2024-02-06
Following Elron's fix, remove all special tweaks  

dafnapension:
Following Elron's fix, remove all special tweaks  to not import modules in init, and artificially postpone to process.  After Elron's fix, these tweaks are not needed.


elronbandel (approved):
Wow! thanks for this catch!

PR #562, opened 2024-02-07
Set loaders default to avoid using streaming

elronbandel:


yoavkatz (approved):
Approved.

yoavkatz:
Note this failure:

  File "/home/runner/work/unitxt/unitxt/tests/test_recipe.py", line 135, in test_standard_recipe_with_balancer
    self.assertEqual(counts["entailment"], counts["not entailment"])
AssertionError: 253 != 246

PR #565, opened 2024-02-08
Migrate (task-related) Instruction into Template, and introduce (task independent) SystemPrompt

dafnapension:
the card runs all the way through, and completes with exitcode = 0.

### DIFF
--- a/prepare/cards/20_newsgroups.py
+++ b/prepare/cards/20_newsgroups.py
@@ -8,6 +8,7 @@ from src.unitxt.blocks import (
     TaskCard,
 )
 from src.unitxt.operators import FilterByCondition
+from src.unitxt.templates import InputOutputTemplate
 from src.unitxt.test_utils.card import test_card
 
 dataset_name = "20_newsgroups"
@@ -36,7 +37,7 @@ map_labels = {
 }
 
 card = TaskCard(
-    loader=LoadHF(path=f"SetFit/{dataset_name}"),
+    loader=LoadHF(path=f"SetFit/{dataset_name}", streaming=True),
     preprocess_steps=[
         FilterByCondition(values={"text": ""}, condition="ne"),
         SplitRandomMix(
@@ -47,13 +48,25 @@ card = TaskCard(
         AddFields(
             fields={
                 "classes": list(map_labels.values()),
-                "text_type": "sentence",
+                "text_type": "text",
                 "type_of_class": "topic",
             }
         ),
     ],
     task="tasks.classification.multi_class",
-    templates="templates.classification.multi_class.all",
+    templates=[
+        InputOutputTemplate(  # based on "templates.classification.multi_class.default_no_instruction",
+            input_format="Text: {text}",
+            output_format="{label}",
+            target_prefix="Topic: ",
+            instruction="Classify the {type_of_class} of the following {text_type} to one of these options: {classes}.\n",
+            postprocessors=[
+                "processors.take_first_non_empty_line",
+                "processors.lower_case_till_punc",
+            ],
+        )
+    ],
 )
+
 test_card(card, debug=False)
 add_to_catalog(artifact=card, name=f"cards.{dataset_name}", overwrite=True)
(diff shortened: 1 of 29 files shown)


prepare/cards/20_newsgroups.py
```diff
@@ -53,7 +55,20 @@
         ),
     ],
     task="tasks.classification.multi_class",
-    templates="templates.classification.multi_class.all",
+    templates=[
+        StandardTemplate(
+            input_format="Text: {text}",
+            output_format="{label}",
+            target_prefix="Topic: ",
+            instruction=f"Classify the topic of following text to one of these options: {list(map_labels.values())}.",
```
elronbandel, line 63: I think it should also consume the fields of the task and not to be hard coded: 
```suggestion
            instruction=f"Classify the topic of following text to one of these options: {classes}.",
```
or in whatever way it was in the old templates

dafnapension: A neat idea!  I made instruction a formatting string, in analogy with input_format and output_format. 
tweaked the card some, by adding (temporarily) a few loc-s that print out the prompts (

elronbandel (changes requested):
The StandardTemplate should return instruction not instruction_format. This is because the Template should be the sole owner of the verbalization. 
This will create an issue because the EmptyInstruction will override it. For this reason i suggest that we remove the EmptyInstruction.
This ofcourse will create another issue that we wont have empty instruction for other templates. So my suggestion here is that the old templates will each return empty instruction and empty target prefix. this will ensure backward compatibility and will enable the new behaviour.

dafnapension:
A great idea!  I added, as you suggested, fields "instruction" and "target_prefix" to class Template, and thus no need for a separate StandardTemplate.  All templates behave as you expected.  I removed EmptyInstruction altogether, and took any remains of Instruction from BasicRecipe, and from StandardRecipeWithIndex, and things are still running..
Having passed all tests now, I will dare to remove Instruction from TaskCard, and see if smokes come out from the testing..

src/unitxt/templates.py
```diff
@@ -14,12 +13,21 @@ class Template(StreamInstanceOperator):
     """The role of template is to take the fields of every instance and verbalize it.
 
     Meaning the template is taking the instance and generating source, target and references.
+
+    Args:
+        skip_rendered_instance (bool): if "source", "target", and "references" are already defined fields in the instance, skip its processing
+        postprocessors: a list of strings being artifact names of text processors, to be applied on the model output
+        instruction: a formatting string that yields an instruction with potential participation of values from the "inputs" part of the instance
```
yoavkatz, line 20: There is a delicate issue here.  Since instruction will not be rendered per sample (e.g. in few shot), it can not use  any fields that are not the same of in all instances.   It can use fields which are the same over all instances. 

For example :

Classify the sentiment of this text to one of the following classes: positive, negative.  <-- classes aer common

Text: do do
Sentiment: positive

Text: bla bla
Sentiment: negative

Text: bla bla
Sentiment: negative

Text: foo foo
Sentiment:

yoavkatz: So we should validate, when formatting, that all instructions from the sample and demo instances are the same.

dafnapension: StandardRecipe invokes Operator Template before Operator AddDemosField (see lines 150-160 of [standard.py](https://github.com/IBM/unitxt/pull/565/files#diff-a4122a708bc5f51b81dc0316fb8034bff822104b0e9d82bbe30d44761979ec36) . So the demo instances have the very same instructions as the instances of train, test, and validation. 
Does this answer your concern?

dafnapension: Now looking at this, I wanted to ask you: is this OK that 
```
if self.augmentor.augment_model_input:
            self.steps.append(self.augmentor)
```
occurs after the demos are planted in train and validation and test? the augmentors will not affect the demos that already sit inside each instance.  Is this alright?

elronbandel (approved):

PR #567, opened 2024-02-09
Add new table operators for serialization & truncation

csrajmohan:
* All tabular data related operators are now part of the file table_operators
* Couple of table truncation operators added for limiting tables at cell level and row level to control overall context length
* Table serialization operators moved from serializers.py to table_operators file 
* WikiTableQuestions(wikitq) task card updated to use table truncation operators
* test_table_operators updated to cover test methods for all table operators


elronbandel (changes requested):
Overall looks great. thanks for this really cool contribution!
It seems like you need to add the fake file import also to dataset.py (and remove the old import of serializers). 
Also can you add a reference to the table_operators in the operators file header docstring?

csrajmohan:
> Overall looks great. thanks for this really cool contribution! It seems like you need to add the fake file import also to dataset.py (and remove the old import of serializers). Also can you add a reference to the table_operators in the operators file header docstring?

Done. Thanks.

elronbandel (approved):

PR #568, opened 2024-02-11
Support execution of metrics on a remote host

matanor:
- A metrics http-based service (under service/metrics), optionally supports token-based authentication.
- A new dockerfile and relevant commands in makefile which exemplify how to containerize a unitxt metrics service.
- The evaluate(..) tool supports usage of remote metrics (only `MetricPipeline` objects at this point).
- The remote metrics client configuration is loaded from environment variables.

New requirements:
- httpretty - for mocking http servers, in tests


elronbandel:
If you want pepole to use it when installing unitxt it has to sit in the src/unitxt/service
Then i would reccomend to add its depedncies to .rqr file called requirements/service.rqr
Lastly i would add in the setup.py a predfined action unitxt-metric-service  that can allow to use it from the command line? 
what do you think @matanor ?

matanor:
> If you want pepole to use it when installing unitxt it has to sit in the src/unitxt/service

OK, i will move it. Just to make sure i understand, when users `pip install unitxt`, they only get the code which is under `src/unitxt`? e.g. they don't get all the `prepare` scripts?

> Then i would reccomend to add its depedncies to .rqr file called requirements/service.rqr

OK, will update

>  Lastly i would add in the setup.py a predfined action unitxt-metric-service that can allow to use it from the command line?

You mean an action that starts the service? (start it locally on the machine that it runs on?)
Can you point me to such an existing action? You meant they are in [this](https://github.com/IBM/unitxt/blob/main/setup.py) file?

matanor:
@elronbandel re moving the new `pydantic` dependency from `base.rqr` to a new `service.rqr` file -- 

The thing is that `pydantic.BaseModel` is used as a base class for two new classed in `metric_utils` ([here](https://github.com/IBM/unitxt/blob/service/src/unitxt/metric_utils.py#L167) and [here](https://github.com/IBM/unitxt/blob/service/src/unitxt/metric_utils.py#L173)). So it is imported in the module level [here](https://github.com/IBM/unitxt/blob/service/src/unitxt/metric_utils.py#L6). If i move the dependency, then this module becomes unusable, right?

If this module was only used within a function, i would have just made a local import. Do you know if there is any similar alternative for the above case?

requirements/base.rqr
```diff
@@ -4,3 +4,4 @@ mecab-python3
 absl-py
 dpath
 ipadic
+pydantic==2.6.0
```
yoavkatz, line 7: I think  should move this to a new requirements/service.rqr file, because people who don't need the service, should be be required to use it.

service/metrics/Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+FROM registry.access.redhat.com/ubi8/ubi:latest
+
+# Disable Red Hat Subscripstion
+RUN sed -i 's/1/0/g' /etc/yum/pluginconf.d/subscription-manager.conf
+
+# Set up conda env var
+ENV CONDA_HOME=/opt/conda \
+    PATH=/opt/conda/bin:/usr/local/nvidia/bin:$PATH \
+    CONFIG_DIR=/root/config_dir \
+    BUILD_DIR=/tmp/wml_report \
```
yoavkatz, line 12: Why WML?

service/metrics/requirements.txt
```diff
@@ -0,0 +1,4 @@
+pydantic==2.6.0
```
yoavkatz, line 1: Shy is this requirement file here?

requirements/base.rqr
matanor: yeah that's what @elronbandel wrote too. the thing is that when i try to move it i encounter an issue i mentioned above: https://github.com/IBM/unitxt/pull/568#issuecomment-1945599423
any suggestion how to workaround that?

service/metrics/Dockerfile
matanor: fixed, changed to `BUILD_DIR=/tmp/unitxt_metric_service`

assaftibm (approved):

service/metrics/requirements.txt
matanor: Removed the pydantic dependency from this file, indeed its not needed for the service.
Is this what you meant?

matanor:
> @elronbandel re moving the new `pydantic` dependency from `base.rqr` to a new `service.rqr` file --
> 
> The thing is that `pydantic.BaseModel` is used as a base class for two new classed in `metric_utils` ([here](https://github.com/IBM/unitxt/blob/service/src/unitxt/metric_utils.py#L167) and [here](https://github.com/IBM/unitxt/blob/service/src/unitxt/metric_utils.py#L173)). So it is imported in the module level [here](https://github.com/IBM/unitxt/blob/service/src/unitxt/metric_utils.py#L6). If i move the dependency, then this module becomes unusable, right?
> 
> If this module was only used within a function, i would have just made a local import. Do you know if there is any similar alternative for the above case?

Discussed an alternative solution suggested by @elronbandel :
We will not use pydantic at all. The `MetricRequest` and MetricResponse classes will inherit from unitxt `DataClass`. Then will use `to_dict()` to convert the request into json (with json.dumps too). In the other direction, will simply construct the `MetricResonse` from the json response using the default `DataClass` constructor.

requirements/base.rqr
matanor: Removed this requirement, applied the solution described [here](https://github.com/IBM/unitxt/pull/568#issuecomment-1951159347).

elronbandel (changes requested):
We have a mechanism to add and read new environment variables in 'settings_utils.py' I think it's better if you use it. It gives us more control over the settings and allow to review all the settings in one place. It also log new settings changes and have some more nice features.

elronbandel (approved):

PR #569, opened 2024-02-11
Fixing classnames and adding news_headline_json

ilyashnil:


elronbandel:
@ilyashnil Can you remove the commit of contract NLI as it is not part of this pr. Also make sure each dataset has both its prep python file as well as the resulting jsons

elronbandel (changes requested):
see my comment

Issue #325, opened 2023-11-22
Artifacts cache returns mutable items


matanor:
Consider the following scenario:

1. Get an artifact from the artifacts repo (e.g. an Accuracy metric).
2. Make an adjustment to it (e.g. disable confidence interval calculation). 
3. Retrieve the artifact again (e.g. to run another evaluation with the Accuracy metric). 
4. The artifact is then retrieved from the in-memory cache of the artifacts repo, the retrieved item will be the modified artifact object.

This may lead to unexpected results. For example, i was not getting any confidence intervals, because i had previously disabled that for the cached Accuracy metric while running on another stream.

Generally, current usage of the artifactory-repo cache can be dangerous. There is no explicit way to prevent changes to objects returned from the cache. Maybe the returned artifact should be a copy of what is present in the cache.

matanor:
@elronbandel @yoavkatz 

@lilacheden has just encountered this issue as well, changing a field on an artifact persists for artifacts with the same name that are retrieved later.. was wondering what is the motivation behind the artifacts cache? is it something that can be removed?

matanor:
discussing with @elronbandel :
- We need to deepcopy the instances returned from `fetch_artifact`().
- One option is below.
- Place `copying_lru_cache` in the `utils.py` module
- Optionally later add an option for avoiding the deepcopy, in case this is needed for better runtime performance.

```
from functools import lru_cache
from copy import deepcopy

def copying_lru_cache(maxsize=10, typed=False):
    def decorator(f):
        cached_func = lru_cache(maxsize=maxsize, typed=typed)(f)
        def wrapper(*args, **kwargs):
            return deepcopy(cached_func(*args, **kwargs))
        return wrapper
    return decorator

@copying_lru_cache()
def f(arg):
    print(f"Called with {arg}")
    x = [0, 1, arg]  # Stand-in for some long computation
    return x
```

PR #573, opened 2024-02-13
Add code needed for evaluating metrics as models

lilacheden:
1. Speed up correlation metrics by not running correlation on each instance
2. Return global scores and disable confidence intervals for evaluate() in eval_utils

### DIFF
--- a/prepare/metrics/kendalltau.py
+++ b/prepare/metrics/kendalltau.py
@@ -21,6 +21,31 @@ metric = MetricPipeline(
     metric=KendallTauMetric(),
 )
 
+
+singleton_instance_target = [
+    {
+        "kendalltau_b": np.nan,
+        "score": np.nan,
+        "score_name": "kendalltau_b",
+        "p_val": np.nan,
+    }
+]
+
+singleton_global_target = {
+    "score_name": "kendalltau_b",
+    "kendalltau_b": np.nan,
+    "score": np.nan,
+    "p_val": np.nan,
+}
+
+outputs = test_metric(
+    metric=metric,
+    predictions=["1"],
+    references=[["1"]],
+    instance_targets=singleton_instance_target,
+    global_target=singleton_global_target,
+)
+
 predictions = ["1.0", " 2.0", "1.0"]
 references = [["-1.0"], ["1.0"], ["0.0"]]
 
--- a/prepare/metrics/kendalltau.py
+++ b/prepare/metrics/kendalltau.py
@@ -26,8 +26,8 @@ references = [["-1.0"], ["1.0"], ["0.0"]]
 
 instance_targets = [
     {
-        "kendalltau_b": None,
-        "score": None,
+        "kendalltau_b": np.nan,
+        "score": np.nan,
         "score_name": "kendalltau_b",
     }
 ] * 3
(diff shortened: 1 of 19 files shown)


assaftibm:
nice additions

src/unitxt/eval_utils.py
```diff
@@ -1,29 +1,70 @@
...
+    global_scores = {}
     for metric_name in metric_names:
         multi_stream = MultiStream.from_iterables({"test": dataset}, copying=True)
         metrics_operator = SequentialOperator(steps=[metric_name])
+
+        if not compute_conf_intervals:
+            first_step = metrics_operator.steps[0]
+            if isinstance(first_step, MetricPipeline):
+                n_samples_before = first_step.metric.n_resamples
+                first_step.metric.n_resamples = None
```
matanor, line 34: perhaps call this `disable_confidence_interval_calculation` (see [here](https://github.com/IBM/unitxt/blob/main/src/unitxt/metrics.py#L77)) on the metric, its the same, but wraps the inner use of the `n_resamples` field.

src/unitxt/eval_utils.py
```diff
@@ -1,29 +1,70 @@
...
         instances = list(metrics_operator(multi_stream)["test"])
         for entry, instance in zip(dataset, instances):
             entry[metric_name] = instance["score"]["instance"]["score"]
-    return dataset
+
+        if len(instances) > 0:
+            global_scores[metric_name] = instances[0]["score"].get("global", {})
+
+        if not compute_conf_intervals:
+            first_step = metrics_operator.steps[0]
```
matanor, line 47: Please document this hack with a comment, and mention this issue https://github.com/IBM/unitxt/issues/325 in the comment. Also once this PR is merged, please add a reference from the issue to this code. With the motivation being that once the issue 325 is handled, we will remember to go back here and remove this hack.

prepare/metrics/kendalltau.py
```diff
@@ -21,6 +21,31 @@
     metric=KendallTauMetric(),
 )
 
+
+singleton_instance_target = [
```
matanor, line 25: do you think we need single instance test also for spearmanr?

matanor (changes requested):
hi @lilacheden left some comments in addition to what we talked (avoiding per-instance computation in global metrics) .. pls have a look ..

prepare/metrics/kendalltau.py
lilacheden: After the recent change that we discussed, we don't need it for either of them.

src/unitxt/metrics.py
```diff
@@ -989,9 +1006,7 @@ def compute(
         kendall_results = self.kendalltau(references, predictions, variant=self.variant)
         corr = kendall_results.correlation
         return {
-            "score_name": self.main_score,
```
matanor, line 992: just checking - why are these removed?

matanor (approved):
LGTM, left one questions, just to make sure something is as expected ..
Thanks!

src/unitxt/metrics.py
lilacheden: because they are set outside anyway if globalmetric:
```
def _compute(
        self,
        references: List[List[str]],
        predictions: List[str],
        additional_inputs: List[Any],
    ) -> dict:
        result = self.compute(references, predictions, additional_inputs)
        result["score"] = result[self.main_score]
        result["score_name"] = self.main_score
        return result
```

PR #574, opened 2024-02-14
Add plugin operators

elronbandel:
This is a proposal for a new behaviour that allows to change card operators from the final command such that:
`load_dataset("card=cards.wikitq,table_serializer=serializers.table.markdown")`
will be loading the wikitq with different table serializer, markdown serializer in this case.

### DIFF
--- a/prepare/cards/wikitq.py
+++ b/prepare/cards/wikitq.py
@@ -2,12 +2,14 @@ from src.unitxt.blocks import (
     AddFields,
     CopyFields,
     LoadHF,
+    RenameFields,
     SerializeTableAsIndexedRowMajor,
     TaskCard,
     TruncateTableCells,
     TruncateTableRows,
 )
 from src.unitxt.catalog import add_to_catalog
+from src.unitxt.operator import PlugInOperator
 from src.unitxt.test_utils.card import test_card
 
 card = TaskCard(
@@ -18,7 +20,13 @@ card = TaskCard(
         AddFields({"context_type": "table"}),
         TruncateTableCells(max_length=15, table="table", text_output="answer"),
         TruncateTableRows(field="table", rows_to_keep=50),
-        SerializeTableAsIndexedRowMajor(field_to_field=[["table", "context"]]),
+        PlugInOperator(
+            field="table_serializer",
+            default=SerializeTableAsIndexedRowMajor(
+                field_to_field=[["table", "table"]]
+            ),
+        ),
+        RenameFields(field_to_field={"table": "context"}),
     ],
     task="tasks.qa.with_context.extractive",
     templates="templates.qa.with_context.all",
--- a/src/unitxt/operator.py
+++ b/src/unitxt/operator.py
@@ -508,3 +508,40 @@ class SequentialOperatorInitilizer(SequentialOperator):
         for operator in self.steps[1 : self._get_max_steps()]:
             multi_stream = operator(multi_stream)
         return multi_stream
+
+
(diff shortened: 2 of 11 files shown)


yoavkatz:
If I understand there are two changes here.

1. The ability to add constant fields to the datasets via the recipe (what happens If it overwrites an existing field?)
2. An operator PlugIn operators that by defaults apply an operator to the stream, but allows changes the operator.

The Plugin operator is a bit tricky to understand (even the example is).

src/unitxt/operator.py
```diff
@@ -508,3 +508,40 @@ def process(self, *args, **kwargs) -> MultiStream:
...
+    """Applies an operator to each instance in a stream based on specific field.
+
+    Args:
+        field (str): name of the field that contains a single name of the operator to be applied
+        for the processing of the instance. Each operator is equipped with 'process_instance()'
+        method.
+
+        default_operator (List[str]): A list of default operators to be used if no operators are found in the instance.
+
+    Example:
```
yoavkatz, line 523: This is example is really not clear to me. Why is "processors.to_string" added to field "c"?

prepare/cards/wikitq.py
```diff
@@ -18,7 +20,13 @@
         AddFields({"context_type": "table"}),
         TruncateTableCells(max_length=15, table="table", text_output="answer"),
         TruncateTableRows(field="table", rows_to_keep=50),
-        SerializeTableAsIndexedRowMajor(field_to_field=[["table", "context"]]),
+        PlugInOperator(
+            field="table_serializer",
```
yoavkatz, line 24: I would call it "CustomizableOperator" because it's more directly describes what you can do with it.

Then I would change "field" to "overide_operator_field" so it will be more clear what this field means.

yoavkatz: Note that people can replace the default operator by any other operator, so it can dramatically change the semantic of the card.

tests/library/test_recipe.py
```diff
@@ -52,6 +52,46 @@ def test_standard_recipe(self):
...
+            ),
+            format=SystemFormat(
+                demo_format="User:{source}\nAgent:{target}\n\n",
+                model_input_format="{instruction}\n\n{demos}User:{source}\nAgent:",
+            ),
+            additional_field="test",
+        )
+        stream = recipe()
+
+        for instance in stream["train"]:
```
yoavkatz, line 71: What is tested in this example? We add a field but don't check or use it.

tests/library/test_recipe.py
```diff
@@ -52,6 +52,46 @@ def test_standard_recipe(self):
...
+            instruction=TextualInstruction(text="classify"),
+            template=InputOutputTemplate(
+                input_format="{premise}",
+                output_format="{label}",
+            ),
+            format=SystemFormat(
+                demo_format="User:{source}\nAgent:{target}\n\n",
+                model_input_format="{instruction}\n\n{demos}User:{source}\nAgent:",
+            ),
+            additional_field="test",
```
yoavkatz, line 67: Whta if people will start adding fields that will not be compatible with different changes.
Perhaps, we will require a prefix like 'custom_field_override_serializer".

dafnapension:
A naïve question.. : Since the card is a json file, a simpler, and perhaps better understood way (addressing @yoavkatz concern) would be to first load that json file into a json object, then change that json object, and use the edited version as the input to the recipe?
Or the card must be instantiated via fetch_artifact?  can that ad-hoc edited card be added to the catalog via add_card?

PR #577, opened 2024-02-16
load dataset via custom script & add tabfact task

csrajmohan:
* Added a data loader that can load a dataset via custom HF dataset loading script. This gives flexibility to write custom loading scripts and load into Unitxt pipeline when needed. The loading script is placed under catalog/loaders directory and referenced in the taskcard.
* Added tabfact(Table Fact checking) task card - This uses custom dataset loading script for loading the dataset via custom processing.


src/unitxt/loaders.py
```diff
@@ -394,3 +395,27 @@ def process(self):
             )
 
         return MultiStream.from_iterables(dataset)
+
+
+class LoadHFCustomDatasetScript(Loader):
+    """Loads a dataset using a custom HF dataset loading script from catalog/loaders path."""
```
elronbandel, line 401: This need a more extensive documentation maybe with a link to huggingface docs explaining how to create custom datasets as a python file.

src/unitxt/loaders.py
```diff
@@ -394,3 +395,27 @@ def process(self):
...
+        script_path = os.path.join(catalog_path, "loaders", self.file)
+
+        assert (
+            os.path.exists(script_path) is True
+        ), "Script file does not exist in the local catalog path"
+
+        loader = LoadHF(path=script_path, streaming=False)
+        ms = loader.process()
+        dataset = ms.to_dataset()
+        return MultiStream.from_iterables(dataset)
```
elronbandel, line 421: Why not?
```suggestion
        return  LoadHF(path=script_path, streaming=self.streaming)
```

elronbandel (changes requested):
Great work! see my comments. 
I think it will be better seperation to have the python file outside the catalog. 
in `src/unitxt/custom/datasets` and have also equivalent `src/unitxt/custom/metrics`
This would represent better their role and place in the repo. 
That way we can also save in the catalog a json entry for those datasets/metrics.
What do you think? Also do you think you could do such HFCustomMetric to fix the currently broken `src/metrics/* `?

src/unitxt/loaders.py
csrajmohan: added further documentation and a link

src/unitxt/loaders.py
csrajmohan: Tried this way of calling LoadHF directly first but it caused following error 
`Traceback (most recent call last):
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/prepare/cards/tabfact.py", line 49, in <module>
    test_card(card)
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/test_utils/card.py", line 260, in test_card
    test_with_eval(
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/test_utils/card.py", line 160, in test_with_eval
    examples = load_examples_from_standard_recipe(
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/test_utils/card.py", line 78, in load_examples_from_standard_recipe
    examples = print_recipe_output(
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/test_utils/card.py", line 134, in print_recipe_output
    message, examples = construct_recipe_output_message(
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/test_utils/card.py", line 104, in construct_recipe_output_message
    multi_stream = recipe()
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/operator.py", line 478, in __call__
    return super().__call__()
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/operator.py", line 157, in __call__
    result = self._process_multi_stream(multi_stream)
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/operator.py", line 168, in _process_multi_stream
    result = self.process(multi_stream)
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/operator.py", line 486, in process
    multi_stream = operator(multi_stream)
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/operator.py", line 157, in __call__
    result = self._process_multi_stream(multi_stream)
  File "/Users/rajmohan/Projects/EvalFramework/gitrepos/unitxt/src/unitxt/operator.py", line 202, in _process_multi_stream
    for stream_name, stream in multi_stream.items():
AttributeError: 'LoadHF' object has no attribute 'items'`

Then used the method followed in test_loaders.py as above which worked.

csrajmohan:
> Great work! see my comments. I think it will be better seperation to have the python file outside the catalog. in `src/unitxt/custom/datasets` and have also equivalent `src/unitxt/custom/metrics` This would represent better their role and place in the repo. That way we can also save in the catalog a json entry for those datasets/metrics. What do you think? Also do you think you could do such HFCustomMetric to fix the currently broken `src/metrics/* `?

Sure, Thanks Elron. Let me think about this and get back

csrajmohan:
Moved the location of custom dataset scripts to `src/unitxt/custom/datasets` and updated the loader method to reflect this. let me know if you would like me to make similar change for metrics too.

yoavkatz:
@csrajmohan - First, thanks for the great work integrating your use case into unitxt.  I think it's an important example of how user feedback and commitment improves the technology for everyone , creating a win-win ecosystem.

Regarding this particular issue - I wanted to discuss additional options.  In this case, you have several Huggingface dataset definition that you integrated with unitxt.    The solution proposed is to place these HF dataset definition inside unitxt.   What about to option to place them directly in HF hub (https://huggingface.co/ibm) ? They are open source in any case.

The reason we are less inclined to put this code in unitxt, is that this loader code is very specific for the dataset, but is in HF format and not unitxt format.

csrajmohan:
@yoavkatz I agree with your point. Even we felt the same that this dataset specific loading code should not be part of core Unitxt code. But was unclear where to keep it. 

Placing it in HF hub (https://huggingface.co/ibm) looks like best option for the public datasets. I was not aware of this option. I can go ahead and do it this way for this task.

For private datasets, do you have any suggestions to handle a similar scenario? I suppose a user can pass local script path and load dataset using the same method.
We probably don't need a separate loader called LoadHFCustomDatasetScript as LoadHF serves the purpose though it may be unclear for some from the name of the operator.

yoavkatz:
I'm glad the HF option is acceptable for the open source datasets.

For internal datasets, they are currently in FM-eval.  There, we are a bit more flexible to hold external code. However, the true prefer option is to move the post processing to the code to the dataset card, without the need for a HF definition. We do have options to run python code as a step in the processor.

csrajmohan:
That's good to know. Thanks. 
For this particular task, let me do the following.
I will add the dataset to IBM Org in HF hub and modify the task card to read from there. I will remove all the code related to custom script loading. Then will request for a fresh review.

Note: This particular dataset is already there in HF but we have customized the loading script to load dataset in a particular format by some processing. This is what we are uploading back to HF hub inside IBM Org.

@yoavkatz @elronbandel

csrajmohan:
As considerable amount of newly added code is to be removed from this PR, I decided to close this PR and create a fresh PR with just the required file(s) alone.

PR #579, opened 2024-02-18
cards for cohere for ai aya dataset

dafnapension:


gitMichal:
thanks @dafnapension , it looks good. Two minor issues:
1. could you please change the metric to normalized_sacrebleu
2. why the dolly_human_edited is commented out?

prepare/cards/cohere_for_ai.py
```diff
@@ -0,0 +1,74 @@
...
+            loader=LoadHF(path=dataset_name, name=subset, streaming=True),
+            preprocess_steps=[
+                SplitRandomMix(
+                    {"train": "test[90%]", "validation": "test[5%]", "test": "test[5%]"}
+                ),
+                FilterByCondition(values={"language": lang}, condition="eq"),
+                RenameFields(
+                    field_to_field={"inputs": "question", "targets": "answers"}
+                ),
+                ExecuteExpression(expression="[answers]", to_field="answers"),
```
elronbandel, line 42:
```suggestion
                ListFieldValues(fields=["answers"], to_field="answers"),
```

prepare/cards/cohere_for_ai.py
```diff
@@ -0,0 +1,74 @@
...
+from src.unitxt.logging_utils import get_logger
+from src.unitxt.operators import ExecuteExpression, FilterByCondition
+from src.unitxt.settings_utils import get_settings
+from src.unitxt.standard import StandardRecipeWithIndexes
+from src.unitxt.test_utils.card import test_card
+
+settings = get_settings()
+orig_settings = (settings.allow_unverified_code, settings.allow_unverified_code)
+settings.allow_unverified_code = True
+settings.global_loader_limit = 25000  # to ensure language is encountered
```
elronbandel, line 17: Why is the loader limit necessary?

elronbandel (dismissed):
Make sure to use the listing operator, ExecuteExpression should be kept for special cases.

prepare/cards/cohere_for_ai.py
dafnapension: First loading, then filtering by language. If the language that you filter by does not show among the first (that are loaded) your stream is exhausted, throwing exception.
In testing, the limit written in settings overrides the one passed to test_card.

prepare/cards/cohere_for_ai.py
dafnapension: Neat! Gladly accepted!

prepare/tasks/qa/tasks.py
```diff
@@ -25,7 +25,7 @@
 add_to_catalog(
     FormTask(
```
yoavkatz, line 26: We should change all QA tasks to support "answers" and not "answer.  Right now it will not be consistent.
add_to_catalog(
    FormTask(
        inputs=["context", "context_type", "question"],
        outputs=["answer"],
        metrics=["metrics.squad"],
    ),
    "tasks.qa.with_context.extractive",
    overwrite=True,
)

add_to_catalog(
    FormTask(
        inputs=["context", "context_type", "question"],
        outputs=["answer"],
        metrics=["metrics.rouge"],
        augmentable_inputs=["context", "question"],
    ),
    "tasks.qa.with_context.abstractive",
    overwrite=True,
)

gitMichal (approved):
reviewed with Elad

prepare/tasks/qa/tasks.py
dafnapension: Done!

yoavkatz: Note that if we change "tasks.qa.with_context.abstractive" to "answers", then we need to fix all the cards that us it.

dafnapension: yes, and change some templates too.

PR #581, opened 2024-02-19
Add ability to load from catalog with arguments overwrite

elronbandel:
This PR allows to specify artifact names with modifications. For example if you have `metrics.accuracy` in the catalog with a field `reduction="mean"` you can load it with a modification to median by: `metrics.accuracy[reduction=median]`

### DIFF
--- a/docs/catalog.py
+++ b/docs/catalog.py
@@ -7,10 +7,11 @@ from unitxt.utils import load_json
 
 
 def write_title(title, label):
-    underline_char = "-"
-    underline = underline_char * len(title)
+    title = f"📁 {title}/"
+    wrap_char = "="
+    wrap = wrap_char * len(title)
 
-    return f".. _{label}:\n\n----------\n\n{title}\n{underline}\n\n"
+    return f".. _{label}:\n\n{wrap}\n{title}\n{wrap}\n\n"
 
 
 def custom_walk(top):
@@ -88,7 +89,10 @@ def make_content(artifact, label, all_labels):
             references.append(f":ref:`{label_no_catalog} <{label}>`")
     if len(references) > 0:
         result += "\nReferences: " + ", ".join(references)
-    return result
+    return (
+        result
+        + "\n\n\n\n\nRead more about catalog usage :ref:`here <using_catalog>`.\n\n"
+    )
 
 
 class CatalogEntry:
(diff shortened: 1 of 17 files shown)


yoavkatz:
Elron can you add a description on what this allows? Can we change any field, or only literal strings?

Could we use it to override the default metrics in the task?

Also not the tests fails.

elronbandel:
@yoavkatz this pr will allow only single pr modifications we can also add list modifications which will be useful for both editing metrics as well as posprocessing steps etc... A possible syntax can be: `tasks.classification.multi_label[metrics=[accuracy,f1]]` what do you think?

yoavkatz:
> we can also add list modifications which will be useful for both editing metrics as well as posprocessing steps etc... A possible syntax can be:


I think the we need to see the syntax is similar to what would appear in the json.

In the json it is:

"metrics": [
        "metrics.accuracy",
        "metrics.f1_macro_multi_label"
    ],

So I think it should be:
tasks.classification.multi_label[metrics=["accuracy","f1"]]

elronbandel:
This syntax will not be allowed in the config names of hugginface datasets. The syntax i implemented is as i suggested:
`tasks.classification.multi_label[metrics=[metrics.accuracy, metrics.f1]]`

docs/catalog.py
```diff
@@ -254,5 +254,35 @@ def run(self):
...
+
+def create_catalog_docs():
     CatalogDocsBuilder().run()
+    current_dir = os.path.dirname(__file__)
+    catalog_rst_file = os.path.join(current_dir, "catalog.rst")
+    source_str = "Catalog\n-------\n"
+    target_str = (
+        "Catalog\n-------\n\n"
+        "The catalog is a place for the Unitxt community to share assets used for processing datasets.\n\n"
+        "You can load directly from the catalog with `fetch_artifact('artifact.id')`.\n"
```
yoavkatz, line 280: fetch_artifacts needs some import, right?
Also it's not clear what 'artifact.id' means.

docs/catalog.py
```diff
@@ -254,5 +254,35 @@ def run(self):
...
+        "You can also place anywhere in Unitxt catalog ids instead of realclasses\n"
```
yoavkatz, line 281: This is also unclear without more context.

docs/catalog.py
```diff
@@ -254,5 +254,35 @@ def run(self):
...
+        "Finally you can also state an ID with a modification to the asset parameters `asset.id[key=value]`.\n"
```
yoavkatz, line 282: Again, this requires more explanations.  I suggest that instead of placing this description in every page.  Have a link to page on loading and manipulating artifacts.

PR #582, opened 2024-02-19
Add tab_fact dataset, a dataset for classification of textual entailment from tables

csrajmohan:
* Added a new taskcard for Table Fact Checking using TabFact public dataset
* The dataset is loaded from ibm org HF hub(https://huggingface.co/datasets/ibm/tab_fact) where a custom dataset loading script is added to load dataset in appropriate format.

### DIFF
--- a/prepare/cards/tab_fact.py
+++ b/prepare/cards/tab_fact.py
@@ -1,13 +1,12 @@
 from src.unitxt.blocks import (
     AddFields,
-    FormTask,
     LoadHF,
     MapInstanceValues,
+    RenameFields,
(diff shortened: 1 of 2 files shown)


elronbandel (changes requested):
Could you adjust this PR to build on top this one:
https://github.com/IBM/unitxt/pull/586

More sepcificlly probably you will want to define:

- `type_of_relation` as either `entilment` or `factuality`
- `text_a_type` as `table`
- `text_b_type` as `statement`

Then just use the task and templates from there

csrajmohan:
Done

elronbandel (approved):

PR #584, opened 2024-02-20
filter on loading rather than increase loading limit

dafnapension:
per @elronbandel comment in [recent PR](https://github.com/IBM/unitxt/pull/579):  
Perhaps employing the HuggingFace filter upfront can ease load on the system, better than increase loading_limit only to allow the following pre-processing step that filters out.

### DIFF
--- a/prepare/cards/cohere_for_ai.py
+++ b/prepare/cards/cohere_for_ai.py
@@ -6,16 +6,9 @@ from src.unitxt.blocks import (
 )
 from src.unitxt.catalog import add_to_catalog
 from src.unitxt.logging_utils import get_logger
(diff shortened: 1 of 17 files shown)


dafnapension:
My understanding of the HF api, is that a function returning True or False, when fed with an insrance, is needed there. We need it to be artifact-able, so can be cataloged. Lambda seemed most apropriate.
Another, safer, solution can be laid on the shoulders of recipe: if it sees a filter by condition, or by expression, operator among the stream, the recipe should position any stream refiner by number - **following** the filter. perhaps have the refiners -- the last in the pipeline?

elronbandel (approved):

PR #585, opened 2024-02-20
Add judge metric

perlitz:
Adding a judge metric example


src/unitxt/metrics.py
```diff
@@ -1387,6 +1392,98 @@ def compute(
...
+        metric = LlamaIndexCorrectnessMetric()
+        metric.prepare()
+        scores = metric.compute(references, predictions, additional_inputs)
+    """
+
+    reduction_map = {"mean": ["score"]}
+    main_score = "score"
+    # model_name: str
+
+    _requirements_list: List[str] = []
```
elronbandel, line 1417:
```suggestion
    _requirements_list: List[str] = ["llama_index"]
```

elronbandel, line 1417:
```suggestion
    _requirements_list: List[str] = ["llama_index"]
```

src/unitxt/metrics.py
```diff
@@ -1387,6 +1392,98 @@ def compute(
...
+        main_score (str): The main score used for evaluation.
+        _requirements_list (List[str]): A list specifying any additional requirements for the metric.
+
+    Methods:
+        prepare(self): Initialization method for the metric.
+        compute(self, references, predictions, additional_inputs): Method to compute the metric.
+
+    Usage:
+        metric = LlamaIndexCorrectnessMetric()
+        metric.prepare()
```
elronbandel, line 1409:
```suggestion
```
happens automatically

prepare/metrics/llama_index_corectness.py
```diff
@@ -0,0 +1,57 @@
...
+#     metric=LlamaIndexCorrectnessMetric(),
+#     postpreprocess_steps=[
+#         # CopyFields(
+#         #     field_to_field=[
+#         #         ("score/instance/score", "score"),
+#         #     ],
+#         #     use_query=True,
+#         # )
+#     ],
+# )
```
elronbandel, line 24:
```suggestion
```
why this endded up not working? Just curious

src/unitxt/metrics.py
```diff
@@ -1265,6 +1266,10 @@ class TokenOverlap(InstanceMetric):
     def compute(
         self, references: List[Any], prediction: Any, additional_inputs: List[Dict]
     ) -> dict:
+        import ast
+
+        references = ast.literal_eval(references)
```
yoavkatz, line 1271: Why is this needed?

prepare/metrics/llama_index_corectness.py
```diff
@@ -0,0 +1,57 @@
+from unitxt import add_to_catalog
+from unitxt.metrics import LlamaIndexCorrectnessMetric
+from unitxt.test_utils.metrics import test_metric
```
yoavkatz, line 3: rename file from corectness-> correctness

src/unitxt/metrics.py
```diff
@@ -1387,6 +1392,98 @@ def compute(
         return self.pipe(inputs, batch_size=self.batch_size)
 
 
+class LlamaIndexCorrectnessMetric(BulkInstanceMetric):
+    """Custom metric class for evaluating correctness using Llama Index.
+
```
yoavkatz, line 1397: It requires OpenAI API .  How is it specified?

prepare/metrics/llama_index_corectness.py
dafnapension: `ast.literal_eval(references)`   where `references` is a list of two strings. but `literal_eval` expects `node_or_string`

src/unitxt/metrics.py
perlitz: I might be doing something wrong but references were definitely a string 
```python
"['reference1',...]"
```

prepare/metrics/llama_index_corectness.py
perlitz: sure

src/unitxt/metrics.py
perlitz: Using an API key in an environmental variable.

BTW, this is something we need to be careful about (or even consider having) as proprietary data can be leaked this way to OPENAI

Generally, this is a major issue with some packages like llama-index as they sometimes use OPENAI based models as default.

yoavkatz: I think the 

class LlamaIndexCorrectnessMetric(BulkInstanceMetric):

should br sn abstract class with method get_llm() .

Then there should be a class 

class LlamaIndexCorrectnessMetricByOpenAI(LlamaIndexCorrectnessMetric):

which would receive a parameter (model name, could be with default "gpt-3.5-turbo")
 
 and the get_llm method of it would be
 
  return OpenAI(mode_name)

yoavkatz: Then the name in the catalog would also be

correctness_by_lamma_index_openai

Then it will be clear which metric is used, and that it's with open.

yoavkatz: Also, I assume this should be the name of the returned score.

src/unitxt/metrics.py
yoavkatz: I think the issue is that RAG task you defined as a field answers (which is a list of possible answers), but you used InputOutputTemplate and not MultiReferenceTemplate.  So the list of answers was converted to a string representation.

@elronbandel - we need to better deal with this.

perlitz: even when I am using MultiReferenceTemplate I have the same problem
note that when I test the metric I am not facing it, only when I test it within the card testing

PR #587, opened 2024-02-20
Eval metrics

lilacheden:
1. Standard api for disabling/enabling confidence interval computation of metric/metricpipeline
2. make kendalltau and roc_auc metrics and not metric pipelines (to temporarily overcome the multiple metric pipeline issue)

### DIFF
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -62,6 +62,14 @@ class Metric(Artifact):
     def main_score(self):
         pass
 
+    @abstractmethod
+    def disable_confidence_interval_calculation(self):
+        pass
+
+    @abstractmethod
+    def set_n_resamples(self, n_resample):
+        pass
+
 
 class MetricWithConfidenceInterval(Metric):
     # The number of resamples used to estimate the confidence intervals of this metric.
@@ -78,9 +86,6 @@ class MetricWithConfidenceInterval(Metric):
         return np.random.default_rng(hash(get_seed()) & _max_32bit)
 
     def disable_confidence_interval_calculation(self):
-        self.n_resamples = None
-
-    def disable_confidence_interval_calculation_return_n_resamples(self):
         n = self.n_resamples
         self.n_resamples = None
         return n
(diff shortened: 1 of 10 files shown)


src/unitxt/metrics.py
```diff
@@ -80,6 +80,14 @@ def new_random_generator():
     def disable_confidence_interval_calculation(self):
         self.n_resamples = None
 
+    def disable_confidence_interval_calculation_return_n_resamples(self):
```
matanor, line 83: maybe just return `self.n_resamples` from the `disable_confidence_interval_calculation` method?
that shouldn't affect the existing code at all, no?

src/unitxt/metrics.py
```diff
@@ -997,15 +1020,20 @@ def prepare(self):
...
+            references = [reference[0] for reference in references]
+        references = [to_float_or_default(r) for r in references]
+        predictions = [to_float_or_default(p) for p in predictions]
+
         kendall_results = self.kendalltau(references, predictions, variant=self.variant)
         corr = kendall_results.correlation
         return {
             self.main_score: corr,
-            "p_val": kendall_results.pvalue,
+            "kendalltau_p_val": kendall_results.pvalue,
```
matanor, line 1036:
```suggestion
            f"{self.main_score}_p_val": kendall_results.pvalue,
```

src/unitxt/operators.py
```diff
@@ -1603,12 +1603,7 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
             ), f"Operator {metric_name} must be a Metric"
 
             if not self.calc_confidence_intervals:
-                if isinstance(metric, MetricWithConfidenceInterval):
-                    metric.disable_confidence_interval_calculation()
-                elif isinstance(metric, MetricPipeline) and isinstance(
-                    metric.metric, MetricWithConfidenceInterval
-                ):
-                    metric.metric.disable_confidence_interval_calculation()
+                metric.disable_confidence_interval_calculation()
```
matanor, line 1606: do we also need an abstract `disable_confidence_interval_calculation()` in `Metric`?

matanor: currently IIUC only the "overrides" were defined ..

matanor (approved):
LGTM! left some small comments ..

PR #589, opened 2024-02-20
structured data operators for serializing tablerows, triples and keyvalue pairs added

csrajmohan:
* Added new serialization operators for structured data: Table rows, Triples & KeyValue pairs
* Renamed table_operators -> struct_data_operators so that it best captures all structured data operators like the ones for Tables, Triples etc. and makes better sense to add other structured data operators to this in future.
* Added 3 new datatask cards to validate new operators
- wiki_bio : KeyValue Pair to Text(Data2Text task), Uses KeyValuePairs Serializer
- dart : Triples to Text(Data2Text task), Uses TriplesSerializer
- tablerow_classify : Table row classification using a Kaggle dataset(Heart disease prediction), Uses TableRowSerializer

### DIFF
--- a/prepare/cards/tablerow_classify.py
+++ b/prepare/cards/tablerow_classify.py
@@ -23,7 +23,7 @@ card = TaskCard(
         MapInstanceValues(mappers={"label": {"0": "Normal", "1": "Heart Disease"}}),
         AddFields(
             fields={
-                "text_type": "Person",
+                "text_type": "Person medical record",
                 "type_of_class": "Heart Disease Possibility",
             }
         ),
--- a/prepare/cards/tablerow_classify.py
+++ b/prepare/cards/tablerow_classify.py
@@ -1,6 +1,5 @@
 from src.unitxt.blocks import (
     AddFields,
-    FormTask,
     LoadFromKaggle,
     MapInstanceValues,
     RenameFields,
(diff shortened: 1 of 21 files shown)


prepare/cards/tablerow_classify.py
```diff
@@ -0,0 +1,54 @@
...
+    ),
+    preprocess_steps=[
+        SplitRandomMix(
+            {"train": "train[70%]", "validation": "train[10%]", "test": "train[20%]"}
+        ),
+        RenameFields(field_to_field={"HeartDisease": "label"}),
+        MapInstanceValues(mappers={"label": {"0": "Normal", "1": "Heart Disease"}}),
+        AddFields(
+            fields={
+                "text_type": "Person",
```
elronbandel, line 26:
```suggestion
                "text_type": "person medical record",
```
I think it is more descrptive, no?

prepare/tasks/generation.py
```diff
@@ -3,7 +3,7 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["input"],
+        inputs=["input", "type_of_input"],
```
elronbandel, line 6: Maybe would be also good to have type of output?
```suggestion
        inputs=["input", "type_of_input", "type_of_output"],
```

elronbandel (changes requested):
Overall this looks really good. I really appreciate that you use the existing tasks and even improve them!!  Go over my comment and ask for review when done and we can get it merged!

elronbandel:
In case its needed: the test is currently failing because of import from `unitxt` rather then `src.unitxt`

prepare/cards/tablerow_classify.py
csrajmohan: Done

prepare/tasks/generation.py
csrajmohan: Done

csrajmohan:
import for test file fixed

elronbandel (approved):
Great contribution!

PR #591, opened 2024-02-21
Docfix: Fix incorrect artifact names in Adding Dataset doc

yifanmai:
Change the task and template artifact names to the correct names.


elronbandel (approved):
Thanks!

PR #592, opened 2024-02-21
fixes to perplexity metric, updates to catalog

assaftibm:


matanor (approved):
hi @assaftibm LGTM, pls just see my comments pertaining to some names that i think could be improved, and one comment that i tried to clarify a bit ..

matanor:
hi @lilacheden , FYI, @assaftibm here has fixed an issue with the perplexity metric, and also has wrapped the `global_scores` you added as a results to `evaluate` with a dataframe.

PR #598, opened 2024-02-25
Allow dicts in overwrite args of fetched artifact

dafnapension:


elronbandel (changes requested):
Keep in mind hugginface forbid the use of this character in datasets query strings:
`<>:/\|?*`
Meaning you can not use the new syntax with `:` so if you can come up with syntax that can leave under this restriction it will be good. Also maybe we should add internally an assertion that forbid this characters in names and strings with a proper message that huggingface wont support it.

dafnapension:
Thanks, @elronbandel ,  you mean the ":" is forbidden even it does not go to inside any name, only between names?
Is it OK to use ";" instead?

elronbandel (approved):

PR #603, opened 2024-02-28
banking77 needs to be shuffled

ilyashnil:


prepare/cards/banking77.py
```diff
@@ -27,6 +30,7 @@
             {"train": "train[85%]", "validation": "train[15%]", "test": "test"}
         ),
         MapInstanceValues(mappers={"label": map_label_to_text}),
+        Shuffle(page_size=sys.maxsize),
```
elronbandel, line 33: shouldnt be before the splitting?

prepare/cards/news_category_classification_headline.py
```diff
@@ -1,3 +1,7 @@
+import sys
+
+from unitxt.operators import Shuffle
```
elronbandel, line 3:
```suggestion
from src.unitxt.operators import Shuffle
```

elronbandel (approved):

PR #604, opened 2024-02-28
Fixed formats and system prompts

elronbandel:


eladven (commented):
Hi Elron. Why the new formats have both system_prompt and instruction.

elronbandel:
@eladven We have both instruction (coming from the template serving as the task instruction) as well as the system_prompt serving as a task agnostic general instruction to the model

eladven (approved):

PR #612, opened 2024-02-29
Return copies of artifacts from the artifacts cache

matanor:
This prevents any accidental changes to returned artifacts from affecting the cached objects.

Closes https://github.com/IBM/unitxt/issues/325

### DIFF
--- a/src/unitxt/artifact.py
+++ b/src/unitxt/artifact.py
@@ -298,7 +298,19 @@ class UnitxtArtifactNotFoundError(Exception):
         return f"Artifact {self.name} does not exist, in artifactories:{self.artifactories}"
 
 
-@lru_cache(maxsize=None)
+def copying_lru_cache(maxsize=10, typed=False):
+    def decorator(f):
+        cached_func = lru_cache(maxsize=maxsize, typed=typed)(f)
+
+        def wrapper(*args, **kwargs):
+            return deepcopy(cached_func(*args, **kwargs))
+
+        return wrapper
+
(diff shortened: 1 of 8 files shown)


matanor:
Discussion with @elronbandel 
Move the cache to include the json loaded from the file, and not the artifact itself. This is because the artifact cannot always be deepcopied (some objects within cannot be 'pickle'd)

elronbandel (changes requested):
My only concern is that the artifact saving is reseting the cache for everyone else possible using the lru_cached_load_json. I think we should have a cache dedicated only for artifacts loading. (the fix is just to rename it to load_artifact_json_with_caching or similar

matanor:
> My only concern is that the artifact saving is reseting the cache for everyone else possible using the lru_cached_load_json. I think we should have a cache dedicated only for artifacts loading. (the fix is just to rename it to load_artifact_json_with_caching or similar

@elronbandel 
Renamed to:
```
@lru_cache(maxsize=None)
def artifacts_json_cache(artifact_path):
    return load_json(artifact_path)
```
so now the usage is:
```
    @classmethod
    def load(cls, path, artifact_identifier=None, overwrite_args=None):
        d = artifacts_json_cache(path)
        new_artifact = cls.from_dict(d, overwrite_args=overwrite_args)
        new_artifact.artifact_identifier = artifact_identifier
        return new_artifact
```
WDYT?

elronbandel (approved):

PR #613, opened 2024-02-29
suggest new features of Tempate for ffqa

dafnapension:
I was playing with the newly added ffqa cards.
I like the song  [these boots are made for walking](https://youtu.be/sbAM2HGGCVY), and was glad to see it among the first instances.. 

I suggest to shorten the text used for context. That text is so long, that printing it out generates "message too long" error on a linux console (I highlighted the error message):   
[ffqa_now_in_main.pdf](https://github.com/IBM/unitxt/files/14452506/ffqa_now_in_main.pdf)

I think that the  re.DOTALL flag used now misleads the regular expression engine, and it skips the \n\n looked for.

Last, and most importantly, the new features of Template allow to separate the instruction from the input, which comes handy in particular when using demos.
Here is how the printout looks when applying the above items, shortening context to only first 300 words. The latter modification is done ad-hoc, not inside the card. We may want to consider adding it into the card:   
[ffqa_suggested.pdf](https://github.com/IBM/unitxt/files/14452555/ffqa_suggested.pdf)

### DIFF
--- a/prepare/cards/ffqa_filtered.py
+++ b/prepare/cards/ffqa_filtered.py
@@ -94,12 +94,12 @@ def add_card(split: str):
                 condition="lt",
             ),
             ExecuteExpression(
-                expression='re.search(r"Document:\\s(.*)(\\n\\n|$)", inputs, re.DOTALL).group(1)',
+                expression='re.search(r"Document:\\s(.*)(\\n\\n|$)", inputs).group(1)',
                 imports_list=["re"],
                 to_field="context",
             ),
             ExecuteExpression(
-                expression='re.search(r"Question:\\s(.*)(\\n\\n|$)", inputs, re.DOTALL).group(1)',
+                expression='re.search(r"Question:\\s(.*)(\\n\\n|$)", inputs).group(1)',
                 imports_list=["re"],
                 to_field="question",
             ),
--- a/prepare/templates/qa/with_context.py
+++ b/prepare/templates/qa/with_context.py
@@ -40,11 +40,10 @@ add_to_catalog(
 # Template from https://huggingface.co/datasets/abacusai/WikiQA-Free_Form_QA
 add_to_catalog(
     MultiReferenceTemplate(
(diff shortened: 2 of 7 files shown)


elronbandel:
@marukaz can you review this?

marukaz:
@dafnapension I actually ran the previous code and your code and yes, there was a problem and it seems to have been fixed and improved. Thank you so much!

marukaz (approved):

marukaz:
@elronbandel I approved, but it seems that others need to approve it as well since I don't have write access.

elronbandel (approved):

dafnapension:
Dear @marukaz , thank you for your kind response, I am very glad you liked my suggestions.
Please note that one of the factors that yielded the change in the output (from [what it was](https://github.com/IBM/unitxt/files/14452506/ffqa_now_in_main.pdf to [how it is printed now](https://github.com/IBM/unitxt/files/14452555/ffqa_suggested.pdf)) is shortening each context to at most 300 (space separated) words.  This shortening is not part of the card yet. For the run of the printout, and just for the run, I simply appended it to the Card that is cataloged.   If you like that shortening, or want to make a variation thereof a formal part of the card, you may want to consider adding it [to the card](https://github.com/IBM/unitxt/blob/main/prepare/cards/ffqa_filtered.py).

marukaz:
@dafnapension Thank you for your suggestion! In my understanding, you are saying about this:

```py
ExecuteExpression(
  expression='" ".join(context.split()[:300])',
  to_field="context",
)
```

This is a great extension to shorten inputs.
However, my purpose in adding this dataset was to see how the model behaves in a longer context. Another problem is that if you simply truncate the text, the answers may disappear from the input. So I would like to keep the card as is.

dafnapension:
I see, @marukaz , and absolutely agree. Just wanted to make sure that I was sufficiently clear about what's in the updated card, and what I added ad-hoc just for the print-out.

PR #617, opened 2024-03-03
Draft new schema proposal

elronbandel:


elronbandel:
The idea is to have 

Tasks:

-

PR #618, opened 2024-03-03
Draft for now: try to avoid resampling where clearly possible

dafnapension:


dafnapension:
@matanor did a much cleaner job

PR #620, opened 2024-03-03
Touch up metrics

elronbandel:


matanor (approved):
LGTM

src/unitxt/metrics.py
```diff
@@ -111,21 +110,15 @@ def set_global_score(instances, global_score: Dict[str, Any]):
                 scores["global"] = {}
```
dafnapension, line 110: this whole if sentence seems redundant to me. In any case, the following assignment sentence sets the final value of `scores["global"]`

matanor: i agree, i think the `if` can be removed.. i can't think of why i added it..

PR #621, opened 2024-03-04
Eval metrics

lilacheden:
Add items required for the eval metrics pipeline:
1. InputOutputReferenceTemplate: the template receives two separate output fields for target and reference - allows to map between the two in the dataset level.
2.  DiverseLabelsSamplerFromClasses: diverselabelssampler to choose only from certain classes (to be able to show only definitive answer in the demos). Also filters the demo pool to make sure the demo is different from the instance.
3. Fix takeFirstWord post-processor
4. A workaround to fix the issue where the metric pipelines modifies the input stream 
5. fix CopyFields: copy a deep copy of the value

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -964,10 +964,7 @@ class CopyFields(FieldOperator):
     """
 
     def process_value(self, value: Any) -> Any:
-        try:
-            return copy.deepcopy(value)
-        except Exception:
-            return value
+        return copy.deepcopy(value)
 
 
 class GetItemByIndex(FieldOperator):
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1578,13 +1578,6 @@ class ApplyStreamOperatorsField(SingleStreamOperator, ArtifactFetcherMixin):
         yield from stream
 
 
-def get_merged_stream(stream_orig, stream_new):
-    for instance1, instance2 in zip(stream_orig, stream_new):
-        instance1["score"] = deepcopy(instance2["score"])
-
-    yield from stream_new
-
-
 class ApplyMetric(SingleStreamOperator, ArtifactFetcherMixin):
     """Applies metric operators to a stream based on a metric field specified in each instance.
 
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -33,6 +33,7 @@ General Operaotrs List:
 ------------------------
 """
 import collections
+import copy
 import operator
 import os
 import uuid
@@ -963,7 +964,10 @@ class CopyFields(FieldOperator):
     """
 
     def process_value(self, value: Any) -> Any:
-        return value
+        try:
+            return copy.deepcopy(value)
+        except Exception:
+            return value
 
 
 class GetItemByIndex(FieldOperator):
@@ -1574,6 +1578,13 @@ class ApplyStreamOperatorsField(SingleStreamOperator, ArtifactFetcherMixin):
         yield from stream
 
 
+def get_merged_stream(stream_orig, stream_new):
+    for instance1, instance2 in zip(stream_orig, stream_new):
+        instance1["score"] = deepcopy(instance2["score"])
+
+    yield from stream_new
+
+
 class ApplyMetric(SingleStreamOperator, ArtifactFetcherMixin):
     """Applies metric operators to a stream based on a metric field specified in each instance.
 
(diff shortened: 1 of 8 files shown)


lilacheden:
> @lilacheden Trying to understand the motivataion for the new reference template, why not using a postprocessor that uses the `classes` / `labels` field to restore the index of the label? (both for the predicitons and the references)

1. I want to have this mapping specific for the dataset and this would require a different post processor per mapping
2. More importantly - the mapping is not necessarily 1:1

For example, in one dataset the labels are 1, 0.5, 0 and I want the targets to be yes, no, maybe respectively.
On another one I have a label on scale 1-5 and I want to map the targets shown to the model as: 1-2: no, 3: maybe, 4-5: yes.
The label can also be a float, and so on - I need to access the actual field from the input dataset, it can't be deduced from the target.

elronbandel:
In this case, i think the right way to do it is to define a new task for ordinary classification (or verbal regression) where one of the task fields (either in the input or output) is this mapping i.e `{"maybe": 0.5, "yes": 1, "no": 0}`. Then you can have a postprocessor that uses this field to recover the original data. 

The reason for all of this hustle, it that we want to have the refrences, targets and model predictions of the same "dimension" or "state" so pepole can see the references and understand what is expected by the model to do. This asumption is also used in practice for automatic testing of cards and postprocessors. Lastly, if the mapping is not and explicit variable of the task how could we use it in post-processing to translate the predictions of the model back to the numeric space?

lilacheden:
@elronbandel - several issues regarding the sampler:

1. I see I should use the inputs[self.label] field instead of target - will fix (if we keep this class)

2. can I get the same result by using diverselabelssampler and setting the "choices" field only to the required classes?

3. Regarding the filtering for the example ("filter_source_by_instance"): I have a problem with the current implementation of the demos pool since I need the demos taken from the test but without removing them from the dataset. As a workaround I use the same data file for train and test, have demos taken from train, and add this filtering to make sure no example is given as a demo for itself. any more straight forward way to do this? Isn't it better to keep the entire test set in the first place and add this filtering when the demos are taken from the test?

elronbandel:
> can I get the same result by using diverselabelssampler and setting the "choices" field only to the required classes?
yes! 
>  As a workaround I use the same data file for train and test, have demos taken from train, and add this filtering to make sure no example is given as a demo for itself. any more straight forward way to do this? Isn't it better to keep the entire test set in the first place and add this filtering when the demos are taken from the test?
I afraid there isnt a better way, but maybe we should make this behaviour a default? as in, whenever we are sampling demos we will always make sure that its from the pool without the example in hand?

lilacheden:
@elronbandel 
> 
> I afraid there isnt a better way, but maybe we should make this behaviour a default? as in, whenever we are sampling demos we will always make sure that its from the pool without the example in hand?

Ok, will add it in sampler. Can we also change the demos_pool sampling so that the demos are not removed from the split, but copied from it?

lilacheden:
> @lilacheden In this case, i think the right way to do it is to define a new task for ordinary classification (or verbal regression) where one of the task fields (either in the input or output) is this mapping i.e `{"maybe": 0.5, "yes": 1, "no": 0}`. Then you can have a postprocessor that uses this field to recover the original data.
> 
> The reason for all of this hustle, it that we want to have the refrences, targets and model predictions of the same "dimension" or "state" so pepole can see the references and understand what is expected by the model to do. This asumption is also used in practice for automatic testing of cards and postprocessors. Lastly, if the mapping is not an explicit variable of the task how could we use it in post-processing to translate the predictions of the model back to the numeric space?

AIUC, This will work for my first example but not the second, where the mapping is from reference to target. In some cases we need this mapping even if there's no way to convert back from target to reference:

1. Since models are better with coarse grained classificatio, we want to simplify the question and the model will only return certain values (e.g. yes/no postprocessed to 1/0). For consistency with previous work, we still need to correlate/compare this with the original labels, which are fine grained or even continues. 
2.  In some cases we ask the model for yes or no response and want to use the logprobs to generate continues values for each label - which we want to compare to the original continues labels. so the targets are yes/no, but the references aren't.

elronbandel (approved):

lilacheden:
@elronbandel 
You told me before that by setting the choices in the DiverseLabelsSampler to the required labels I can assure that only demos with these labels are chosen, but that's not the case, since if there are examples when non of the choices are in the labels, they are considered as label [], and the sampling balances on this label as well.

e.g. if we set choices = ["dog","cat"] and have examples with the labels ["cow"], the sampler will balance on demos with dog, demos with cat and demos with none of them.

Is that a bug (should we remove the option for empty label) or is that the expected behavior and I just not understand you before?

Issue #606, opened 2024-02-28
Redundant warning message in confidence intervals


elronbandel:
This happen in catalog preparation tests and really feeling up the logs and make it hard to see other problems
@matanor 

```
[Unitxt|INFO|artifact.py:332] 2024-02-28 15:48:01,499 >> Artifact tasks.classification.multi_class is fetched from EnvironmentLocalCatalog(type='environment_local_catalog', artifact_identifier=None, is_local=True, name='local', location='/home/runner/work/unitxt/unitxt/src/unitxt/catalog')
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/scipy/stats/_resampling.py:144: RuntimeWarning: invalid value encountered in scalar divide
  a_hat = 1/6 * sum(nums) / sum(dens)**(3/2)
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/scipy/stats/_resampling.py:97: DegenerateDataWarning: The BCa confidence interval cannot be calculated. This problem is known to occur when the distribution is degenerate or the statistic is np.min.
  warnings.warn(DegenerateDataWarning(msg))
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/scipy/stats/_resampling.py:144: RuntimeWarning: invalid value encountered in scalar divide
  a_hat = 1/6 * sum(nums) / sum(dens)**(3/2)
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/scipy/stats/_resampling.py:97: DegenerateDataWarning: The BCa confidence interval cannot be calculated. This problem is known to occur when the distribution is degenerate or the statistic is np.min.
  warnings.warn(DegenerateDataWarning(msg))
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/scipy/stats/_resampling.py:144: RuntimeWarning: invalid value encountered in scalar divide
  a_hat = 1/6 * sum(nums) / sum(dens)**(3/2)
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/scipy/stats/_resampling.py:97: DegenerateDataWarning: The BCa confidence interval cannot be calculated. This problem is known to occur when the distribution is degenerate or the statistic is np.min.
  warnings.warn(DegenerateDataWarning(msg))
[...]

Issue #482, opened 2024-01-16
Warning in running confidence intervals


yoavkatz:
In some test run of cards the confidence internal returns a warning.  This confuses the user. 
I believe we should not calculate confidence interval on degenerate (all predictions are equal).

Yoav

This is version that reflects watson version of CFPB. I got the following warning, hope that this is ok:
/u/ilyashn/.conda/envs/bzsm2.2/lib/python3.9/site-packages/scipy/stats/_resampling.py:100: DegenerateDataWarning: The BCa confidence interval cannot be calculated. This problem is known to occur when the distribution is degenerate or the statistic is np.min.
warnings.warn(DegenerateDataWarning(msg))
/u/ilyashn/.conda/envs/bzsm2.2/lib/python3.9/site-packages/scipy/stats/_resampling.py:147: RuntimeWarning: invalid value encountered in scalar divide
a_hat = 1/6 * sum(nums) / sum(dens)(3/2)
/u/ilyashn/.conda/envs/bzsm2.2/lib/python3.9/site-packages/scipy/stats/_resampling.py:100: DegenerateDataWarning: The BCa confidence interval cannot be calculated. This problem is known to occur when the distribution is degenerate or the statistic is np.min.
warnings.warn(DegenerateDataWarning(msg))
/u/ilyashn/.conda/envs/bzsm2.2/lib/python3.9/site-packages/scipy/stats/_resampling.py:147: RuntimeWarning: invalid value encountered in scalar divide
a_hat = 1/6 * sum(nums) / sum(dens)(3/2)
/u/ilyashn/.conda/envs/bzsm2.2/lib/python3.9/site-packages/scipy/stats/_resampling.py:100: DegenerateDataWarning: The BCa confidence interval cannot be calculated. This problem is known to occur when the distribution is degenerate or the statistic is np.min.
warnings.warn(DegenerateDataWarning(msg))

PR #632, opened 2024-03-05
Avoid RuntimeWarning in confidence interval computation

matanor:
Closes https://github.com/IBM/unitxt/issues/606

- For score based confidence interval computation, skip the computation if all instance-based scores have the same value (as suggested by @yoavkatz [here](https://github.com/IBM/unitxt/issues/482)).
- For global metric based confidence internal computation, filter out `RuntimeWarning` during the computation. This is for the case that the same metric value is computed over all resamplings. Since the metric computation per resampling is done in a callback used by scify's `bootsrtap` method, the computation cannot be simply aborted.

### DIFF
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -199,7 +199,7 @@ class MetricWithConfidenceInterval(Metric):
         for score_name in score_names:
             # If all computed instance level scores are the same, there is no point in computing
             # confidence intervals. So skip to the next score.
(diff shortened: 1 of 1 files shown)


src/unitxt/metrics.py
```diff
@@ -49,8 +49,6 @@ def abstract_field():
 
 
 def nan_mean(x):
-    import warnings
```
matanor, line 52: remove since already imported at module level

PR #639, opened 2024-03-07
Group shuffle

sam-data-guy-iam:
Add a sub-class of Shuffle operator that, rather than shuffling all the instances individually, first groups instances within a page by desired grouping features, then shuffles the groups as a block.  This is necessary for our robustness metric work, where we want a test dataset to contain complete-blocks of question paraphrases (the groups).

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1751,7 +1751,7 @@ class FeatureGroupedShuffle(Shuffle):
 
         # now flatten the list so it consists of individual dicts, but in (randomized) block order
         return list(itertools.chain(*page_blocks))
-
+        
 
 class EncodeLabels(StreamInstanceOperator):
     """Encode each value encountered in any field in 'fields' into the integers 0,1,...
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1730,6 +1730,9 @@ class FeatureGroupedShuffle(Shuffle):
             See outputs_2features_bigpage and outputs_2features_smallpage in test_grouped_shuffle.
     """
 
+    grouping_features: List[str] = None
+    shuffle_within_group: bool = False
+
     def process(self, page: List[Dict], stream_name: Optional[str] = None) -> Generator:
         if self.grouping_features is None:
             super().process(page, stream_name)
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -32,8 +32,10 @@ The rest of this section is dedicated for general operators.
 General Operaotrs List:
 ------------------------
 """
+import collections
 import copy
 import operator
+import os
 import uuid
 import zipfile
 from abc import abstractmethod
@@ -1749,7 +1751,7 @@ class FeatureGroupedShuffle(Shuffle):
 
         # now flatten the list so it consists of individual dicts, but in (randomized) block order
         return list(itertools.chain(*page_blocks))
-        
+
 
 class EncodeLabels(StreamInstanceOperator):
     """Encode each value encountered in any field in 'fields' into the integers 0,1,...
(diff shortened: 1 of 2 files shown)


dafnapension (approved):

dafnapension:
Hi Sam, I managed to bring the PR to (almost) all green. Your signature is still a mystery for me: what does this DCO expect from you? :-)  At any rate, I overruled it pushing the button: "set DCO to pass".  I suggest that in future commits, you use the -s flag as well. Namely: you commit by:
`git commit -s -m "commit message for this commit"`

You can now merge this PR into main anytime you like, by pushing `squash and merge`

But, as the hood is still open, please consider the following two (potential) issues:

Is there an issue with your code grouping together also instances that were separated although their signature is the same. Namely: יאסוף זרויים
you collect into one list in `groups_to_instances` also instances that perhaps the user separated them intentionally from one another, by pushing in-between them an instance with a different `tuple(item[ff] for ff in self.grouping_features)`

Is there an issue with this?

And second question:
Is it important to gather **all** the instances of same signature  (signature meaning  `tuple(item[ff] for ff in self.grouping_features)` )  ?
If so, and if there are not too many signatures, perhaps better first find out all the different signatures via operator `Unique` and then do a loop of `SplitByValue` ?

sam-data-guy-iam:
> > Hi @sam-data-guy-iam , I managed to bring the PR to (almost) all green. Your signature is still a mystery for me: what does this DCO expect from you? :-) At any rate, I overruled it pushing the button: "set DCO to pass". I suggest that in future commits, you use the -s flag as well. Namely: you commit by: `git commit -s -m "commit message for this commit"`
> > You can now merge this PR into main anytime you like, by pushing `squash and merge`
> > But, as the hood is still open, please consider the following two (potential) issues:
> > Is there an issue with your code grouping together also instances that were separated although their signature is the same. Namely: יאסוף זרויים you collect into one list in `groups_to_instances` also instances that perhaps the user separated them intentionally from one another, by pushing in-between them an instance with a different `tuple(item[ff] for ff in self.grouping_features)`
> > Is there an issue with this?
> > And second question: Is it important to gather **all** the instances of same signature (signature meaning `tuple(item[ff] for ff in self.grouping_features)` ) , also across pages? If so, and if there are not too many signatures, perhaps better first find out all the different signatures via operator `Unique` and shuffle these unique-s, and finally filter-by-condition. Something in analogy with `SplitByValue` ?
> 
> @dafnapension, thanks for pointing this out: The assumption is that by specifying self.grouping_features, the user is saying that the unique combinations of values of these features define a group. Thus, it would not make sense for the user to, say, separate items that should belong to the same group, by inserting other rows between them (this would negate the fact that self.grouping_features defines the groups). If the user really wanted to do this, they would have used another feature to define the groups, which would account for this separation. However, shuffle_by_grouping_features groups items by unique combinations of self.grouping_features **within the page**. Items that should be in the same group could appear on different pages if either 1) the dataset was not sorted by these features, or 2) the dataset was sorted but a block of items in a group is split by a page break. The first case corresponds to what you said, and is also one of the reasons we wanted to include this shuffle option, to allow the analysis to be done properly. This dependence on page_size is included to allow it to function within the superclass (and thus it should not automatically look for all unique combinations across pages). I added a note in the documentation that if the user wants this behavior, they should set the page_size to be very big. I'll commit my changes, please take a look.

By the way, is יאסוף זרויים meant as a reference to אשאל אלהי?

dafnapension:
Of course!  שלום שבאזי.  Envisioning the collection of same-signature-but-spread apart..
I knew you would recognize, and by your user-name that you selected, I understood that you are fond of classics..

sam-data-guy-iam:
> Of course! שלום שבזי. Envisioning the collection of same-signature-but-spread apart, that you collect together to same list.. I knew you would recognize, and by your user-name that you selected, I understood that you are fond of classics..

What do you mean?  That is so funny yes I am half temani (nobody believes me because I am Ashkenazi) and a big fan of the music.

PR #652, opened 2024-03-12
fix summarization template

gitMichal:
fix the summarization template to not take only the first line.


elronbandel (commented):
Arent they suppose to be one sentence?

elronbandel (approved):

PR #654, opened 2024-03-13
Add metrics for binary tasks with float predictions

lilacheden:
1. Calculate accuracy, f1, precision and recall using a threshold
2. Calculate max F1 and max accuracy


elronbandel (approved):
Can you add one sentence docstring  about the BinaryMax metric

PR #658, opened 2024-03-13
Standard metrics

dafnapension:


yoavkatz:
@dafnapension @elronbandel - Can you explain the motivation for this PR? What are standard metrics and how do they relate to the existing metrics?

yoavkatz:
@elronbandel @dafnapension - I'm sure you discussed it between you alot, but I want to provide a different perspective.

I think stream in unitxt may be useful if unitxt  used for large scale training - however, it also has significant cost in terms of code and API complexity.   In evaluation , where typically only hundres of samples are tested, streaming will have no significant value.

We need metrics API that are 
(1) independent from each other
(2) easy for users to add  AND debug them.

Therefore, I think it's worth to have a discussion if this direction will have a net gain in terms of unitxt acceptance.

(@eladven - will be glad your input as well).

dafnapension:
Leave for now, if at all, continue via https://github.com/IBM/unitxt/pull/845

PR #659, opened 2024-03-13
20newsgroup from sklearn

ilyashnil:
It become apparent that there are at least 2 versions of 20_newsgroups. One with metadata (email heads like From) and one without. HuggingFace is without, sklearn does contain headers. Can we integrate it?


elronbandel (changes requested):
The LoadFromSklearn is a really nice contribution! 
Regarding naming:
cards.20newsgroup_sklearn -> cards.20newsgroup.sklearn (so pepole can browse the catalog and find all 20newsgroup variations under one directory)

ilyashnil:
renamed cads.20newsgroups_sklearn to cards.20_newsgroups.sklearn

PR #662, opened 2024-03-13
Add rerank recall metric to unitxt

### LINKED ISSUE #661: Add recall at rank metric
This metric computes recall at k globally for query,passage pairs.  This is to support perplexity reranking in fm-eval.

jlqibm:
Closes #661 in support of adding perplexity reranking to fm-eval.


jlqibm:
@yoavkatz please review

src/unitxt/metrics.py
```diff
@@ -3181,3 +3182,90 @@ def compute(
...
+    query_id_field selects the field containing the query id for an instance.
+    passage_id_field selects the field containing the passage id for an instance.
+    at_k selects the value of k used to compute recall.
+
+    """
+    main_score = "recall_at_5"
+    query_id_field: str = "query_id"
+    passage_id_field: str = "passage_id"
+    at_k: List[int] = [1, 2, 5]
+
```
elronbandel, line 3235:
```suggestion
     _requirements_list: List[str] = ["pytrec_eval"]
```
Unitxt by default does not operate with requirements but each operator/metric decalre its requirements

requirements/base.rqr
```diff
@@ -4,4 +4,5 @@ mecab-python3
 absl-py
 dpath
 ipadic
-scipy
\ No newline at end of file
+scipy
+pytrec_eval
```
elronbandel, line 8:
```suggestion
```
This should be moved to `requirements/tests.rqr`
See the other comment for full explanation

elronbandel (changes requested):
Fix the requirements according to unitxt no-requirements policy explained in the comments

src/unitxt/metrics.py
jlqibm: done

requirements/base.rqr
jlqibm: done

elronbandel:
You need to add the pytrec_eval to the`requirements/tests.rqr`

jlqibm:
@elronbandel pytrec-eval is added.

yoavkatz:
The tests failed with this:

image

The pre-commit failed with this:

image

(Do you have pre-commit installed with `pre-commit install` ? )

jlqibm:
@yoavkatz The `pandas` package is missing.  So there are 2 question here.  One, why didn't this fail at the import statement?  Two, which requirements file does pandas need to be added to.
Sorry about the test failure.

src/unitxt/metrics.py
```diff
@@ -12,6 +12,7 @@
 import evaluate
 import numpy
 import numpy as np
+import pytrec_eval
```
yoavkatz, line 15: This import should move inside the  'pytrec_eval_at_k' function so it will not be required by people not using the metric.

src/unitxt/metrics.py
```diff
@@ -3205,3 +3206,92 @@ def compute(
...
+    for k in at_k:
+        metric[f"{metric_name}@{k}"] = 0.0
+
+    metric_string = f"{metric_name}." + ",".join([str(k) for k in at_k])
+    # print('metric_string = ', metric_string)
+    evaluator = pytrec_eval.RelevanceEvaluator(
+        qrels, {"ndcg", metric_string}
+    )  # {map_string, ndcg_string, recall_string, precision_string})
+    scores = evaluator.evaluate(results)
+    scores = pd.DataFrame(scores).transpose()
```
yoavkatz, line 3226: Also need to

import pandas as pd 

in this function.

yoavkatz:
> @yoavkatz The `pandas` package is missing. So there are 2 question here. One, why didn't this fail at the import statement? Two, which requirements file does pandas need to be added to. Sorry about the test failure.


There is a missing import to pandas.   In general, if metrics needs specific packages like pandas or trec_eval, we import inside the relevant metric code, and not at the top of the metrics.py file, so If people don't need the metric, they will not need the include.

src/unitxt/metrics.py
jlqibm: Fixed.

jlqibm:
@elronbandel I already fixed the requested change but github still thinks it's pending.

jlqibm:
@yoavkatz can you tell what's holding up this merge?  There appears to be an outstanding change request but I already fixed that a while ago.

yoavkatz:
I reviewed the code, and it looks fine so I approved.

@elronbandel  - I think we the main issue is that the coverage is low - although tests were added (but in the prepare file).
The reason is that coverage is not measured on the prepare files.

We should probably give guidance to write the test in test_metrics and not in the prepare file (unless using a wrapper like HuggingFaceMetric).

PR #667, opened 2024-03-14
Add generic mechanism to check prediction and reference types in metrics

yoavkatz:
Up to now each metric checked whether the predictions and reference are of the right type independently (most did not have any checks).

There is now generic code in Metric to check the type of predictions/references and whether the metric can support a single reference per predictions or more.

These are the new parameter for each metric:

   ```
    # Override 'prediction_type' with the expected type of predictions
    # and references.  Example: "List[str]", "List[Dict]"", "string".
    # If left with default "None", a warning will be displayed.
    # In future version of unitxt, this will be an error.
    prediction_type: str = None

    # Standard metrics can receive multiple references per predictions (in a list)
    # Some metrics support only a single reference per prediction (one element in the list)
    single_reference_per_prediction: bool = False
```

*Non backward compatible change:*

unitxt assumes that there are multiple references per predictions. For example, in HF dataset definition (string for predictions, and list of string of references). It then post processes the prediction and each of string references using the post processors.

In in some metrics (retrieval metrics) the predictions are a list of documents id, and the reference are also a list of document ids . This means that in few shot learning, only one document will be presented in the example output.
It was fixed so that metics will receive list of lists.

### CODE REVIEW

**Comment by elronbandel** [comment]
A general comment, typing validation is something we could use in many other operators other then the metrics, which might suggest a more general mechanism. It could be activated by default for the metrics but could be applied for other operators considering some of them also de-facto perform type checking. 

Addressing this i would suggest to make a general mechanism for validation for operators. Possibly with following generic fields:

```python
required_fields: Dict[str, Any]
optional_fields: Dict[str, Any]
output_fields:  Dict[str, Any]
```

then for example the metric will have:

```python
required_fields = {"prediction": "str", "references": "List[str]"}
optional_fields: = {"task_data": {"choices": "List[str]"}}
output_fields: = {"score": "float"}
```

This mechanism will be then used for validation of all operators, if an operators declared outputs, then we can check its compliance with the next operator at compilation time. 

A side note: if we have references list of size one, then its type should be changed to the type of its signle list item, That way we can keep using the same mechanism of such validation

yoavkatz:
I agree it should be extended to other operators. 

The question is whether a generic solution can work.

Note that predictions field in metrics is really a List[str] and references is a List[List[str]].
This may be confusing compared to just specifying for metrics the prediction_type of 'str'.

Also if we consider Task , we need to consider naming (E.g. "output_fields:" vs "output")

I suggest the following:

  FormTask(
        inputs=["text", "text_type", "class"],
        outputs=["class", "label"],
        **field_types={"text":"str", "text_type":"str", "class": "str", "label":"List[str]"} **

        metrics=[
            "metrics.accuracy",
        ],
        augmentable_inputs=["text"],


Also:
 
1. Where do you see the need for optional params? 
2. Do we need to check the output type?

elronbandel:
> @yoavkatz: Note that predictions field in metrics is really a List[str] and references is a List[List[str]].
This may be confusing compared to just specifying for metrics the prediction_type of 'str'.

I think the definitions should be at the instance level meaning: `{"prediction": "str", "references": "List[str]"}`

> @yoavkatz : I suggest the following:

I like this suggestion in general, maybe it will be more concise like that:
```python
  FormTask(
        inputs={"text":"str", "text_type":"str", "class": "str"},
        outputs={"class": "str", "label":"List[str]"} ,
        metrics=[
            "metrics.accuracy",
        ],
        augmentable_inputs=["text"],
```
For backward competability and ease of use we can allow pepole to use either list or dict depending on if they want to define types.

> @yoavkatz :  Where do you see the need for optional params?

In operators and metrics that can use things from the stream, we want to validate that their type is correct but if they are missing its ok.

> @yoavkatz :  Do we need to check the output type?

It can help us. If we check them even just in testing times, we can look at a pipeline and without having any instances check staticlly every two adjacent operators are compatible by comparing their outputs and inputs.

yoavkatz:
I suggest accepting this PR and in a new PR handle type checking in FormTask.  @elronbandel @eladven

src/unitxt/metrics.py
```diff
@@ -74,6 +74,98 @@ class Metric(Artifact):
...
+    allow_none_prediction: bool = False
+
+    # Standard metrics can receive multiple references per predictions (in a list)
+    # Some metrics support only a single reference per prediction (one element in the list)
+    single_reference_per_prediction: bool = False
+
+    # Some metrics (e.g. HF metrics) do not follow the standard unitxt metric API
+    # where references are passed as list (e.g. in spearmanr references is a scalar).
+    # In this case, and other special cases, we can disable type validation.
+    disable_type_validation: bool = False
```
elronbandel, line 91:
```suggestion
    prediction_type: string
```
The prediction type can detect it all. 
Allow None prediction can be done with:
`prediction_type = Optional[str]`
Single reference should be:
disable_type_validation can be done with:
`prediction_type = Any`
The only one that cannot be done directly is single_reference_per_prediction wchich in my opinion should not be part of this pr.

src/unitxt/metrics.py
```diff
@@ -74,6 +74,98 @@ class Metric(Artifact):
...
+            "int",
+            "float",
+            "List[str]",
+            "List[float]",
+            "Dict",
+            "Dict[str,str]",
+            "List[Tuple[str,str]]",
+        ]:
+            if self.prediction_type == type:
+                return eval(type)
```
elronbandel, line 158: See https://github.com/IBM/unitxt/pull/688 for a more complete solution

yoavkatz: Looks good. I will use it.  I merged to this PR to test it.  Once the other PR is merged, the changes will disappear from this PR.

src/unitxt/metrics.py
yoavkatz: > The prediction type can detect it all. Allow None prediction can be done with: `prediction_type = Optional[str]` Single reference should be: disable_type_validation can be done with: `prediction_type = Any` The only one that cannot be done directly is single_reference_per_prediction wchich in my opinion should not be part of this pr.

Thanks.  I changed the ' prediction type ' to check for None with Optional as you suggested.  Also removed disable_type_validation .   

However 'single_reference_per_prediction' is an important one and exposed some inconsistency in our use of metrics.    unitxt assumes that there are multiple references per predictions.  For example,  in  HF dataset definition (string for predictions, and list of string of references). It then post processes the prediction and each of string references using the post processors.     

In n some metrics (retrieval metrics) the predictions are a list of documents id, and the reference are also a list of document ids .  This means that in few shot learning, only one document will be presented in the example output.  

In multi label , on the other hand, and the predictions are a list of labels and references are a list of list of labels. 
 
Bottom line - all metrics should receive a list of references.  Metrics that can only handle one reference, should set the  'single_reference_per_prediction'  as prediction.
]

elronbandel (approved):

PR #669, opened 2024-03-15
patches to reuters, unfair_tos, and babi  cards

dafnapension:
setting  `settings.test_card_disable=None` at the top of `test_preparation`, we get exceptions.
This PR suggests some modifications, and hopefully useful additions, to overcome this problem.

The motivation was trying to chase the way Accuracy for multi-label classification is currently computed. It does not seem to be computed. No where do true negatives seem to be figured out. But Multilabel Accuracy is a different issue, beyond the scope of this PR.


dafnapension:
covered by https://github.com/IBM/unitxt/pull/674

PR #671, opened 2024-03-16
Pettiness re matthews ci 

dafnapension:
`prepare/metrics/matthews_correlation.py` contains a test case of three simple instances.
the mcc (Mathews Correlation Coefficient) of this 3 pair-item long sequence is: 0.5, as rightfully expected in the `global_target` in `prepare/metrics/matthews_correlation.py`.
The targeted global ci, however, is rather disappointing: for 3-item long sequence, there are 27 different resamples with replacement. All 27 are equi-probable, by definition of selection with replacement.
6 out of these 27 resamples have mcc score == 1.0. 
ci_high reports the 97.5 percentile. It is expected, then, to hit into the 1.0 scores. Seeing targeted `"matthews_correlation_ci_high": 0.5`, is thus disappointing.
The cause here is the small number, 3, of resamples done.

So I suggest to enlarge the number of resamples done to 30 (the sequences are so short, that no big deal), or write an explaining comment in `prepare/metrics/matthews_correlation.py` .

Once we set `settings.test_metric_disable = None`,  in `test_preparation`, we see a few more examples, for which the new values of ci_low and ci_high seem more adequate for a skim through inspection.

### DIFF
--- a/prepare/metrics/bleu.py
+++ b/prepare/metrics/bleu.py
@@ -46,10 +46,10 @@ global_target = {
     "reference_length": 8,
     "score": 0.9,
     "score_name": "bleu",
-    "bleu_ci_low": 0.9,
-    "bleu_ci_high": 0.91,
-    "score_ci_low": 0.9,
-    "score_ci_high": 0.91,
+    "bleu_ci_low": 0.0,
+    "bleu_ci_high": 1.0,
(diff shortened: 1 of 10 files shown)


dafnapension:
wait for times or pettiness

PR #677, opened 2024-03-17
Fix bug in references with none

elronbandel:


src/unitxt/operators.py
```diff
@@ -1480,6 +1480,11 @@ class RemoveValues(FieldOperator):
 
     def verify(self):
         super().verify()
+        if self.process_every_value:
```
dafnapension, line 1483: why is this needed?
the operator operates on whats given it it. The operator just needs to verify that it is given a list and not a scalar.  Why should it care of the source of what is given to it? whether it was every value of a parent list or not?

elronbandel: I dont know, but it does not part of this PR its part of the reverting of your last commit. (to fix the bugs in the most minimal way) we can do this in a different commit.

PR #680, opened 2024-03-18
removal of dpath -- ready for review

dafnapension:
Hi Elron and Yoav,
d_path is out of dict_utility, and of both requirement files where it showed.

I also cancelled the ban of  `process_every_value`  not allowed for operator `RemoveValues` (a ban laid only on it..)
We do need `process_every_value` here.

Speaking of which (the family of field operators, including all the postprocessors, that process a (single) prediction and `process every value` of list of references), why is `EncodeLabels`  so much of an outsider?  
There is a straightforward implementation thereof as a field operator. Why was not it used?

### DIFF
--- a/requirements/base.rqr
+++ b/requirements/base.rqr
@@ -2,6 +2,5 @@ datasets>=2.16.0
 evaluate
 mecab-python3
 absl-py
-dpath
 ipadic
 scipy
--- a/requirements/docs.rqr
+++ b/requirements/docs.rqr
@@ -8,6 +8,5 @@ sacrebleu
 absl-py
 rouge_score
 scikit-learn
-dpath
 jiwer
 editdistance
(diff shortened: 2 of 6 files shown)


dafnapension:
@elronbandel , @yoavkatz , I suggest to remove the boolean parameter  `set_multiple` from `dict_set`.  
If the * comes at the end of the query, the same result can achieved by assigning a whole vector (for arg `value` of `dict_set()`);  and if the * comes in the middle of the query, it is not always clear that the meaning was breaking the input value (the vector) into its individual components, and assign one to each path. And what if more than one * on the query?   
Other than enjoying to play with it in the testing, I do not see any real value for this `set_multiple` parameter.  

Again: the meaning thereof is clear if * comes at the end, but the same result is achievable by assigning the whole vector as one unit, to the parent of the * in the query.
The meaning thereof for * in the middle of the query was guessed by me, and thus implemented..  Can see some examples in `test_dict_utils`.

elronbandel (approved):
This is amazing Dafna! I think any comments on top of this implementation should be in a seperate PR as this one defentily satisfy its original goal! well done!

Issue #681, opened 2024-03-18
New operator to  duplicate instance in a stream


yoavkatz:
Add a new operator that duplicate the each instance a given number of times. 

 This is to allow , for example, same questions with different order of answers  in multiple choice, or to allow the model to solve generate multiple answer for the same question with some randomness (this latter point would also require changes to the seed mechanism).

steps:

1. Add an DuplicateInstances operator in unitxt/src/unitxt/operators.py.

It should receive a "num_duplications" int parameter

This should be similar to the ilterByCondition operator (except that instead of filtering, it should "yield" the same instance multiple times.  Not the value yielded  has to be deep copied, so changes to one instances will not effect other instances.

2. Add tests in unitxt/tests/library/test_operators.py

elronbandel:
I think it should be duplicated by expanding a field, for example if you have few questions you want new instance per question. Just something to consider.  

p.s. in the context of conversations it will be useful to have an operator that duplicate instance by every sub list. so if you have conversation you can have instance with the conversation until turn i for every turn.

pawelknes:
> I think it should be duplicated by expanding a field, for example if you have few questions you want new instance per question. Just something to consider.
> 
> p.s. in the context of conversations it will be useful to have an operator that duplicate instance by every sub list. so if you have conversation you can have instance with the conversation until turn i for every turn.

I've made for now an operator which just simply yields deep copies of instances. That was what you meant or the operator should rather yield iterables of `num_duplications` copies for each instance? Something like below:

```
{"a": 1, "b": 2} -> {"a": [1, ...], "b": [2, ...]}
```

yoavkatz:
I think we should add another option to specify a field to store a duplication id with a running index (0,1,2....num_duplications-1).  

This will also us to perform operation based on this (for example, truncating a list as @elronbandel mentioned)

pawelknes:
So we want to have something like below?
```
[{"a": 1, "b": 2}] -> [{"a": 1, "b": 2, "duplication_id": 0}, {"a": 1, "b": 2, "duplication_id": 1}
```

yoavkatz:
> So we want to have something like below?
> 
> ```
> [{"a": 1, "b": 2}] -> [{"a": 1, "b": 2, "duplication_id": 0}, {"a": 1, "b": 2, "duplication_id": 1}
> ```

Yes.  The duplication_id field name should be a parameter of the operator.

yoavkatz:
Done. Thanks a lot.

PR #682, opened 2024-03-18
DuplicateInstances operator

pawelknes:
Added new operator `DuplicateInstances` which duplicates each instance in stream a given number of times

For: #681


elronbandel (approved):
This was fast!

PR #687, opened 2024-03-19
Duplicate instance operator - new functionality

pawelknes:
For #681 

Operator can now also add field `duplication_id` to duplicated instances, which will contain id of each duplication


src/unitxt/operators.py
```diff
@@ -1911,14 +1911,21 @@ class DuplicateInstances(SingleStreamOperator):
 
     Attributes:
         num_duplications (int): How many times each instance should be duplicated (1 means no duplication).
+        duplication_index_field (bool):
```
yoavkatz, line 1914: I think this should be a string, so the name of the index will not be hard coded.

yoavkatz (changes requested):
Please see the requested change.

yoavkatz (approved):

PR #688, opened 2024-03-19
Add safe and complete type parsing function to type_utils, for allowing better type checking.

elronbandel:


src/unitxt/utils.py
```diff
@@ -82,3 +84,30 @@ def is_module_available(module_name):
...
+
+    Note:
+        This function should be used carefully, as it employs `eval`, which can
+        execute arbitrary code. The function attempts to mitigate security risks
+        by restricting the available tokens and not exposing built-in functions.
+    """
+    allowd_sub_strings = list(context.keys()) + allowed_tokens
+    if is_made_of_sub_strings(expression, allowd_sub_strings):
+        return eval(expression, {"__builtins__": {}}, context)
+    raise ValueError(f"Unreocgnized Expression: {expression}")
```
yoavkatz, line 113: Please add a more informative message.  

The following expression is can not be evaluated because it contains tokes outside the allowed list of "...."

yoavkatz (approved):

Issue #695, opened 2024-03-20
Support loadTSV


benjaminsznajder:
Currently, we support CSV format into the loadCSV https://github.com/IBM/unitxt/blob/main/src/unitxt/loaders.py#L228

it is easy to add support to TSV format by adding the sep parameter into the pd.read_csv call.

PR #696, opened 2024-03-20
Add pandas_load_args for LoadCSV

elronbandel:
Closes: https://github.com/IBM/unitxt/issues/695

### DIFF
--- a/tests/library/test_loaders.py
+++ b/tests/library/test_loaders.py
@@ -75,13 +75,13 @@ class TestLoaders(UnitxtTestCase):
             dfs = {}
 
             for file in ["train", "test"]:
-                path = os.path.join(tmp_dir, file + ".csv")  # Adding a file extension
+                path = os.path.join(tmp_dir, file + ".tsv")  # Adding a file extension
                 df = pd.DataFrame({"x": [1, 2, 3, 4, 5]})  # Replace with your data
                 dfs[file] = df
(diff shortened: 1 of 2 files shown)


tests/library/test_loaders.py
```diff
@@ -68,6 +68,28 @@ def test_load_csv(self):
...
                     self.assertEqual(saved_instance[1].to_dict(), loaded_instance)
 
+    def test_load_csv_with_pandas_args(self):
+        # Using a context for the temporary directory
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            files = {}
+            dfs = {}
+
+            for file in ["train", "test"]:
+                path = os.path.join(tmp_dir, file + ".csv")  # Adding a file extension
```
benjaminsznajder, line 78: not critical, but makes sense to call it tsv.

benjaminsznajder (commented):
It works. But I have a general comment.

The API by requesting optional pandas_load_args argument makes the assumption that the implementation is by pd.read_csv()
The user can not make this assumption... shouldn't we just give the 'sep' argument?

tests/library/test_loaders.py
elronbandel: done

elronbandel:
Done.

benjaminsznajder (approved):

Issue #690, opened 2024-03-20
Add DynamicFormat that uses different SystemFormat based on size of demos list and existence of instruction.


elronbandel:
```python 
class DynamicFormat(Format):
  few_shot_format: Format
  few_shot_with_instruction_format: Format
  zero_shot_format: Format
  zero_shot_with_instruction_format: Format
      
  def process(instance):
     has_instruction =  "instruction" not in instance or len(instance["instruction"] == 0
     has_demos =  len(instance["demos"]) == 0
     if has_demos:
         if has_instruction:
             format = self.few_shot_with_instruction_format
         else:
             format = self.few_shot_format
     else:
         if has_instruction:
            format = self.zero_shot_with_instruction_format
         else:
            format = self.zero_shot_format
     return format.process_instance(instance)
```

PR #697, opened 2024-03-20
Implement dynamic format for selecting the correct format in case of

eladven:
dependency in the existence of instruction and num shots.

Solve issue  https://github.com/IBM/unitxt/issues/690


elronbandel (commented):
I think the name is too long. I personally much rather a name pepole can remember. Maybe MultiFormat?

PR #705, opened 2024-03-24
encode labels for single-value, and for all levels of nestings of containing lists

dafnapension:
Now, with multi-labeling, field references may become a "deep" (!) list.
`EncodeLabel `tried to chase the depth via specifying * and `set_multiple`  features of `dict_set`.
Basing on the fact that `EncodeLable` expects either a str or a list of str -s (and I understand that imminently -- a list of lists of str-s), suggested here is a simple `EncodeLabel` that needs no *, and no `set_multiple`, and is ready for the new references, no matter how deep their lists are.

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1682,20 +1682,18 @@ class Shuffle(PagedStreamOperator):
 class EncodeLabels(StreamInstanceOperator):
     """Encode each value encountered in any field in 'fields' into the integers 0,1,...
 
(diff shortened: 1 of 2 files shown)


dafnapension:
Dangerous if value is a list of scalars and lists

yoavkatz (dismissed):
Reviewed also by Elton.

dafnapension:
closing for now. Perhaps in the future, when more levels of list nesting will show in `references`.

PR #706, opened 2024-03-24
infer type_string from a given object, and infer_type therefrom via parse_type_string

dafnapension:


src/unitxt/type_utils.py
```diff
@@ -44,6 +44,43 @@ def parse_type_string(type_string: str) -> typing.Any:
...
 
+def encode_type_of_obj(obj: typing.Any) -> str:
+    """Encodes the type of a given object into a string.
+
+    Args:
+        obj:Any
+
+    Returns:
+        a string representation of the type of the object. e.g. 'str', 'List[int]', 'Dict[str, Any]'
+
```
dafnapension, line 55: Does not work for union-s

elronbandel (changes requested):
I would call it `infer_type` and it should return type rather than string describing the type.

dafnapension:
@elronbandel , completed the  object->`infer_type_string`, and readily added  `infer_type` as you noted.

elronbandel (approved):

Issue #693, opened 2024-03-20
Add type checking for task definition.


yoavkatz:
Today, tasks fields have no types

```
FormTask(
inputs=["text", "text_type", "class"],
outputs={"class" , "label"} ,
metrics=[
    "metrics.f1_micro_multi_label", 
    "metrics.f1_macro_multi_label",  
    "metrics.accuracy",
],
 )
```

This makes it hard for the user to know how to transform their dataset into the right format .  They can mistake pass a list for a string, for example and get odd errors or even just unexpected behavior .

We want to add typing definitions:

```
FormTask(
     inputs={"text":"str", "text_type":"str", "class": "str"},
     outputs={"class": "str", "label":"List[str]"} ,
     prediction_type="str". 
     metrics=[
         "metrics.f1_micro_multi_label",   
         "metrics.f1_macro_multi_label", 
          "metrics.accuracy"
        ]
)
```

So the FormTask operator will check that all the fields exist and are of the correct type.
 It will also check that the prediction_type of the task is compatible with the prediction_type of all the metrics (#667 )

PR #710, opened 2024-03-25
Type checking for task definition

pawelknes:
For #693 

Modified definition of tasks so that type of each field in `inputs` and `outputs` is checked as well as `prediction_type` for metrics.

During instance processing, when a field is of a different type than expected one, an error will be raised. Also an additional attribute `prediction_type` has been added which is used to check if metrics are properly chosen based on their respective property. Changes are backward compatible; when a user passes just a list of field names, it will be converted to a dict with `Any` types, and `prediction_type` is inferred from used metrics in case it is not specified.


prepare/tasks/classification.py
```diff
@@ -3,8 +3,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["text", "text_type", "class"],
-        outputs=["class", "label"],
+        inputs={"text": "str", "text_type": "str", "class": "str"},
+        outputs={"class": "str", "label": "str"},
```
yoavkatz, line 7: label should be List[str] in binary.  This is because it can be either be an empty list or a list with one element which is the class name.

prepare/tasks/completion/multiple_choice.py
```diff
@@ -3,8 +3,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["context", "context_type", "choices"],
-        outputs=["answer", "choices"],
+        inputs={"context": "str", "context_type": "str", "choices": "List[str]"},
+        outputs={"answer": "str", "choices": "List[str]"},
```
yoavkatz, line 7: Answer should be an int .  It's the index of the correct answer from "choice".

prepare/tasks/completion/multiple_choice.py
```diff
@@ -23,8 +25,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["context", "context_type", "completion_type"],
-        outputs=["completion"],
+        inputs={"context": "str", "context_type": "str", "completion_type": "str"},
+        outputs={"completion": "str"},
+        prediction_type="Any",
```
yoavkatz, line 30:
```suggestion
        prediction_type="Dict[str,Any]",
```

prepare/tasks/generation.py
```diff
@@ -3,8 +3,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["input", "type_of_input", "type_of_output"],
-        outputs=["output"],
+        inputs={"input": "str", "type_of_input": "str", "type_of_output": "str"},
+        outputs={"output": "str"},
+        prediction_type="Any",
```
yoavkatz, line 8:
```suggestion
        prediction_type="str",
```

prepare/tasks/qa/multiple_choice/tasks.py
```diff
@@ -3,8 +3,14 @@
...
-        inputs=["context", "context_type", "question", "choices"],
-        outputs=["answer", "choices"],
+        inputs={
+            "context": "str",
+            "context_type": "str",
+            "question": "str",
+            "choices": "List[str]",
+        },
+        outputs={"answer": "str", "choices": "List[str]"},
+        prediction_type="Any",
```
yoavkatz, line 13: I think the answer should be an "int", but I think right now the post_processor return a "str"

prepare/tasks/qa/multiple_choice/tasks.py
```diff
@@ -14,8 +20,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["topic", "question", "choices"],
-        outputs=["answer", "choices"],
+        inputs={"topic": "str", "question": "str", "choices": "List[str]"},
+        outputs={"answer": "str", "choices": "List[str]"},
+        prediction_type="Any",
```
yoavkatz, line 25: Change all QA prediction_Type to "str"...

prepare/tasks/qa/tasks.py
```diff
@@ -24,8 +26,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["question"],
-        outputs=["answers"],
+        inputs={"question": "str"},
+        outputs={"answers": "str"},
```
yoavkatz, line 30: Answers should be a list[str]

prepare/tasks/translation/directed.py
```diff
@@ -3,8 +3,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["text", "source_language", "target_language"],
-        outputs=["translation"],
+        inputs={"text": "str", "source_language": "str", "target_language": "str"},
+        outputs={"translation": "str"},
+        prediction_type="Any",
```
yoavkatz, line 8:
```suggestion
        prediction_type="str",
```

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
+    def process_type(data_type: Any) -> Any:
+        if isinstance(data_type, str):
+            data_type = data_type.replace("typing.", "")
+            return parse_type_string(data_type)
+        return data_type
+
     def verify(self):
+        for io_type in ["inputs", "outputs"]:
+            data = self.inputs if io_type == "inputs" else self.outputs
+            if not isinstance(data, Dict):
```
yoavkatz, line 50:
```suggestion
            if not isinstance(data, Dict[str,str]):
```

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
+                get_logger().warning(
+                    f"'{io_type}' should be a dict of field names and their types, instead only "
```
yoavkatz, line 52:
```suggestion
                    f"'{io_type}' field of Task should be a dictionary of field names and their types. For example, { 'text': 'str', 'classes': 'List[str]' }.   Instead "
```

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
+                    f"of unitxt this will raise an exception."
+                )
+                data = {key: Any for key in data}
+                if io_type == "inputs":
+                    self.inputs = data
+                else:
+                    self.outputs = data
+
+        if not self.prediction_type:
+            get_logger().warning(
```
yoavkatz, line 63: I think in this case, it should be set to 'Any'.

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
+                "'prediction_type' was not passed. It must be compatible with used metrics. "
```
yoavkatz, line 64:
```suggestion
                "'prediction_type' was not set in Task.  It is used to check the output of template post processors is compatible with the expected input of the metrics.   "
```

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
+                "Trying to infer `prediction_type` based on specified metrics. In future "
```
yoavkatz, line 65:
```suggestion
                "Setting `prediction_type`  to 'Any' (no checking is done). In future "
```

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
         "metrics" -- to contain the value of Arg 'metrics'
-
     """
 
-    inputs: List[str]
-    outputs: List[str]
+    inputs: Union[Dict[str, Any], List[str]]
+    outputs: Union[Dict[str, Any], List[str]]
     metrics: List[str]
+    prediction_type: Optional[Any] = None
```
yoavkatz, line 37:
```suggestion
    prediction_type: Optional[str] = None
```

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
+                or metric_prediction_type == Any
+            ):
+                continue
+
+            raise ValueError(
+                f"Given prediction type '{prediction_type}' and metric '{metric_name}' "
+                f"prediction type '{metric_prediction_type}' are different."
+            )
+
+    def access_instance_value(
```
yoavkatz, line 98: Maybe it's a style, but I think it would be clearer to have two separate functions

get_input_value(instance, key)
get_output_value(instance_key)

The erro message will be slightly repeated, but the code would be clearer.

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
-    inputs: List[str]
-    outputs: List[str]
+    inputs: Union[Dict[str, Any], List[str]]
+    outputs: Union[Dict[str, Any], List[str]]
     metrics: List[str]
+    prediction_type: Optional[Any] = None
     augmentable_inputs: List[str] = []
 
+    @staticmethod
+    def process_type(data_type: Any) -> Any:
```
yoavkatz, line 41: Perhaps avoid these conditional type conversion, see below in usage.

src/unitxt/task.py
```diff
@@ -10,42 +13,141 @@ class Tasker:
...
-            ) from e
+    def check_metrics_type(self) -> None:
+        for metric_name in self.metrics:
+            metric = fetch_artifact(metric_name)[0]
+            metric_prediction_type = metric.get_prediction_type()
+
+            if self.prediction_type == metric_prediction_type:
+                continue
+
+            prediction_type = self.process_type(self.prediction_type)
```
yoavkatz, line 84: instead of process_type.  Let's create get_prediction_type() in the Task, which returns the type after parsing.  And do all comparison between the types.

tests/library/test_tasks.py
```diff
@@ -51,3 +53,51 @@ def test_multiple_choice_task(self):
...
+            outputs={"label": str},
+            prediction_type="str",
+            metrics=["metrics.wer", "metrics.rouge", "metrics.roc_auc"],
+        )
+
+        operator.check_metrics_type()
+
+        operator.prediction_type = "Dict"
+        with self.assertRaises(ValueError):
+            operator.check_metrics_type()
```
yoavkatz, line 69: Best to also check the message - as in:

 with self.assertRaises(RuntimeError) as e:
            check_operator(
                operator=FilterByCondition(
                    values={"b": ["weird_value"]}, condition="in"
                ),
                inputs=inputs,
                targets=[],
                tester=self,
            )
        self.assertEqual(
            str(e.exception),
            "FilterByCondition filtered out every instance in stream 'test'. If this is intended set error_on_filtered_all=False",
        )

yoavkatz (changes requested):
Thanks @pawelknes .  Please see my  embedded comments.

prepare/tasks/generation.py
pawelknes: When calling `get_prediction_type` for `normalized_sacrebleu` it returns `Any` as the metrics does not have the `prediction_type` parameter set. So in such a case, we want the attribute to be compatible with a given task?

src/unitxt/task.py
pawelknes: I intended to make it more general so that a user can pass a direct type as well, not only type strings. For example: 
`{"some_field": List[str]}` would be also valid instead of only `{"some_field": "List[str]"}`. In that case, the `data` from above would be `Dict[str,Any]`.
What do you think? Or should I change it so that types are specified as strings only?

yoavkatz: I also opted to use the actual type,  but it does not work (I had the same issue with the metrics).  This is because Dict[str,Any] is not serializable and hence would not be written to the json correctly.   So. I suggest using only string.

elronbandel (changes requested):
I suggest to give it a new name: `TaskSchema` (or similar) and `@deprecation` decorator to the the old `FormTask`.

The syntax should be:
```python 
from .deprecation_utils import deprecation

@deprecation("1.8.0", alternative="TaskSchema")
class FormTask:
```

This will allow us to release version that will be compatible with the old `FormTask` and raise a warning so pepole whos using it can adjust the task themselves.

pawelknes:
> Thanks @pawelknes . Please see my embedded comments.

I've applied suggested changes and also found wrong typing for regression in task definition so it's fixed as well

pawelknes:
> I suggest to give it a new name: `TaskSchema` (or similar) and `@deprecation` decorator to the the old `FormTask`.
> 
> The syntax should be:
> 
> ```python
> from .deprecation_utils import deprecation
> 
> @deprecation("1.8.0", alternative="TaskSchema")
> class FormTask:
> ```
> 
> This will allow us to release version that will be compatible with the old `FormTask` and raise a warning so pepole whos using it can adjust the task themselves.

I tried to add this decorator but the entire code fails. The cause seems to be `MultipleChoiceTask` as it inherits from `FormTask` (commenting that class out or removing decorator from `FormTask` solves the issue). I'll investigate why it fails and how to fix it

prepare/tasks/ner.py
```diff
@@ -3,8 +3,14 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["text", "entity_type"],
-        outputs=["spans_starts", "spans_ends", "text", "labels"],
+        inputs={"text": "str", "entity_type": "str"},
+        outputs={
+            "spans_starts": "List[str]",
```
yoavkatz, line 8: Shouldn't this be List[int]? If not this way right now, then can you open a separate issue for correcting the types?

prepare/tasks/qa/tasks.py
```diff
@@ -3,8 +3,9 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["context", "context_type", "question"],
-        outputs=["answers"],
+        inputs={"context": "str", "context_type": "str", "question": "str"},
+        outputs={"answers": "List[str]"},
+        prediction_type="Any",
```
yoavkatz, line 8: Should this be "str"?

prepare/tasks/ner.py
pawelknes: Right, it's like this in `span_labeling` so I guess it'll be the same here

prepare/tasks/qa/tasks.py
pawelknes: I tried to find `prediction_type` but: `metrics.squad metric does not set the 'prediction_type' parameter so input type checking is not performed`. I can change it to `str` though and see if it passes

src/unitxt/task.py
```diff
@@ -10,42 +13,145 @@ class Tasker:
...
+        value_name: str,
+        io_type: Literal["inputs", "outputs"],
+    ) -> Any:
+        data_type = self.get_prediction_type(data_type)
+
+        if isoftype(value, data_type):
+            return value
+
+        raise ValueError(
+            f"Passed {io_type} value {value} under key {value_name} is not of required "
```
yoavkatz, line 124:
```suggestion
            f"Passed {io_type} value '{value}' of field  {field_name} is not of required "
```

src/unitxt/task.py
```diff
@@ -10,42 +13,145 @@ class Tasker:
...
-                f" \n available names:{list(instance.keys())}\n given output names:{self.outputs}"
+                f"Unexpected Task 'outputs' names: {[k for k in self.outputs if k not in instance]}."
+                f"The available names: {list(instance.keys())}."
             ) from e
 
+    def check_instance_value_type(
+        self,
+        value: Any,
+        data_type: str,
+        value_name: str,
```
yoavkatz, line 115:
```suggestion
       field_name: str,
```

src/unitxt/task.py
```diff
@@ -10,42 +13,145 @@ class Tasker:
...
+
+            if (
+                prediction_type == metric_prediction_type
+                or prediction_type == Any
+                or metric_prediction_type == Any
+            ):
+                continue
+
+            raise ValueError(
+                f"Given prediction type '{prediction_type}' and metric '{metric_name}' "
```
yoavkatz, line 89:
```suggestion
                f"The task's  prediction type ({prediction_type}) and '{metric_name}' metric's "
```

src/unitxt/task.py
```diff
@@ -10,42 +13,145 @@ class Tasker:
...
+                f"prediction type '{metric_prediction_type}' are different."
```
yoavkatz, line 90:
```suggestion
                f"prediction type ('{metric_prediction_type}') are different."
```

prepare/tasks/regression/tasks.py
```diff
@@ -3,8 +3,14 @@
 
 add_to_catalog(
     FormTask(
-        inputs=["text", "attribute_name", "min_value", "max_value"],
-        outputs=["attribute_value"],
+        inputs={
+            "text": "str",
+            "attribute_name": "str",
+            "min_value": "str",
```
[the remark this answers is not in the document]
pawelknes: #731

pawelknes:
@yoavkatz 
There is an issue with type of `label` in `outputs` in `tasks.classification.multi_class`. Previously it was set to `List[str]`, however, that caused errors during testing. I changed it to `str` and while those tests now pass, 3 others fail as they require `List[str]`. It happens for cards: `claim_stance_topic`, `head_qa`, `argument_topic`. They all return one-element list though, for example: `["biology"]`.
Is it a problem with that particular data or the field itself should be changed to union?

elronbandel:
@pawelknes thanks for noting its now fixed: https://github.com/IBM/unitxt/pull/735

src/unitxt/task.py
```diff
@@ -10,42 +13,145 @@ class Tasker:
...
+
+        raise ValueError(
+            f"Passed {io_type} value '{value}' of field {field_name} is not of required "
+            f"type ({data_type})."
+        )
+
+    def process(
+        self, instance: Dict[str, Any], stream_name: Optional[str] = None
+    ) -> Dict[str, Any]:
+        self.check_metrics_type()
```
elronbandel, line 131:
```suggestion
```
This should be execcuted in the verify() function rather than for every instance.

src/unitxt/task.py
```diff
@@ -10,42 +13,145 @@ class Tasker:
...
+        }
+        outputs = {
+            key: self.check_instance_value_type(
+                self.get_output_value(instance, key),
+                data_type,
+                key,
+                "outputs",
+            )
+            for key, data_type in self.outputs.items()
+        }
```
elronbandel, line 153: Instead of implementing this logic twice, i suggest to add a function to `type_utils.py`:
```python
def verify_required_schema(required_schema_dict, input_dict):
    ...
```
Reasons: (1) It should be tested as an independent unit. (2) In the near future it will be used for other operators (to possibly verify both their inputs and outputs).

pawelknes:
@yoavkatz @elronbandel I found that for card `coedit` used metrics have `prediction_type` only `str` instead of `List[str]`. I tried to change it with something like below:
`task=f"tasks.classification.binary[metrics=[{','.join(error_detection_metrics)}],prediction_type={'str'}]"` ,
and while that solution seems to be alright, also its label is `str` instead of `List[str]`. So I wanted to ask if this is expected and I should just change types in this particular case, or the card should have the same types like others in binary classification?

pawelknes:
> @yoavkatz @elronbandel I found that for card `coedit` used metrics have `prediction_type` only `str` instead of `List[str]`. I tried to change it with something like below: `task=f"tasks.classification.binary[metrics=[{','.join(error_detection_metrics)}],prediction_type={'str'}]"` , and while that solution seems to be alright, also its label is `str` instead of `List[str]`. So I wanted to ask if this is expected and I should just change types in this particular case, or the card should have the same types like others in binary classification?

Similar case happens in `test_artifact_loading_with_overwrite_args_list` in `test_artifacts`; Rouge metrics has `str` as its `prediction_type`, and default metrics: `f1_micro_multi_label` and `f1_macro_multi_label` have `List[str]` (`accuracy` uses `Any`)

yoavkatz:
This relates to a long discussion in related to #736.  I think we have now two types of binary tasks.  One that return list[Str] either ["Class-name"] or [], and one that return either 0 or 1 like this one:


I think we need to make these two separate tasks. So they won't conflict in types (in the future we need to unite them)

So I suggest we add:

tasks.classification.binary.zero_or_one with type "str"

pawelknes:
> > I suggest to give it a new name: `TaskSchema` (or similar) and `@deprecation` decorator to the the old `FormTask`.
> > The syntax should be:
> > ```python
> > from .deprecation_utils import deprecation
> > 
> > @deprecation("1.8.0", alternative="TaskSchema")
> > class FormTask:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will allow us to release version that will be compatible with the old `FormTask` and raise a warning so pepole whos using it can adjust the task themselves.
> 
> I tried to add this decorator but the entire code fails. The cause seems to be `MultipleChoiceTask` as it inherits from `FormTask` (commenting that class out or removing decorator from `FormTask` solves the issue). I'll investigate why it fails and how to fix it

@elronbandel There was an issue with adding that decorator to the class since the code failed when `MultipleChoiceTask` tried to inherit from `FormTask` (I believe it's due to using dataclass). 
Instead of using the decorator for `__init__`, I added it to the `process` method of `FormTask` and slightly modified deprecation utils so that some custom name may be passed which is displayed in a warning/error message.

I tried to somehow solve this issue but that was the simplest "fix" which worked. Let me know if I should change approach

prepare/tasks/regression/tasks.py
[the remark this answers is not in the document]
yoavkatz: Yes. I think there should be a post processor than converts from string to float. Today it's done in the pipeline:
unitxt/prepare/metrics/spearman.py

metric = MetricPipeline(
    main_score="spearmanr",
    preprocess_steps=[
        CopyFields(field_to_field=[("references/0", "references")]),
        CastFields(
            fields={"prediction": "float", "references": "float"},
            failure_defaults={"prediction": 0.0},
            use_nested_query=True,
        ),
    ],
    metric=Spearmanr(),
)

I think this casting code should move  to the preprocessor (e.g. "processors.cast_to_float_return_zero_if_failed")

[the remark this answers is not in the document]
pawelknes: I was thinking about this and I'm not sure why this should be moved to `processors`. Won't it be enough to just add it to `preprocess_steps` in `TaskCard` definition. For example:

 ```
       AddFields(
            fields={"attribute_name": "similarity", "min_value": 1.0, "max_value": 5.0}
        ),
        CastFields(
            fields={"min_value": "float", "max_value": "float"},
            failure_defaults={"min_value": 0.0, "max_value": 0.0},
        ),
```

Also, `CastFields` is used in a lot of different places in its current form.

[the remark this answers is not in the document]
yoavkatz: The min and max values are already floats - so you don't need to cast them.
The main point is not to cast the *dataset inputs* but rather to convert the *model predictions* to float.
The model returns a string - and the metric expects a float, so the post processor needs to take the model prediction and cast it to float.  If the case fail, spearman still expect a float to we should cast it to zero , like before.

This is done by the post processors (not the card preprocessors).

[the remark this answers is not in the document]
pawelknes: Okay, I thought that the output is float by default. Now it's clear, thanks

yoavkatz (approved):

PR #711, opened 2024-03-25
Fix handling of boolean environment variables

arielge:
Currently when unitxt settings are set via environment variables, boolean parameters are processed as strings, and thus a value of "False" can be evaluated as True by unitxt. This change casts the relevant setting values to boolean


yoavkatz (approved):

elronbandel (changes requested):
Nice catch! This means the cast_to_type function should be used also in the get attr. Using this function is the best way to handle all possible types in one place. Can you change it so this function will be used?

PR #713, opened 2024-03-26
add perplexity with Mistral model

lilacheden:


elronbandel (approved):
Only regarding the name, not sure if it will be easier to parse it was:
`metrics.rag.context_relvance.mistral` that way all the context_relvance metrics will be in the same directory dedicated to context_relvance metrics.

lilacheden:
> Only regarding the name, not sure if it will be easier to parse it was: `metrics.rag.context_relvance.mistral` that way all the context_relvance metrics will be in the same directory dedicated to context_relvance metrics.

This would require to also change the existing metrics.rag.context_relevance to metrics.rag.context_relevance.flan_t5_small. 
Didn't do it for backwards compatibility, but I can change it if you prefer.

PR #727, opened 2024-03-28
Ready for review: remove use_dpath parameter from dict_get and dict_set

dafnapension:
(1) removed argument use_dpath from dict_get
(2) removed argument use_dpath from dict_set
(3) removed use_query flag from all operators
(4) fixed  code of dict_get, to have *, when being last component on qpath, match an empty list too.
(5) added more tests to test_dict_utils, and improved dealing with edge-conditions


elronbandel (approved):
Is the strip in the query needed? Other than that everything looks great.

PR #733, opened 2024-03-28
fix wrong handling of list in dict_get

yoavkatz:
When accessing a nested empty list, returned an error, while this is legal.

Also added tests.


elronbandel (changes requested):
I think we should have a test for cases where you ask for the nth item in the list (e.g `references\0`) and the list is empty.

yoavkatz:
> I think we should have a test for cases where you ask for the nth item in the list (e.g `references\0`) and the list is empty.

Done.

elronbandel (approved):

PR #736, opened 2024-03-31
Ability to dynamically change InstanceMetric inputs + grammar metrics

arielge:


yoavkatz:
Ariel, I'm not sure about the generic part change.   It breaks unitxt concepts.  
The task_data defines the fields of the task which should be shared by all metric.
So it's not a place to pass specific field names to specific metrics..

elronbandel (approved):

arielge:
@yoavkatz The task fields are still shared across all metrics. But there are many use cases - specifically, reference-less ones - where you want to apply an existing metric but use a different field (an input field, which is still one of the task fields). I don't think it makes sense to create a new metric for every possible field that one would want to pass.
@elronbandel

yoavkatz:
I see the point.  Some points to consider:

image


1. There is an implicit conversion to list  - which will be hard to keep track of.  And will cause problems when passing references which are lists (e.g. when the "references" are a list of  of context_ids, for example).

arielge:
So I do not have the full picture, but regarding both 2 and 3 my feeling is that the real underlying issue is that the different types of metric classes do not really share methods, which is something that should probably be improved regardless of this specific question. Once the different types of metrics have more methods in common, things like the field modification I did here and the type checking can both happen in a more centralized place, which will solve most of the issues you mention.

I do understand what you are saying about complicating the API. But since tasks -- even similar ones -- often differ in the name of their input fields, my feeling is that the cost of having a separate metric for every possible field name will very quickly blow out of proportion, which is also a serious problem. In other words, the API may be simpler but the catalog will be more complicated.

yoavkatz:
@arielge - it think it's not only a matter of which filed to use  We should also give clear score names

right now we 

char_edit_distance 
chat_edit_distance_accuarcy 

But the second needs to be something like

char_edit_distance_from_original_text

 to be differentiated .

arielge:
I agree, this is what I was aiming for with this issue https://github.com/IBM/unitxt/issues/737. 
Now, if this is done not via overwrite, then you will need to add a metric to the catalog for `char_edit_distance_from_original_text`, and for another task you would add `char_edit_distance_from_input`, and for another you would add `char_edit_distance_from_document`, when they all do exactly the same calculation.

yoavkatz:
I see.   Today we have the explicit way of doing this:

metric = MetricPipeline(
    main_score="score",
    preprocess_steps=[
        ListFieldsValue(fields=["task_data/original_text"], to_field="references", use_query=True)
    ],
    metric=metrics["metrics.char_edit_distance"],
    postpreprocess_steps=[
         RenameFields(field="score/global/chat_edit_distance",to_field="score/global/chat_edir_distance_from_original_input"
            use_query=True)
    ],
)
add_to_catalog(metric, "metrics.char_edit_distance_from_original_text", overwrite=True)

I agree it would be nicer to have it shorter.

metrics.char_edit_distance[reference_field=original_text, score_name=char_edit_distance_from_original_input]

We need to make sure:
1. We always encapsulate reference_field in a list - to get consistent behavior. If we have need multiple references, we should have a explicit option like:

metrics.char_edit_distance[multiple_references_field=original_text, score_name=char_edit_distance_from_original_input]
 

2. Add the functionality to the base  Metric class.

arielge:
That sounds good to me

PR #739, opened 2024-03-31
Add production-time recipe processing capability to unitxt

elronbandel:
With this PR Unitxt can be used to process data in production. 

First define a recipe:

```python
  recipe = "card=cards.wnli,template=templates.classification.multi_class.relation.default,demos_pool_size=5,num_demos=2"
```

Second prepare an instance in the exact schema of the task in that recipe:

```python
  instance = {
    "label": "?",
    "text_a": "It works perfectly",
    "text_b": "It works!",
    "classes": ["entailment", "not entailment"],
    "type_of_relation": "entailment",
    "text_a_type": "premise",
    "text_b_type": "hypothesis",
  }
```

Then you can produce that model-ready data with the `produce` function:

```python
  from unitxt import produce

  result = produce(instance, recipe)
```

Then you have the production ready instance in the result. If you `print(instance["source"])` you will get:

```log
    Given a premise and hypothesis classify the entailment of the hypothesis to one of entailment, not entailment.

    premise: When Tatyana reached the cabin, her mother was sleeping. She was careful not to disturb her, undressing and climbing back into her berth., hypothesis: mother was careful not to disturb her, undressing and climbing back into her berth.
    The entailment class is entailment

    premise: The police arrested all of the gang members. They were trying to stop the drug trade in the neighborhood., hypothesis: The police were trying to stop the drug trade in the neighborhood.
    The entailment class is not entailment

    premise: It works perfectly, hypothesis: It works!
    The entailment class is
```


eladven (approved):

OfirArviv:
Can I past a list of instances and get them produces in one go?

OfirArviv:
also I want to be able to use the post processors without using the metric, is it possible?

OfirArviv (dismissed):
asked a few questions

docs/docs/production.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. note::
+
+   To use this tutorial, you need to :ref:`install unitxt <install_unitxt>`.
+
+=====================================
+Using in production
+=====================================
+
+Unitxt can be used to process data in production. First define a recipe:
```
OfirArviv, line 11: is production the correct word? i.e. dynamically, "on-the-fly", etc?

docs/docs/production.rst
```diff
@@ -0,0 +1,58 @@
...
+
+  recipe = "card=cards.wnli,template=templates.classification.multi_class.relation.default,demos_pool_size=5,num_demos=2"
+
+
+Second prepare an instance in the exact schema of the task in that recipe:
+
+
+.. code-block:: python
+
+  instance = {
```
OfirArviv, line 23: is there a way from a recipe to know its from a speficic task?
I want the metric to only accept cards that are from specific task, so I know the recipe beforehand?

src/unitxt/standard.py
```diff
@@ -121,9 +122,69 @@ def prepare_metrics_and_postprocessors(self):
             metrics = self.metrics
         return metrics, postprocessors
 
-    def prepare(self):
+    def set_pipelines(self):
+        self.loading = SequentialOperator()
+        self.metadata = SequentialOperator()
```
yoavkatz, line 127: I think this should be called "add_metadata" .

src/unitxt/standard.py
```diff
@@ -121,9 +122,69 @@ def prepare_metrics_and_postprocessors(self):
...
+        self.standardization = SequentialOperator()
+        self.processing = SequentialOperator()
```
yoavkatz, line 129: It's not clear what is the difference between "standardization" and "processing".

Maybe "standardize_to_task", "process_task_fields"

tests/library/test_api.py
```diff
@@ -38,3 +38,32 @@ def test_evaluate_with_metrics_external_setup(self):
         predictions = ["2.5", "2.5", "2.2", "3", "4"]
         results = evaluate(predictions, dataset["train"])
         self.assertAlmostEqual(results[0]["score"]["global"]["score"], 0.2, 3)
+
+    def test_produce_with_reicpe(self):
```
yoavkatz, line 42:
```suggestion
    def test_produce_with_recipe(self):
```

yoavkatz (dismissed):
I think it's a great feature.  See some comments . Specifically , the need to use existing cards is requires loading existing data from the data source, which I'm not sure is the natural way to use it and is also slow.

I think a dedicated recipe without a card (only receive task, an instance , and optional instances for demos ) will be more clear and efficient.

OfirArviv:
is it possible to ask the processor what is the type of instance its expecting? This way I can do validation beforhand

elronbandel:
@yoavkatz notice that loading the data happen only when the recipe requires demonstrations. I agree with the general claim, when someone want to produce data points without demos we should allow it without stating a card but instead stating a task.

docs/docs/production.rst
elronbandel: Yes, the regular use case of unitxt is either in training or evaluation. The use case being addressed here is the use in production time.

elronbandel:
> @OfirArviv   Can I past a list of instances and get them produces in one go?

done
> @OfirArviv also I want to be able to use the post processors without using the metric, is it possible?

In a saparate PR

> @OfirArviv is it possible to ask the processor what is the type of instance its expecting? This way I can do validation beforhand

The function as it is verify types schema compatibility

PR #741, opened 2024-03-31
Add DeprecatedFIeld for more informative procedure for deprecating fields of artifacts

dafnapension:


elronbandel (changes requested):
This was quick. While doing the job I have two concerns:
1) Seperation: so far dataclass was an independent module which does not know unitxt. It was done so to avoid circular imports and allow any unitxt module to use dataclass. I rather keep it that way if possible.
2) Self-Contained Logic: Every class in unitxt should be self contained and understood from its definition. That way when someone read the code of class they can understand the way it works. Having the deprecation declared in the settings dis obey this principle.

If we don't find other way, this PR does answer the need for more informed field deprecation process.

My alternative suggestion is to create in the data class a new field type: DeprecatedField which will still be a viable field just produce a warning with a custom message given to the field constructor. That way both my concerns are addressed.

What do you think @dafnapension ?

dafnapension:
I think that you are right as always, @elronbandel , and I respectfully accept your comment, and will implement!

dafnapension:
Hi @elronbandel , please see if my implementation gets any close to what you had in mind.

dafnapension:
Hi @elronbandel , please see my implementation per your metadata idea

elronbandel (approved):

PR #744, opened 2024-04-03
Add new char (\N) to our formatting language

eladven:
(In python you should write \\N). The char is an opportunistic \n means that it will be a newline if there is no newline before it and it is not the first char in the string.


elronbandel:
started more comprhensive PR here: https://github.com/IBM/unitxt/pull/751

PR #752, opened 2024-04-04
restructure aggregation of InstanceMetric , enabling to expand grouping and filtering option on to GlobalMetric

dafnapension:


yoavkatz:
@elronbandel , @dafnapension   

There are currently additional ways to group metrics (note I'm not sure if they are the best way either), but my general message is  we see how add functionality without expanding the APIs of metrics more and more.

We should have one recommended way to  group, and if we implement a new one, and replace other uses.

The proposed method here  allows arbitrary conditions, buts requires listing them specifically.  On the other hand, grouping by field value as done today is more natural is some cases, where the options are not known.

Another  suggestion is to share the proposed specification before the PR, so feedback can be given  early.

dafnapension:
Hi @yoavkatz ,  I hope the current suggestion for a class that takes care of splitting/grouping  is more in the direction you had in mind.
I think that if accepted, this class could also play a role in GlobalMetric (as it does in the current PR), and perhaps will unify all the different grouping that are tailored to Accuracy (and only to it). 
Another issue that I think we may want to address is the name of the score stored in ["score"]["global"] for this groupped version of the evaluation. 
I vote for generating children to ["score"]["global"], each child named by the group it represents (as suggested in this PR) and inside this child, things will look exactly as they look now (no splitting / grouping whats so ever) in ["global"]

src/unitxt/metrics.py
```diff
@@ -953,10 +998,91 @@ def aggregation_function(
...
+
+# aggregates (mean, max, or var) numeric scores read from a given field (score_field) found in instance["score"]["instance"]
+# and stores the result aggregated over instances into field named aggregate_score_field
+# (e.g  max_accuracy) in instance["score"]["global"]
+# potentially, the aggregation is carried over subsets of the instances, as induced by filter_grouper
+# the scores in instance["score"]["instance"] are expected to be there following the processing of an
+# InstanceMetric's  compute_instance_scores(stream)
+
+
+class Aggregator(Artifact):
```
yoavkatz, line 1030: Is this related to this PR?

yoavkatz:
> Hi @yoavkatz , I hope the current suggestion for a class that takes care of splitting/grouping is more in the direction you had in mind. I think that if accepted, this class could also play a role in GlobalMetric (as it does in the current PR), and perhaps will unify all the different grouping that are currently tailored to Accuracy (and only to it). Another issue that I think we may want to address is the name of the score stored in ["score"]["global"] for this groupped version of the evaluation. I vote for generating children to ["score"]["global"], each child named by the group it represents (as suggested in this PR) and inside this child, things will look exactly as they look now (no splitting / grouping whats so ever) in ["global"]

Yes. I think better.  Elron said there is a grander design, which we should review.

src/unitxt/metrics.py
dafnapension: A start of an attempt to unify the aggregators over the InstanceMetrics

dafnapension:
Hi @elronbandel and @yoavkatz,
I suggest here a restructuring of the different variations of aggregation (grouping - and then which ci, filtering, control-comparison) such that: (1) they are independent, and hence filtering and control-comparison are applicable to the whole stream, and not only when grouping, and (2) easily applicable to GlobalMetric as well, as demonstrated in the tests I added, as the variations 'engine' now sits in MetricWithConfidenceInterval.
Please see if this makes sense, and I will also complete for BulkInstanceMetric.

dafnapension:
replaced by https://github.com/IBM/unitxt/pull/845

Issue #715, opened 2024-03-26
Ability to parse all valid type aliases


pawelknes:
Currently function `parse_type_string` (in `src/unitxt/test_utils.py`) fails when a user passes built-in non-capitalised type aliases, for example: `list` instead of `List`, which is valid since Python 3.9. 
The same happens when using a "bitwise or operator" for union types, for example: `int | str` instead of `Union[int, str]`. 
The function also fails when a type string contains full module name for example: `typing.Dict`.

Suggestion: expanding the functionality so that `parse_type_string` can process all valid type aliases

pawelknes:
@elronbandel what do you think about this idea?

elronbandel:
@pawelknes it could be fixed easily but if i. understand correctly it wont work with python 3.8 no?

pawelknes:
> @pawelknes it could be fixed easily but if i. understand correctly it wont work with python 3.8 no?



Since `parse_type_string` accepts strings I was thinking about adding a step which will firstly preprocess string so that it's compatible with `typing` no matter if version is before or after Python 3.9, maybe like this snippet (just an initial idea):

```
def format_type_string(type_string: str) -> str:
    types_map = {"list": "List", "dict": "Dict", "tuple": "Tuple", "typing.": ""}
    for k, v in types_map.items():
        type_string = type_string.replace(k, v)
    return type_string
```

And it could also deal with `|` into `Union`. I can make a PR if you accept the idea (or am I missing something?)

elronbandel:
You are right. Sounds cool and should work!

pawelknes:
https://github.com/IBM/unitxt/pull/772

PR #754, opened 2024-04-08
Support for parsing all strings representing valid Python type hints

pawelknes:
For #715 

Extends functionality of parsing type strings so that they can represent any valid Python type hint no matter the version (before or after Python 3.9). This includes union types represented by `|` or non-capitalised types, for example `list` instead of `List` etc.


pawelknes:
@elronbandel I've added functionality discussed in #715 
The function `convert_union_type` is probably overcomplicated but that was the only approach which worked for all cases

tests/library/test_type_utils.py
```diff
@@ -262,3 +263,15 @@ def test_parse_unsupported_type(self):
...
             parse_type_string("List[[int]]")
+
+    def test_format_type_string(self):
+        self.assertEqual(
+            format_type_string(
+                "typing.Tuple[int,float,int|list[int|dict[str,int|float]]]"
+            ),
+            "Tuple[int,float,Union[int,List[Union[int,Dict[str,Union[int,float]]]]]]",
+        )
+        self.assertEqual(
```
dafnapension, line 274: Hi @pawelknes, I am confused: from the first assert, I learn that you want to convert `typing.Tuple` into `Tuple`, whereas from the second: `tuple` into typing.Tuple`.  Why, then, stop there, and not convert `tuple` into `Tuple'?

src/unitxt/type_utils.py
```diff
@@ -7,6 +7,114 @@
 from .utils import safe_eval
 
 
+def convert_union_type(type_string: str) -> str:
```
dafnapension, line 10: Hi @pawelknes , is the | allowed from python 3.9 or from 3.10?
I see this: https://peps.python.org/pep-0604/

tests/library/test_type_utils.py
pawelknes: The first assert just checks if `format_type_string` properly formats a type string so that it is compatible with `parse_type_string` function which requires strings in such format. 

While the second one performs the entire pipeline of parsing a type string into a type hint; `typing.Tuple[int, str, typing.List[typing.Union[int, float]]]` is just a type hint and is checked against output of `parse_type_string("tuple[int, str, typing.List[int | float]]")` to ensure if parsing works correctly for such type strings. It could be written as just `Tuple` without `typing.` but this is the same

src/unitxt/type_utils.py
```diff
@@ -7,6 +7,114 @@
...
+    """Converts new Python union type hints (valid since Python 3.9) into form compatible with any Python version.
+
+    Args:
+        type_string (str): A string representation of a Python type hint. Examples include
+                           'List[int]', 'Dict[str, Any]', 'Optional[List[str]]', etc.
+
+    Returns:
+        str: A type string which is compatible with typing module. For example
+            'Union[int,float]' instead of 'int|float'.
+
```
dafnapension, line 20: Hi @pawelknes , can you please formally describe the allowed input for your method, and also formally describe the generated output?

Is the following formal definition, using a mixture of regexes and Context Free Grammar,  in the direction of the input you have in mind?:

type -> basic | (typing.)?(List[type] | Dict[type, type] | Optional[type] | Tuple[type (, type)*] ) | type"|" type  ? 

basic ->  int, float, str, bool

Do you aim at all the "residents" of typing?  e.g. Literal, Annotated, Callable?

tests/library/test_type_utils.py
dafnapension: I see. thanks.    Do you flatten the Union-s?  e.g. Union inside Union?  or Union of some types, two of which are identical?   Or just transform from | to Union?

src/unitxt/type_utils.py
pawelknes: To be honest I didn't consider remaining types but it should work anyway as long as a string represents a correct Python type hint (I'll extend documentation as you suggested). The function works based on positions of sings: `'[', ']', ','` so it won't matter if a dict or a literal is used. I also ran the following test and the output is correct:

`convert_union_type('Tuple[Annotated[int|float,ValueRange(-10, 5)],List[str|Literal["string"]],Callable[[int|float],[bool,str|List[str]]]]') -> 'Tuple[Annotated[Union[int,float],ValueRange(-10, 5)],List[Union[str,Literal["string"]]],Callable[[Union[int,float]],[bool,Union[str,List[str]]]]]'`

tests/library/test_type_utils.py
dafnapension:
```
format_type_string(
                "Tuple[int,float,int|list[int|dict[str,int|float]]] | bool | str")  => 
```
                
`Union[Tuple[int,float,Union[int,List[Union[int,Dict[str,Union[int,float]]]]]],Union[bool,str]]`

I think that Union should not be used inside Union. I.e., I think that here, the expected result should be:
`Union[Tuple[int,float,Union[int,List[Union[int,Dict[str,Union[int,float]]]]]],bool,str]`

pawelknes: It just transforms a string, it doesn't simplify it in any way since that would probably require a different approach

pawelknes: > ```
> format_type_string(
>                 "Tuple[int,float,int|list[int|dict[str,int|float]]] | bool | str")  => 
> ```
> 
> `Union[Tuple[int,float,Union[int,List[Union[int,Dict[str,Union[int,float]]]]]],Union[bool,str]]`
> 
> I think that Union should not be used inside Union. I.e., I think that here, the expected result should be: `Union[Tuple[int,float,Union[int,List[Union[int,Dict[str,Union[int,float]]]]]],bool,str]`

You're right. I didn't test it for multiple options in a single union. I'll make a fix

src/unitxt/type_utils.py
dafnapension: What is the meaning of encapsulation by [ ]  "for no reason"?  I do not know..
I see in your example: `Callable[[int|float],[bool,str|List[str]]]`  what is the meaning of `[int | float]` ?  is it different from `int|float`  ?    Is this notatioin (encapsulating wihtin  [] ) allowed?

tests/library/test_type_utils.py
pawelknes: I thought about this and there is no need to parse a type hint further on a string level. It'll be done by `parse_type_string` function anyway; only type string needs to be properly formatted and that is done by the function (since `Union[a, b, c]`  is basically equivalent to `Union[a, Union[b, c]]`)

dafnapension: I see, so if format_type_string is only used as a service for parse_type_string , then its alright.

What do you do if the | sign shows inside a string, being a member of typing.Literal?

`print(format_type_string( "typing.Tuple[int,float,Literal[\"lef|t\"],int|list[int|dict[str,int|float]]]" ))`
prints:
`Tuple[int,float,Literal[Union["lef,t"]],Union[int,List[Union[int,Dict[str,Union[int,float]]]]]]`

dafnapension: I think that your code can not support these surprising Literals, 
However, the current `parse_type_string` in unitxt does not support Literal at all.
I am not sure there is a way to convert Literal (as you used in your code, for example: `Literal["left", "right"]`) to anything digestible by `parse_type_string` that is currently in unitxt.

So I think this should be made clearer in your documentation: it does not 
`"Formats a string representing a Python type hint so that it is compatible with any Python version (before or after Python 3.10)."`

such that the result (of your code) is correctly digestible by `parse_type_string` currently in unitxt.

I think that indeed, as you explained to me, your code can deal with all "residents" of typing of all python levels, because your code only depends on the [ ],  but this is correct for all residents other than resident "Literal".

pawelknes: Good point, I didn't consider that. I could add some additional rules for formatting insides of Literal, but since unitxt doesn't support it, I'll just exclude it and write that Literal isn't supported

dafnapension (changes requested):
Please correct the overall documentation, and in particular - the thrown error messages, as detailed in my comments in this PR

pawelknes:
> Please correct the overall documentation, and in particular - the thrown error messages, as detailed in my comments in this PR

I applied your suggestions and specified that `format_type_string` can format any valid type hint but it cannot contain strings and capitalisation is applied only to tokens allowed by unitxt, and left to `parse_type_string` raising errors in case `type_string` contains some not allowed types

dafnapension:
Thank you, I think the documentation is clear and more accurate now.

Possibly due to my education.. I am used to formally describe the input and output.
I approve your PR. But just for my curiosity: is the following a correct description of the input that your code anticipate in order for your union conversion to work properly?
    # type -> word | type(|type)* | word[type (,type)*]
    # word consists of a sequence of any char but: [ ] , |
    # this means that if any of these 4 chars shows not as a meta char of the input 
    # type_string, but inside some constant string (of Literal, for example), the scheme 
    # will not work.

If this defines the conditions needed for your union_conversion to work, then for these conditions, I can suggest a simpler recursion that can also deal with two or more |-s  (e.g.:   str|bool|List[float]  => Union[str, bool, List[float]] ).  

But since, as you pointed out, `parse_type_string` can digest also with non-flattened Union-s, this is not a must, and if you want, we can continue in a separate PR.

dafnapension (approved):

pawelknes:
> Thank you, I think the documentation is clear and more accurate now.
> 
> Possibly due to my education.. I am used to formally describe the input and output. I approve your PR. But just for my curiosity: is the following a correct description of the input that your code anticipate in order for your union conversion to work properly? # type -> word | type(|type)* | word[type (,type)*] # word consists of a sequence of any char but: [ ] , | # this means that if any of these 4 chars shows not as a meta char of the input # type_string, but inside some constant string (of Literal, for example), the scheme # will not work.
> 
> If this defines the conditions needed for your union_conversion to work, then for these conditions, I can suggest a simpler recursion that can also deal with two or more |-s (e.g.: str|bool|List[float] => Union[str, bool, List[float]] ).
> 
> But since, as you pointed out, `parse_type_string` can digest also with non-flattened Union-s, this is not a must, and if you want, we can continue in a separate PR.

Yup, I think it describes it correctly. Surely, it can be improved and simplified so we can continue in some other PR. There is an issue for it: https://github.com/IBM/unitxt/issues/715 so it could be attached to it

PR #757, opened 2024-04-10
fix a small bug in BinaryMaxAccuracy

dafnapension:
Currently:
      
```
 metric = BinaryMaxAccuracy()
 references = [["0"], ["0"], ["0"]]
 predictions = ["0.3", "0.5", "0"]
 outputs = apply_metric(
         metric=metric, predictions=predictions, references=references
 )
 print(outputs[0]["score"]["global"])
```
     
prints out:

```
{'max_accuracy_binary': 0.6666666666666666, 'best_thr_max_acc': 0.5, 'score': 
0.6666666666666666, 'score_name': 'max_accuracy_binary', 'score_ci_low': nan, 
'score_ci_high': nan, 'max_accuracy_binary_ci_low': nan, 
'max_accuracy_binary_ci_high': nan}
```       
       
Although clearly accuracy = 1.0 can be obtained for any thr > 0.5.  E.g for thr = 0.7.
       
This PR fixes this small bug, and while the hood is open, suggests an  n log n algorithm to replace the n^2 existing one, for the fun of it.

### DIFF
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -3356,7 +3356,7 @@ class BinaryMaxAccuracy(GlobalMetric):
     def compute(
         self,
         references: List[List[str]],
-        predictions: List[List[str]],
+        predictions: List[str],
         task_data: List[Dict],
     ) -> dict:
         float_predictions = [to_float_or_default(p) for p in predictions]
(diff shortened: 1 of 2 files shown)


src/unitxt/metrics.py
```diff
@@ -3356,32 +3356,60 @@ class BinaryMaxAccuracy(GlobalMetric):
...
+        current_thr = fp[0][0]
+        # partition float_predictions into all-passing, none-failing
+        current_acc = sum(r[0] == "1" for r in references)
+        # number of predictions that thr sends to the reference they are paired with
+
+        best_acc = current_acc
+        best_thr = current_thr
+
+        i = 0
+        while i < len(predictions):
```
arielge, line 3398:
```suggestion
        while i < len(predictions) and best_acc < len(predictions):
```

src/unitxt/metrics.py
```diff
@@ -3356,32 +3356,60 @@ class BinaryMaxAccuracy(GlobalMetric):
...
+            delta = fp[i][2]
+            i += 1
+            while i < len(predictions) and fp[i][0] <= fp[i - 1][0]:
```
arielge, line 3401:
```suggestion
            while i < len(predictions) and fp[i][0] == fp[i - 1][0]:
```
effectively the same, but this is what we actually check

arielge (changes requested):

src/unitxt/metrics.py
dafnapension: I agree. I just can not write (some psycho thing)  float == float..  so I go around..

src/unitxt/metrics.py
dafnapension: nice!  I accept!

arielge (approved):

PR #762, opened 2024-04-11
Allow FilterByCondition to condition also on subfields

dafnapension:
not only on keys in the instance

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1164,11 +1164,11 @@ class ApplyOperatorsField(StreamInstanceOperator):
 class FilterByCondition(SingleStreamOperator):
     """Filters a stream, yielding only instances for which the required values follows the required condition operator.
 
-    Raises an error if a required key is missing.
+    Raises an error if a required key/subfield is missing.
 
     Args:
        values (Dict[str, Any]): Values that instances must match using the condition to be included in the output.
-       condition: the name of the desired condition operator between the key and the value in values ("gt", "ge", "lt", "le", "ne", "eq")
+       condition: the name of the desired condition operator between the key/subfield and the value in values ("gt", "ge", "lt", "le", "ne", "eq")
        error_on_filtered_all (bool, optional): If True, raises an error if all instances are filtered out. Defaults to True.
 
     Examples:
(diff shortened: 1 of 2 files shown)


src/unitxt/operators.py
```diff
@@ -1164,11 +1164,11 @@ def process(
...
-    Raises an error if a required key is missing.
+    Raises an error if a required key/subfield is missing.
 
     Args:
        values (Dict[str, Any]): Values that instances must match using the condition to be included in the output.
-       condition: the name of the desired condition operator between the key and the value in values ("gt", "ge", "lt", "le", "ne", "eq")
+       condition: the name of the desired condition operator between the key/subfield and the value in values ("gt", "ge", "lt", "le", "ne", "eq")
        error_on_filtered_all (bool, optional): If True, raises an error if all instances are filtered out. Defaults to True.
 
     Examples:
```
yoavkatz, line 1174: Add example of nested field

src/unitxt/operators.py
```diff
@@ -1164,11 +1164,11 @@ def process(
 class FilterByCondition(SingleStreamOperator):
     """Filters a stream, yielding only instances for which the required values follows the required condition operator.
 
-    Raises an error if a required key is missing.
+    Raises an error if a required key/subfield is missing.
```
yoavkatz, line 1167:
```suggestion
    Raises an error if the field name is missing from the input instances.
```

src/unitxt/operators.py
```diff
@@ -1164,11 +1164,11 @@ def process(
...
 
     Args:
        values (Dict[str, Any]): Values that instances must match using the condition to be included in the output.
-       condition: the name of the desired condition operator between the key and the value in values ("gt", "ge", "lt", "le", "ne", "eq")
+       condition: the name of the desired condition operator between the key/subfield and the value in values ("gt", "ge", "lt", "le", "ne", "eq")
```
yoavkatz, line 1171:
```suggestion
       condition: the name of the desired condition operator between the specified field  and the provided value ("gt", "ge", "lt", "le", "ne", "eq")
```

dafnapension: Thanks for noticing! Corrected.

src/unitxt/operators.py
dafnapension: Done. Thanks.

src/unitxt/operators.py
dafnapension: Done. thanks!

PR #766, opened 2024-04-13
Adding NLI model via perplexity

assaftibm:


elronbandel (changes requested):
I strongly oppose to adding another templating mechanism in a library that is built for templating. 

The way it should be done is by:
1. Using the unitxt text completion and generation tasks. 
2. Adding your custom templates as templates for that task.
3. Using unitxt production api for generating the final data with the required template to the model.

yoavkatz:
> I strongly oppose to adding another templating mechanism in a library that is built for templating.
> 
> The way it should be done is by:
> 
> 1. Using the unitxt text completion and generation tasks.
> 2. Adding your custom templates as templates for that task.
> 3. Using unitxt production api for generating the final data with the required template to the model.

To add more details - there is now a way to use unitxt cards and templates to format input in LLM as judges.

See this documentation:

https://unitxt.readthedocs.io/en/latest/docs/production.html

This is how we plan to handle (work by @eladv / @OfirArviv )

elronbandel (approved):
After discussion, we will deal with my comments later, for now it looks good.

Issue #764, opened 2024-04-11
Fusion does not assign value to field "group" of every instance resulting in errors in metrics computation


elronbandel:
Fusion classes were suppose to add field named "group" to every instance of the fusion streams, stating the name of its origin. 
In turn in metric computation time the metric pipeline is splitting by groups (done here: https://github.com/IBM/unitxt/blob/4bdefa7fdf3a801fd214e07c9c89ec384fbe66b4/src/unitxt/metric_utils.py#L110) and calculate the metrics for each group seperatly based on the group metric. 

This requires two changes: 
1. changing the `origins` field in the Fusion classes to dictionary mapping between origin names to origin stream. 
2. Adding the name of the origin to every instance like here:

```python 
class FixedFusion(BaseFusion):
    """FixedFusion operator that combines multiple streams into one based on a fixed number of examples per task.

    Args:
        origins: List of SourceOperator objects.
        examples_per_task: Number of examples per task. If None, all examples are returned.
        splits: List of splits to include. If None, all splits are included.
    """

    max_instances_per_origin: Optional[int] = None

    def fusion_generator(self, split) -> Generator:
        for origin_name, origin in self.origins.items():
            multi_stream = origin()
            if split not in multi_stream:
                continue
            iterator = iter(multi_stream[split])
            if self.max_instances_per_origin is not None:
                for _ in range(self.max_instances_per_origin):
                    try:
                        instance = next(iterator)
                        instance["group"] = origin_name
                        yield instance
                    except StopIteration:
                        break
            else:
                for instance in iterator:
                    instance["group"] = origin_name
                    yield instance
```

Then ofcourse we need to check it all works with tests.

dafnapension:
Resolved by PR #767

PR #767, opened 2024-04-13
fuse, split, MultiStreamScoreMean, and merge all

dafnapension:
of each instance 
per issue #764

### DIFF
--- a/src/unitxt/fusion.py
+++ b/src/unitxt/fusion.py
@@ -1,30 +1,43 @@
-import copy
 from abc import abstractmethod
-from typing import Generator, List, Optional
+from typing import Dict, Generator, List, Optional, Union
 
 from .dataclass import NonPositionalField
 from .operator import SourceOperator
 from .random_utils import new_random_generator
 from .stream import MultiStream, Stream
+from .type_utils import isoftype
 
 
 class BaseFusion(SourceOperator):
     """BaseFusion operator that combines multiple streams into one.
 
     Args:
-        include_splits: List of splits to include. If None, all splits are included.
+        origins: a dict of named SourceOperators, or a list of such sources
+        include_splits: List of splits to include from each SourceOperator.
+                If None, all splits are included.
     """
 
-    origins: List[SourceOperator]
+    origins: Union[List[SourceOperator], Dict[str, SourceOperator]]
     include_splits: Optional[List[str]] = NonPositionalField(default=None)
 
     @abstractmethod
     def fusion_generator(self, split) -> Generator:
         pass
 
-    def splits(self) -> Generator:
+    def prepare(self):
+        super().prepare()
+        assert isoftype(self.origins, Dict[str, SourceOperator]) or isoftype(
+            self.origins, List[SourceOperator]
+        )
+        self.named_origins = (
+            {i: self.origins[i] for i in range(len(self.origins))}
+            if isinstance(self.origins, list)
+            else self.origins
+        )
+
+    def splits(self) -> List[str]:
         splits = []
-        for origin in self.origins:
+        for _, origin in self.named_origins.items():
             for s in origin().keys():
                 if s not in splits:
                     if self.include_splits is None or s in self.include_splits:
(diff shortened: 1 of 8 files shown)


src/unitxt/fusion.py
```diff
@@ -12,19 +11,21 @@ class BaseFusion(SourceOperator):
...
 
     Args:
-        include_splits: List of splits to include. If None, all splits are included.
+        origins: a dict of named SourceOperators
+        include_splits: List of splits to include from each SourceOperator.
+                If None, all splits are included.
     """
 
-    origins: List[SourceOperator]
+    origins: Dict[str, SourceOperator]
```
elronbandel, line 19:
```suggestion
    origins: Union[List[SourceOperato], Dict[str, SourceOperator]]
```
I think we should allow also anonymus sources merging which are merged without given name per source

src/unitxt/fusion.py
```diff
@@ -44,40 +45,59 @@ class FixedFusion(BaseFusion):
...
-                        yield next(iterator)
+                        instance = next(iterator)
+                        if "group" in instance:
+                            if "past_groups" in instance:
+                                instance["past_groups"] = (
+                                    instance["group"] + "/" + instance["past_groups"]
+                                )
+                            else:
+                                instance["past_groups"] = instance["group"]
+                        instance["group"] = origin_name
```
elronbandel, line 73:
```suggestion
                            group = origin_name + "/" + instance["group"]
                        else:
                            group = origin_name 
                        instance["group"] = group
```
Why not like this? more simple and dont require additional variable

src/unitxt/fusion.py
```diff
@@ -44,40 +45,59 @@ class FixedFusion(BaseFusion):
...
-                yield from iterator
+                for instance in iterator:
+                    if "group" in instance:
+                        if "past_groups" in instance:
+                            instance["past_groups"] = (
+                                instance["group"] + "/" + instance["past_groups"]
+                            )
+                        else:
+                            instance["past_groups"] = instance["group"]
+                    instance["group"] = origin_name
```
elronbandel, line 86: Same here dont think past_groups are necessary

elronbandel (changes requested):
I added some comments, in general:
1. I think we should have only one field "group" containing all the grouping information.
2. I think we should test also hirarchial group naming works withing the unittests
3. I think we should allow also Fusion of streams without names 
4. We should test the tests actually reutrn score per group with the existing mecahinsm (i linked the exact line in the metric pipeline in the related issue)

src/unitxt/fusion.py
dafnapension: I thought that code that reads the "group" (e.g, the splitter) will not need to check if any past is glued to this group, and if glued -- peel it off.   Only if the consuming code is interested in the past, it will look into "past_groups", and do whatever it wants.

src/unitxt/fusion.py
dafnapension: And in such a case, nothing will be written into field "group" ?  or I will generate juicy names like "0", "1", "2" , and store them into field "group"?

src/unitxt/fusion.py
dafnapension: Yes, same thing: trying to work here, in order to save work from the consure (e.g. Splitter)

dafnapension:
> I added some comments, in general:
> 
> 1. I think we should have only one field "group" containing all the grouping information.
> 2. I think we should test also hirarchial group naming works withing the unittests
> 3. I think we should allow also Fusion of streams without names
> 4. We should test the tests actually reutrn score per group with the existing mecahinsm (i linked the exact line in the metric pipeline in the related issue)

OK. I just thought to save work from the consuming code, but accept your judgement, of course.

src/unitxt/fusion.py
dafnapension: Changed per your request, @elronbandel. We will know what to do if the past haunts us..

src/unitxt/fusion.py
dafnapension: Changed here too

src/unitxt/operators.py
```diff
@@ -1560,6 +1560,48 @@ def process(self, multi_stream: MultiStream) -> MultiStream:
         return MultiStream(result)
 
 
+class SplitByGroup(MultiStreamOperator):
```
elronbandel, line 1563:
```suggestion
class SplitByNestedGroup(MultiStreamOperator):
     field: str = "group"
```

dafnapension:
Hi @elronbandel ,  I hope this is more in the direction you had in mind.

src/unitxt/fusion.py
```diff
@@ -41,43 +54,55 @@ def process(
...
-    """FixedFusion operator that combines multiple streams into one based on a fixed number of examples per task.
+    """FixedFusion operator that combines multiple multistreams into one, limiting the number of instances taken from each split of each input multistream.
 
     Args:
-        origins: List of SourceOperator objects.
-        examples_per_task: Number of examples per task. If None, all examples are returned.
-        splits: List of splits to include. If None, all splits are included.
+        origins: Dict of named SourceOperator objects (each to yield a MultiStream), or a list thereof
+        splits: List of splits (stream_names) to include, over all input multistreams. If None, all splits are included.
+        max_instances_per_origin_split: Number of instances to take from each input split of each input multistream.
```
elronbandel, line 62: Great catch! the old name was disambgious!

elronbandel (approved):

PR #769, opened 2024-04-14
Update formats 

elronbandel:


prepare/formats/models/llama2.py
```diff
@@ -9,7 +9,7 @@
 # {{ user_msg_1 }} [/INST] {{ model_answer_1 }} </s><s>[INST] {{ user_msg_2 }} [/INST]
 format = SystemFormat(
     demo_format="{source} [/INST] {target_prefix}{target} </s><s>[INST] ",
-    model_input_format="[INST] <<SYS>>\n{system_prompt}\n\n{instruction}<<SYS>>\n\n\n{demos}{source} [/INST]{target_prefix}",
```
yoavkatz, line 12: What about <s> before [INST]. Is it not needed?

eladven (approved):

PR #772, opened 2024-04-15
simplify bitwiseor-to-union and show a scheme for Literal

dafnapension:


pawelknes (changes requested):
Like this approach

Though I found an issue with union at the top level when there is some more complex type on the right like below:

`convert_union_type("int|List[int]")` produces `int|List[int]`, while `convert_union_type("List[int]|int")` correctly returns `Union[List[int],int]`

And similarly the below properly transforms only inner unions:

`convert_union_type("List[int|float]|Tuple[int|float]")` -> `"List[Union[int,float]]|Tuple[Union[int,float]]"`

dafnapension:
Thank you, @pawelknes , that is a very nice catch!!  I fixed and pushed, and added these cases as test cases in test_type_utils
Please let your sharp eye find more..

pawelknes (approved):

PR #774, opened 2024-04-17
Update few-shot instructions in Unitxt with HELM

yifanmai:


yifanmai:
@elronbandel Would you know why the Explore Unitxt tool always produces `max_train_instances=5` as part of the recipe? This breaks Unitxt with HELM because it triggers the `max_train_instances` run expander to set `max_train_instances` to 5 in the `AdapterSpec`, but because we _always_ want `max_train_instances` to be 0 for Unitxt.

yoavkatz:
This is probably used to speed up the process.  We will try to remove it from the output.

elronbandel (approved):

Issue #778, opened 2024-04-18
test_card fails when preprocess_steps is None


marukaz:
After I updated unitxt from 1.7.6 to 1.7.7, if there is no `preprocess_steps`, `test_card()` fails with

```
  File "/Users/marukaz/ghq/github.com/marukaz/unitxt/src/unitxt/standard.py", line 293, in prepare
    super().prepare()
  File "/Users/marukaz/ghq/github.com/marukaz/unitxt/src/unitxt/standard.py", line 215, in prepare
    self.standardization.steps.extend(self.card.preprocess_steps)
TypeError: 'NoneType' object is not iterable
```

Example card:

```
from unitxt.blocks import (
    AddFields,
    LoadHF,
    TaskCard,
)
from unitxt.catalog import add_to_catalog
from unitxt.test_utils.card import test_card

card = TaskCard(
    loader=LoadHF(path="EdinburghNLP/xsum"),
    # preprocess_steps=[
    #     AddFields(fields={"document_type": "document"}),
    # ],
    task="tasks.summarization.abstractive",
    templates="templates.summarization.abstractive.all",
)

test_card(card)
add_to_catalog(card, "cards.xsum", overwrite=True)
```

The cause is `None` check has been removed in this PR https://github.com/IBM/unitxt/pull/739/files#diff-a4122a708bc5f51b81dc0316fb8034bff822104b0e9d82bbe30d44761979ec36L144

I think either the code for the None check should be reinstated, or the default for `preprocess_steps` should be changed from `None` to an empty list.

PR #780, opened 2024-04-18
Set empty list if preprocess_steps is None

marukaz:
Fix https://github.com/IBM/unitxt/issues/778

I considered three options to solve the problem:

1. reinstate None check
2. Change default of preprocess_steps to `[]`
3. Assign `[]` in post init process if it's None

I chose 3 because

- if we do 1, same check will be needed whenever we add code for preprocess steps
- Regarding 2, as is well known, Python's default arguments are static and shared by all instances. Although this problem does not often cause confusion with Card, I would like to avoid it just in case


marukaz:
I tried to solve `src/unitxt/standard.py:189:9: C901 prepare is too complex (11 > 10)` but I felt the code is reasonable, so added this file to ignore list.

PR #782, opened 2024-04-22
Add open and ibm_genai to llm as judge inference engine 

OfirArviv:


src/unitxt/inference.py
```diff
@@ -31,3 +41,117 @@ def infer(self, dataset):
...
+    temperature: Optional[float] = None
+    top_p: Optional[float] = None
+
+
+class OpenAiInferenceEngine(InferenceEngine):
+    label: str = "openai"
+    model_name: str
+    parameters: OpenAiInferenceEngineParams = OpenAiInferenceEngineParams()
+
+    def prepare(self):
```
yoavkatz, line 115: Let's use PackageRequirementsMixin to manage imports, similar to how it's done in operators.

class PackageRequirementsMixin(Artifact):
    """Base class used to automatically check for the existence of required python dependencies for an artifact (e.g. Operator or Metric).

    The _requirement list is either a list of required packages
    (e.g. ["torch","sentence_transformers"]) or a dictionary between required packages
    and detailed installation instructions on how how to install each package.
    (e.g. {"torch" : "Install Torch using `pip install torch`", "sentence_transformers" : Install Sentence Transformers using `pip install sentence-transformers`})
    Note that the package names should be specified as they are used in the python import statement for the package.
    """

yoavkatz (commented):
Looks good.   One comment about package dependencies.  Also, we need to add some test (with Mock for this).

OfirArviv:
> Looks good. One comment about package dependencies. Also, we need to add some test (with Mock for this).

didn't Elad already wrote some? I do also see a test for the HF pipeline inference engine.
For BAM and OpenAi I don't think we need one. It's external api

yoavkatz (approved):

PR #783, opened 2024-04-23
Update instructions to install HELM from PyPI

yifanmai:
The latest HELM release has the Unitxt integration included, so we can now install from PyPI instead of GitHub.


yifanmai:
Test failure may be due to the Hugging Face outage earlier today.

elronbandel (approved):

Issue #699, opened 2024-03-21
Add new load from dictionary


yoavkatz:
For debugging and regressio  and  special cases of very small datasets it would good to have a way to initialize a stream from constants.

LoadFromConstants(data = {" test" : [ {"input" : "My text1", "output" : "Result1"} ,  {"input" : "My text2", "output" : "Result2"})

PR #784, opened 2024-04-25
Support for loading dictionaries

pawelknes:
Added loader which can be used to load dictionaries as per #699

### DIFF
--- a/src/unitxt/loaders.py
+++ b/src/unitxt/loaders.py
@@ -476,7 +476,23 @@ class MultipleSourceLoader(Loader):
         ).process()
 
 
-class LoadDirectory(Loader):
+class LoadFromDictionary(Loader):
+    """Allows loading data from dictionary of constants.
+
+    The loader can be used, for example, when debugging or working with small datasets.
+
+    Attributes:
+        data (Dict[str, List[Dict[str, Any]]]): a dictionary of constants from which the data will be loaded
+
+    Examples:
+        data = {
+            "train": {"input": "SomeInput1", "output": "SomeResult1"},
+            "test": {"input": "SomeInput2", "output": "SomeResult2"},
+        }
+        loader = LoadFromDictionary(data=data)
+        multi_stream = loader.process()
+    """
+
     data: Dict[str, List[Dict[str, Any]]]
 
     @staticmethod
--- a/src/unitxt/loaders.py
+++ b/src/unitxt/loaders.py
@@ -495,16 +495,5 @@ class LoadFromDictionary(Loader):
 
     data: Dict[str, List[Dict[str, Any]]]
 
-    @staticmethod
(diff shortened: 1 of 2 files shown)


src/unitxt/loaders.py
```diff
@@ -474,3 +474,21 @@ def process(self):
         return FixedFusion(
             origins=self.sources, max_instances_per_origin=self.get_limit()
         ).process()
+
+
+class LoadDirectory(Loader):
```
yoavkatz, line 479: I think it should be either LoadDictionary or LoadFromConstants.

Please also add a documentation string with example of how to use it, so will be appear in the auto generated documentation.

tests/library/test_loaders.py
```diff
@@ -226,3 +232,20 @@ def test_multiple_source_loader(self):
...
+                {"input": "Input2", "output": "Result2"},
+            ],
+            "test": [
+                {"input": "Input3", "output": "Result3"},
+            ],
+        }
+        loader = LoadDirectory(data=data)
+        streams = loader.process()
+
+        for data_type, instances in data.items():
```
yoavkatz, line 249:
```suggestion
        for split, instances in data.items():
```

tests/library/test_loaders.py
```diff
@@ -226,3 +232,20 @@ def test_multiple_source_loader(self):
...
+            for instance, stream in zip(instances, streams[data_type]):
```
yoavkatz, line 250:
```suggestion
            for original_instance, stream_instance in zip(instances, streams[data_type]):
```

yoavkatz (changes requested):
Thanks. Very useful class. I added a few comments.

src/unitxt/loaders.py
```diff
@@ -474,3 +474,37 @@ def process(self):
...
+    """
+
+    data: Dict[str, List[Dict[str, Any]]]
+
+    @staticmethod
+    def yield_instance(instances: List[Dict[str, Any]]) -> Dict[str, Any]:
+        yield from instances
+
+    def process(self) -> MultiStream:
+        return MultiStream(
```
dafnapension, line 503: suggesting:
return MultiStream.from_iterables(self.data)

pawelknes: Good point, thanks

tests/library/test_loaders.py
```diff
@@ -226,3 +232,20 @@ def test_multiple_source_loader(self):
             )
             ms = loader()
             assert len(dfs["test"]) + len(dfs["train"]) == len(list(ms["test"]))
+
+    def test_load_from_directory(self):
```
yoavkatz, line 236:
```suggestion
    def test_load_from_dictionary(self):
```

yoavkatz (approved):
Looks good thanks.

PR #788, opened 2024-05-02
add neg class score for binary precision, recall, f1 and max f1

lilacheden:
only new commits are the one starting from "add neg class score for binary precision, recall, f1 and max f1"
rest are already merged


src/unitxt/metrics.py
```diff
@@ -1524,7 +1529,7 @@ class Spearmanr(HuggingfaceMetric):
     hf_metric_name = "spearmanr"
     main_score = "spearmanr"
     process_single_instances = False
-    prediction_type = "float"
+    prediction_type = "Union[float, str]"
```
yoavkatz, line 1532: Why should Spearman metric accept a string?
Any string to float conversion should be done in a post processor  in the template.

lilacheden: All the metrics we use currently accept strings. We used the metric pipeline that does the conversion to float but it was removed in a recent commit by someone else. 
The best way to go is indeed to change all these metrics to accept floats and change the shared postprocessors accordingly - but this requires to change the api of other metrics in a non backward compatible way. Is that ok?
The relevant metrics: 
```
        "metrics.kendalltau_b",
        "metrics.roc_auc",
        "metrics.f1_binary",
        "metrics.accuracy_binary",
        "metrics.precision_binary",
        "metrics.recall_binary",
        "metrics.max_f1_binary",
        "metrics.max_accuracy_binary",
```

src/unitxt/metrics.py
```diff
@@ -1266,11 +1266,12 @@ class F1Binary(F1):
 
     process_single_instances = False
     main_score = "f1_binary"
-    average = "binary"
+    average = None
     pos_classes = {"1", "1.0", "yes", "true"}
```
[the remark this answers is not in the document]
lilacheden: We can just follow the float input api from the previous comments and have it receive 0/1, but again - not backward compatible

src/unitxt/metrics.py
yoavkatz: This is the right thing to do.  I think most of these were added by you for kpa.   We will fix any remaining issues.

lilacheden: Ok, changed the prediction_type to of the *_binary metrics to Union[float, int], since for binary labels it makes since to allow sending 0/1).

src/unitxt/metrics.py
```diff
@@ -2250,7 +2266,12 @@ def __init__(self, model_name, single_token_mode):
             from transformers import AutoTokenizer
 
             self.model_name = model_name
-            self.device = "cuda:0" if torch.cuda.is_available() else "cpu"
+            if torch.cuda.is_available():
+                self.device = torch.device("cuda:0")
+            elif torch.backends.mps.is_available():
+                self.device = torch.device("mps")
```
lilacheden, line 2272: @yoavkatz - is there a problem with adding MPS here?

yoavkatz (approved):

PR #792, opened 2024-05-05
Support prediction regex match by setting the operator as a postproce…

antonpibm:
This PR adds the ability to extract parts of the predictions with regex and compare them with a reference
The regex extraction only applies for the prediction
The to_list_by_comma only happens for the refernce
Adding the Jaccard index for comparing the two groups

After the regex processor is added it can be configured, example: `processors.prediction_regex_parser[regex=\d+]`


elronbandel (approved):
Great contribution @antonpibm !

Issue #646, opened 2024-03-11
Handling sensitive data sent to remote services


yoavkatz:
With the introductions of metrics that can send data to remote services - one needs a safe way to avoid accidentally sending propriety/confidential data to external services.

 In the common case in unitxt  metrics and datasets are developed by different people, who may not be aware of each other and the implementations, this becomes extremely error prone.

To address this we need a way for 
1) the dataset owner to specify for each dataset (or instance) the data classification.  The taxonomy should be defined by the user.
2) the metric owner should define whether the metric is safe for all data (e.g. running locally) or allow the user of the metric to specify which data classification are allowed to be used in the metric.

Suggested approach:
Each loader, will have an additional list[str] parameter called 'data_classification' .  Different loaders can have difficult default.  For example, LoadHF can be set the default to "public", while another Loader can set it to "propriety" .  The user can override these for specific datasets , e.g. "PII".

```
loader=LoadFromIBMCloud(
        endpoint_url_env="MY_COS_URL",
        aws_access_key_id_env=MY_COS_ACCESS_KEY_ID",
        aws_secret_access_key_env="MY_COS_SECRET_ACCESS_KEY",
        bucket_name="...",
        data_dir=....",
        data_files=["train.jsonl", "test.jsonl"],
        data_classification=["propriety","pii"]
    ),
```
The loaders will add the list as a field to all the instances in the loaded datasets.

Each base metric class will check in the compute() function that all instance data classifications are allowed by 
 check_allowed_data_classification(instance) .
 
 The default implementation of check_allowed_data_classification, will check a metric specific environment variable, for the list of allowed data classification.
 
   If not, the an error message of this type will be generated.

"The following instance  has  data classification of '{instance_data_classification}',   however the {metric} is only configured to support the following data with classification '{allowed_data_classification}.'  To allow, this set the enviromment variable {env_var} to  include  '{instance_data_classification}',"

 
 
 
@elronbandel @eladven @perlitz - Please review.

elronbandel:
I agree @yoavkatz . This is a good solution.

PR #806, opened 2024-05-08
Support for handling sensitive data sent to remote services

pawelknes:
Added modifications to loaders and the `RemoteMetric` class so that a metric can check if it's allowed to handle data which is sent to a remote service as per #646


elronbandel (changes requested):
I think it might simplify things if you use AddFields operator to add the classification:
```python 
self.add_data_classification = AddFields({"data_classification": self.data_classification})
multi_stream = MultiStream.from_iterables(dataset)
return self.add_data_classification(multi_stream)
```
I think this approach is also more plausible from the prespective of code responsibility seperation. 

Also do we want to have some sort of hierarchy of data classification types?

pawelknes:
> I think it might simplify things if you use AddFields operator to add the classification:
> 
> ```python
> self.add_data_classification = AddFields({"data_classification": self.data_classification})
> multi_stream = MultiStream.from_iterables(dataset)
> return self.add_data_classification(multi_stream)
> ```
> 
> I think this approach is also more plausible from the prespective of code responsibility seperation.
> 
> Also do we want to have some sort of hierarchy of data classification types?

Thanks for the suggestion

@yoavkatz what about this hierarchy?

src/unitxt/loaders.py
```diff
@@ -198,7 +205,9 @@ def process(self):
         if self.get_limit() is not None:
             return self.limited_load()
 
-        return MultiStream.from_iterables(dataset)
+        multi_stream = MultiStream.from_iterables(dataset)
+
+        return self.add_data_classification(multi_stream)
```
yoavkatz, line 210: I think that instead of having every loader add this functionality, you can add create a common process() method that would return self.add_data_classification(self.load())  and each loader would loading code will be in load() and not process().  

It wil also allow us to added load_limit support in the load() method.

pawelknes: I used the name `load_data` since `load` is already a method of the `Artifact` class from which loaders indirectly inherit

pawelknes:
> Hi @pawelknes . Thanks for the PR. It looks very good.
> 
> Since the issue started, we saw the need for this expand beyond metrics. For example, a person added a Translate post processor that sends the predicted data to IBM Cloud translation service. There is also a new InferenceEngine class that is sends data to 3rd party services (used by LLM as judges metric)
> 
> So perhaps instead of making metric centric, we should make it more generic (applicable for all Artifacts).
> 
> Similar to how class ComputeExpressionMixin(Artifact) adds ability to compute expressions.
> 
> What do you think?

Sure, it sounds good and I can work on this. Though we want to add this in the current PR, or firstly merge it and then create a new issue for expanding it to a general case?

yoavkatz:
> > Hi @pawelknes . Thanks for the PR. It looks very good.
> > Since the issue started, we saw the need for this expand beyond metrics. For example, a person added a Translate post processor that sends the predicted data to IBM Cloud translation service. There is also a new InferenceEngine class that is sends data to 3rd party services (used by LLM as judges metric)
> > So perhaps instead of making metric centric, we should make it more generic (applicable for all Artifacts).
> > Similar to how class ComputeExpressionMixin(Artifact) adds ability to compute expressions.
> > What do you think?
> 
> Sure, it sounds good and I can work on this. Though we want to add this in the current PR, or firstly merge it and then create a new issue for expanding it to a general case?

I tend to think that we should do it now, because in anywise, we will need to document and instruct the users. We probably dont' want to do it twice.

src/unitxt/metric_utils.py
```diff
@@ -248,6 +248,13 @@ class MetricResponse(Dataclass):
 # For example, 'http://127.0.0.1:8000/compute'
 UNITXT_REMOTE_METRICS_ENDPOINT = "UNITXT_REMOTE_METRICS_ENDPOINT"
 
+# Classification of the data sent to remote services in terms of its sensitivity.
+# It determines, for each specified metric, what data can be used.
+# For example: '{"metrics.rag.context_relevance": ["public"]}'.
+# This variable should be a dictionary, where each key is a given metric's name,
+# and a value is a list of data classifications used by that metric.
+UNITXT_REMOTE_METRICS_DATA_CLASSIFICATION = "UNITXT_REMOTE_METRICS_DATA_CLASSIFICATION"
```
yoavkatz, line 256: We could call this something like "UNITXT_DATA_CLASSSIFICATION_POLICY"

{"metrics.rag.mt_bench.using.open-ai" : ["public"] , "postprocessor.translate" : ["proprietry"]}

pawelknes:
> > > Hi @pawelknes . Thanks for the PR. It looks very good.
> > > Since the issue started, we saw the need for this expand beyond metrics. For example, a person added a Translate post processor that sends the predicted data to IBM Cloud translation service. There is also a new InferenceEngine class that is sends data to 3rd party services (used by LLM as judges metric)
> > > So perhaps instead of making metric centric, we should make it more generic (applicable for all Artifacts).
> > > Similar to how class ComputeExpressionMixin(Artifact) adds ability to compute expressions.
> > > What do you think?
> > 
> > 
> > Sure, it sounds good and I can work on this. Though we want to add this in the current PR, or firstly merge it and then create a new issue for expanding it to a general case?
> 
> I tend to think that we should do it now, because in anycase, we will need to document it in details and instruct the users. We probably don't want to do it twice. Looking the code, it may not be a huge change. More of renaming and moving the code from metrics to artifact.

Okay, so I'll start working on it

src/unitxt/metrics.py
```diff
@@ -2725,16 +2725,58 @@ class RemoteMetric(SingleStreamOperator, Metric):
...
     main_score: str = None
     endpoint: str
     metric_name: str
     api_key: str = None
+    allowed_data_classification: List[str] = None
+
+    def check_allowed_data_classification(
+        self, instances: List[Dict[str, Any]]
+    ) -> None:
+        if not self.allowed_data_classification:
```
yoavkatz, line 2740: Not that data classification is required not only for RemoteMetric (which are standard metrics which are executed in a remote docker container, but also by metrics that call remote APIs directly).   If we move this behavior to DataClassificationMixIn, it will allow us to add it to all metrics that need it.

pawelknes:
@yoavkatz I've moved functionalities so that they can be used by any artifact. As for the operator responsible for verifications, I made it accepting either an artifact name (and read a respective policy from env) or a Python object (`Accuracy` etc.) and read a policy from the attribute

tests/library/test_artifact.py
```diff
@@ -130,3 +137,74 @@ def test_modifying_fetched_artifact_does_not_effect_cached_artifacts(self):
...
+            f"'{artifact_name}' object to include '{instance['data_classification_policy']}'.",
+        )
+
+    def test_checking_data_classification_policy_object(self):
+        instance = {"data": "text", "data_classification_policy": ["public"]}
+
+        metric = F1Binary(data_classification_policy=["public"])
+        metric_verification = VerifyDataClassificationPolicy(artifact=metric)
+
+        output = metric_verification.process(instance)
```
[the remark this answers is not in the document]
pawelknes: Is the current form what we want it to be? Also, is the added documentation okay enough?

docs/docs/data_classification_policy.rst
```diff
@@ -0,0 +1,93 @@
...
+
+.. note::
+
+   To use this tutorial, you need to :ref:`install unitxt <install_unitxt>`.
+
+=====================================
+Sensitive data in unitxt ✨
+=====================================
+
+The section discusses how to properly handle sensitive data in unitxt in order to avoid accidentally sending propriety/confidential data to external services.
```
yoavkatz, line 11:
```suggestion
The section discusses how to properly handle sensitive data in unitxt in order to avoid accidentally sending proprietary/confidential data to external services.
```

docs/docs/data_classification_policy.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Data classification policy
+----------------------------
+Whenever processing streams in unitxt, users can specify a parameter called `data_classification_policy`, which determines how this data should be treated. The taxonomy is defined by a user and may encompass multiple different policies.
+
+Each component used in unitxt (metrics, operators etc.) has the same parameter as well. This allows a given component to verify if it can process instances of a stream.
```
yoavkatz, line 17:
```suggestion
Each component used in Unitxt (metrics, operators, inference engines etc.) has the same parameter as well. This allows a given component to verify if it can process instances of a stream.
```

src/unitxt/artifact.py
```diff
@@ -259,6 +276,84 @@ def save(self, path):
...
+            UNITXT_DATA_CLASSIFICATION_POLICY = json.dumps({"metrics.accuracy": ["pii"]})
+            metric = fetch_artifact("metrics.accuracy")
+            metric.verify_instance(instance)
+        """
+        name = name or self.__id__ or self.__class__.__name__
+        data_classification_policy = get_artifacts_data_classification(name)
+        if not data_classification_policy:
+            data_classification_policy = self.data_classification_policy
+
+        if not data_classification_policy:
```
yoavkatz, line 319: Perhaps we should not generate a warning?  Most operators can work on all data.

pawelknes: Yup, good point because it'd spam a lot of unnecessary warnings

pawelknes: I left a warning when no `data_classification_policy` of an instance is given tough, as an operator won't be able to verify if it should process that instance (and it'll have a policy in such a case)

yoavkatz: Yes.  Perhaps we should call the instance field "data_classification"?  This is because for the data it's not a policy of what to allow - it's just what the data classification  is.  It will also better differentiate between artifact and the data.

yoavkatz: Wait.   I think it can actually cause more confusion.

I'm doing some changes and testing it end to end .

pawelknes: Since all loaders indirectly inherit from `Artifact`, they'll have the `data_classification_policy` attribute anyway. So one solution could be to add an additional attribute `data_classification` just for loaders, and then when the method `add_data_classification` is called, add a field `data_classification` to all instances based on value present in either `data_classification` (priority) or `data_classification_policy`. But again this may be a little confusing so should described in detail somewhere

yoavkatz:
Pawel - thanks for this PR. It's an important one, especially since we have LLM as judges.
I've made some additions and changes to the PR - I'll be glad for your review.

1. I added default data classifications policies to loaders, so users will not be required to add them always.
2. Reviewed and added more  documentation
3. Made the use of the environment variable more consistent with other place
4. Made the default policy 'None' in artifact so we know if the user explicitly set it , or not.

src/unitxt/artifact.py
yoavkatz: I agree. It is more confusing.  So I dismissed this direction.

yoavkatz:
@pawelknes - I think the next steps are:

1. Remove allow_passing_data_to_remote_api from settings.

2. In class InferenceEngine(abc.ABC, Artifact): Make sure infer() checks data classification of the inferred instance
(best to do it in the base method, so people writing inference engines won't need to handle lit).  Make BAM and WML handle 'public' and 'proprietary' data by default. 

3.  LlamaIndexCorrectness(InstanceMetric) - make it process only 'public' data by default.

4. class RemoteMetric(SingleStreamOperator, Metric) - make it handle 'public' and 'proprietary' data by default.

pawelknes:
> @pawelknes - I think the next steps are:
> 
> 1. Remove allow_passing_data_to_remote_api from settings.
> 2. In class InferenceEngine(abc.ABC, Artifact): Make sure infer() checks data classification of the inferred instance
>    (best to do it in the base method, so people writing inference engines won't need to handle lit).  Make BAM and WML handle 'public' and 'proprietary' data by default.
> 3. LlamaIndexCorrectness(InstanceMetric) - make it process only 'public' data by default.
> 4. class RemoteMetric(SingleStreamOperator, Metric) - make it handle 'public' and 'proprietary' data by default.

Yup, I'll make these changes

src/unitxt/inference.py
```diff
@@ -73,6 +73,7 @@ def prepare(self):
             model=self.model_name, trust_remote_code=True, **model_args
         )
 
+    @InferenceEngine.verify_dataset
```
yoavkatz, line 76: Do we prefer to use decorator? It requires the writer of the inference engine to remember to add the decorator.
Maybe better have infer() that calls verify_dataset()  and then _infer() which will be implemented by each inference engine.

pawelknes: Yup, it may be better

yoavkatz (approved):

PR #808, opened 2024-05-08
Added new NER metric using fuzzywuzzy logic

sarathsgvr:
Added a new metric using fuzzywuzzy logic for NER evaluation. This will help to compare the predictions with actual using string match based on predefined threshold.


tests/library/test_metrics.py
```diff
@@ -1432,3 +1433,35 @@ def test_llm_as_judge_metric(self):
...
+                ("Marathahalli ring road", "Location"),
+            ]
+        ]
+    ]
+    outputs = apply_metric(
+        metric=metric, predictions=predictions, references=references
+    )
+    global_target = 1.0
+    self.assertAlmostEqual(global_target, outputs[0]["score"]["global"]["f1_Person"])
+    global_target = 0.0
```
yoavkatz, line 1460: Would Ner() metric return a different result?

prepare/metrics/customf1fuzzy.py
```diff
@@ -0,0 +1,359 @@
...
+from unitxt.test_utils.metrics import test_metric
+
+metric = FuzzyNer()
+# Test1 single line single class
+# 0.1 simple case, multi examples
+predictions = [
+    [("Amir", "Person"), ("Yaron", "Person")],
+    [("Ran", "Person"), ("Yonatan", "Person")],
+]
+references = [[[("Yaron", "Person"), ("Ran", "Person")]], [[("Yonatan", "Person")]]]
```
yoavkatz, line 12: I think we need to modify the tests to see that there Fuzzy matching makes a difference compared to regular NER.

tests/library/test_metrics.py
rrajkumar1990: we have changed the predicton and reference inputs

prepare/metrics/customf1fuzzy.py
rrajkumar1990: we will update different data to check on the values

src/unitxt/metrics.py
```diff
@@ -3642,3 +3644,118 @@ class NormalizedSacrebleu(HuggingfaceMetric):
...
+                        max_score = score
+                        best_total_key = total_key
+
+            if best_total_key is not None:
+                tmp.append(min(actual_group[actual_key], total_group[best_total_key]))
+            else:
+                tmp.append(min(actual_group[actual_key], 0))
+        return sum(tmp), sum(actual_group.values())
+
+    def compute(
```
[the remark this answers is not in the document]
rrajkumar1990: Renamed calculate_groups_ratio_fuzzy to calculate_groups_ratio. And removed the compute method  accordingly to use the same base compute method

tests/library/test_metrics.py
yoavkatz: But NER would match "jar htaras" to "jar htaras" and not match  "Marathahalli" "Marathahalli ring road", exactly as the fuzzy match does.   I would expect the test to match   "Marathahalli" and "Marathahalli ring road".

src/unitxt/metrics.py
```diff
@@ -12,6 +13,7 @@
 import evaluate
 import numpy
 import numpy as np
+from fuzzywuzzy import fuzz
```
yoavkatz, line 16: Please import this where used and not in main scope - so if people don't use the metric, it will not fail.

rrajkumar1990:
> Hi. I think primary what's left is to update the test value, and move the import to where it is used. Is this right?

We have change the Test values and changed the import inside the custom method.

yoavkatz:
> > Hi. I think primary what's left is to update the test value, and move the import to where it is used. Is this right?
> 
> We have changed the Test values and changed the import inside the custom method.

I saw you moved the import.  But I don't see changes in the test values. Am I missing anything?

tests/library/test_metrics.py
yoavkatz: Tests added.

PR #811, opened 2024-05-09
Add RAG (response generation part) tasks and datasets

perlitz:


elronbandel (approved):
Great job looks really good!

PR #819, opened 2024-05-13
Handle empty text in Literal Eval

antonpibm:


src/unitxt/processors.py
```diff
@@ -237,4 +237,6 @@ def process_value(self, text: Any) -> Any:
 
 class LiteralEval(FieldOperator):
     def process_value(self, text: Any) -> Any:
+        if not text:
```
elronbandel, line 240: Can you make this more explicit? (`if text is None or text == "":`)
Also any chance you can raise `ValueError` with detailed message in case that `not is instance(text, str)`
And thanks againg for the contributions @antonpibm !

elronbandel (changes requested):

elronbandel (approved):

PR #822, opened 2024-05-15
Changed  default log verbosity to "info" instead of "debug"

yoavkatz:
Also made artifact checking a debug statement.  This is because it causes a lot for clutter on the screen.


elronbandel (approved):
Crazy we didn't do it earlier!

PR #832, opened 2024-05-15
Ner typing had a mistake.

yoavkatz:
'text' was set to different types in input and outputs.


elronbandel (approved):

elronbandel (approved):
Aren't we checking this in the tests? How did it slip through?

PR #842, opened 2024-05-16
Rename FormTask to Task

elronbandel:
Notice FromTask still will work for backward compatibility


yoavkatz (changes requested):
What about changes in the documentation?

Are there no more FormTasks ?

elronbandel:
> What about changes in the documentation?
> 
> Are there no more FormTasks ?

Done

PR #843, opened 2024-05-17
Eagerizing through ListStreams and GeneratorStreams

dafnapension:
Hi @elronbandel , I had this in mind re the issue you assigned to me.
Is this in the direction you had in mind?
If yes, I will clean the celebration by adding a function to standard.py, maybe, that sticks the filler in between the steps of a given BaseRecipe object.

### DIFF
--- a/src/unitxt/stream.py
+++ b/src/unitxt/stream.py
@@ -149,9 +149,7 @@ class DynamicStream(Stream):
                         )
                         traceback = traceback.tb_next
                     logger.info("End of Stack Trace")
-                    raise ValueError(
-                        f"Could not generate any instance from input generator, due to {e!s}"
-                    ) from e
+                    raise e
 
         if self.stream is None:
             self.stream = GeneratorStream(
--- a/src/unitxt/stream.py
+++ b/src/unitxt/stream.py
@@ -118,15 +118,15 @@ class DynamicStream(Stream):
                 )
 
     def __iter__(self):
-        self.def_self_stream_if_none()
+        self._verify_init()
         return self.stream.__iter__()
 
     def peek(self):
-        self.def_self_stream_if_none()
+        self._verify_init()
         return self.stream.peek()
 
     def take(self, n):
-        self.def_self_stream_if_none()
+        self._verify_init()
         return self.stream.take(n)
 
 
--- a/src/unitxt/stream.py
+++ b/src/unitxt/stream.py
@@ -100,26 +100,33 @@ class DynamicStream(Stream):
     copying: bool = False
 
     def __post_init__(self):
-        if settings.use_eager_execution:
-            instances_list = []
-            for instance in self.generator(**self.gen_kwargs):
-                instances_list.append(instance)
-            self.stream = ListStream(instances_list=instances_list)
-        else:
-            self.stream = GeneratorStream(
-                generator=self.generator,
-                gen_kwargs=self.gen_kwargs,
-                caching=self.caching,
-                copying=self.copying,
-            )
+        self.stream = None
+
+    def def_self_stream_if_none(self):
+        if self.stream is None:
+            if settings.use_eager_execution:
+                instances_list = []
+                for instance in self.generator(**self.gen_kwargs):
+                    instances_list.append(instance)
+                self.stream = ListStream(instances_list=instances_list)
+            else:
+                self.stream = GeneratorStream(
+                    generator=self.generator,
+                    gen_kwargs=self.gen_kwargs,
+                    caching=self.caching,
+                    copying=self.copying,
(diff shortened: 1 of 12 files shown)


elronbandel:
The way i thought about it is that every base operator `StreamInstanceOperator` `StreamOprator` `MultiStreamOperator` `FieldOperator` will have two opeartors they use: StreamingFieldOperator and EagerFieldOperator

so the __call__ of each will look like:

```python
def __call__(multi_stream):
    if settings.eager_mode:
         return EagerFieldOperator.__call__(multi_stream)
    else:
         return StreamingFieldOperator.__call__(multi_stream)

**Comment by dafnapension** [comment]
Thanks @elronbandel , do you see a way to only modify those four ancestor-operators? or need to change all their descendants: each and every operator in operators.py?

**Comment by dafnapension** [comment]
Hi @elronbandel , 
I add one commit here that makes `test_recipe` pass, except one case that I can not explain, where the `recipe_metadata` does not come the same twice in eager_mode. perhaps you can understand.. maybe the random generator (that is part of that field) is not the same object in same address?

File: src/unitxt/stream.py [hunk 1]
```diff
@@ -89,6 +93,43 @@ def take(self, n):
...
+class DynamicStream(Stream):
+    generator: Callable
+    gen_kwargs: Dict[str, Any] = OptionalField(default_factory=dict)
+    caching: bool = False
+    copying: bool = False
+
+    def __post_init__(self):
+        self.stream = None
+
+    def def_self_stream_if_none(self):
```
elronbandel, line 105:
```suggestion
    def verify_init(self):
```

src/unitxt/stream.py
```diff
@@ -89,6 +93,43 @@ def take(self, n):
...
+            else:
+                self.stream = GeneratorStream(
+                    generator=self.generator,
+                    gen_kwargs=self.gen_kwargs,
+                    caching=self.caching,
+                    copying=self.copying,
+                )
+
+    def __iter__(self):
+        self.def_self_stream_if_none()
```
elronbandel, line 121:
```suggestion
        self.verify_init()
```

src/unitxt/stream.py
```diff
@@ -89,6 +93,43 @@ def take(self, n):
...
+        return self.stream.__iter__()
+
+    def peek(self):
+        self.def_self_stream_if_none()
```
elronbandel, line 125:
```suggestion
        self.verify_init()
```

src/unitxt/stream.py
```diff
@@ -89,6 +93,43 @@ def take(self, n):
...
+        return self.stream.peek()
+
+    def take(self, n):
+        self.def_self_stream_if_none()
```
elronbandel, line 129:
```suggestion
        self.verify_init()
```

src/unitxt/stream.py
```diff
@@ -89,6 +93,41 @@ def take(self, n):
...
+    def __post_init__(self):
+        self.stream = None
+        if settings.use_eager_execution:
+            try:
+                instances_list = []
+                for instance in self.generator(**self.gen_kwargs):
+                    instances_list.append(instance)
+                self.stream = ListStream(instances_list=instances_list)
+            except:
+                # input stream is empty
```
elronbandel, line 111: This should except only that exception (If needed we can add EmptyStreamException)

elronbandel (changes requested):
I wish we could solve the problem from its root (making sure there is not code that create streams that are not supposed to be realized) for now i guess it is better than nothing except my comment.

src/unitxt/stream.py
dafnapension: A very good point, @elronbandel . Will dig into this.

dafnapension:
So I digged into the code, debugging from `test_recipe`.
All cases that reached, eventually, to this Exception that bothers you, @elronbandel , stemmed, indeed from "disappearing" streams.
Some of them disappear for years now, we simply did not look for them. I added a check, upon the return from each recipe() in `test_recipe()` that prints out all the streams it carries with it, and what is their status.  And sure enough - many streams are `missing in action..`.  The other source for the exception, are streams that recently began to disappear - with the new tests of Yoav, that pushes a new stream into an existing recipe, replacing the existing streams (that are thus gone). 
Here is a tweaked `test_recipe` that reveals all the classically disappearing streams:

[test_recipe.TXT](https://github.com/user-attachments/files/15538459/test_recipe.TXT)

dafnapension:
Hi @elronbandel , I hope this is closer to what you had in mind

dafnapension:
covered by the better version: https://github.com/IBM/unitxt/pull/888

PR #844, opened 2024-05-17
Ensure consistent evaluation of CI across implementations [Might change previous results]

dafnapension:
The new feature of Metric, sample-from-groups-scores, employs the CI over instances generated ad-hoc, one per group (a group is a subset of the input instances, whose member instances are those instances of the stream having a specific value in a specific field).  CI generates the samples through first generating a list of indices (into the input instances) and then dealing with the instances thus indexed. 
Hence, the result of CI for these instances generated ad-hoc depends on the order in which these instances are fed into CI.
This PR standardizes this order by the alphabetic order of the values that define the groups (as explained above)

### DIFF
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -921,7 +921,9 @@ class InstanceMetric(SingleStreamOperator, MetricWithConfidenceInterval):
                     }
                 }
             }
-            for group_name in sorted(group_to_instance_scores.keys()) # sorted for consistency 
+            for group_name in sorted(
+                group_to_instance_scores.keys()
+            )  # sorted for consistency
         ]
 
     def _set_up_group_mean_aggregation(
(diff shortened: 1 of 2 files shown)


src/unitxt/metrics.py
```diff
@@ -915,11 +915,13 @@ def get_group_scores(
...
                         )
-                        for score_name, score_dict in group_scores.items()
+                        for score_name, score_dict in group_to_instance_scores[
+                            group_name
+                        ].items()
                     }
                 }
             }
-            for group_scores in group_to_instance_scores.values()
+            for group_name in sorted(group_to_instance_scores.keys())
```
elronbandel, line 924:
```suggestion
            for group_name in sorted(group_to_instance_scores.keys()) # sorted for consistency 
```

elronbandel (approved):

PR #845, opened 2024-05-17
Simplified and thereby extended grouping, filtering, and splitting to control-comparison, over to global metrics and to bulk-instance metrics

dafnapension:
Suggest a scheme of "aggregator" to be similarly employed by all three types of MetricsWithConfidenceInterval.


dafnapension:
replaced by #890

PR #867, opened 2024-05-27
add MinimumOneExamplePerLabelRefiner

alonh:
to allow having at least one element per label when using a sub-sample of the train set
unit tests to show the expected behavior.

Signed-off-by: ALON HALFON <ALONHAL@il.ibm.com>

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1990,7 +1990,7 @@ class MinimumOneExamplePerLabelRefiner(StreamRefiner):
     """A class used to return a specified number instances ensuring atleast one example  per label.
 
     For each instance, a signature value is constructed from the values of the instance in specified input 'fields'.
-    By discarding instances from the input stream, MinimumOneExamplePerLabelRefiner maintains at least one
+    MinimumOneExamplePerLabelRefiner takes first instance that appears from each label (each unique signature), and then adds more elements up to the max_instances limit.  In general, the refiner takes the first elements in the stream that meet the required conditions.
     example per label. MinimumOneExamplePerLabelRefiner shuffles the results to avoid having one element
     from each class first and then the rest . If max instance is not set, the original stream will be used
 
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1987,7 +1987,7 @@ class DeterministicBalancer(StreamRefiner):
 
 
 class MinimumOneExamplePerLabelRefiner(StreamRefiner):
-    """A class used to return a specified number instances ensuring atleast one example  per label.
+    """A class used to return a specified number instances ensuring at least one example  per label.
 
     For each instance, a signature value is constructed from the values of the instance in specified input 'fields'.
     MinimumOneExamplePerLabelRefiner takes first instance that appears from each label (each unique signature), and then adds more elements up to the max_instances limit.  In general, the refiner takes the first elements in the stream that meet the required conditions.
@@ -1996,7 +1996,7 @@ class MinimumOneExamplePerLabelRefiner(StreamRefiner):
 
     Attributes:
         fields (List[str]): A list of field names to be used in producing the instance's signature.
-        max_instances (Optional, int): Number of elements to select. Note that max_instances of StreamRefiners that are passed to the recipe (e.g. 'train_refiner'. `test_refiner`) are overridden by the recipe parameters ( `max_train_instances`, `max_test_instances`) 
+        max_instances (Optional, int): Number of elements to select. Note that max_instances of StreamRefiners that are passed to the recipe (e.g. 'train_refiner'. `test_refiner`) are overridden by the recipe parameters ( `max_train_instances`, `max_test_instances`)
 
     Usage:
         balancer = MinimumOneExamplePerLabelRefiner(fields=["field1", "field2"], max_instances=200)
@@ -2004,9 +2004,9 @@ class MinimumOneExamplePerLabelRefiner(StreamRefiner):
 
     Example:
         When input [{"a": 1, "b": 1},{"a": 1, "b": 2},{"a": 1, "b": 3},{"a": 1, "b": 4},{"a": 2, "b": 5}] is fed into
-        MinimumOneExamplePerLabelRefiner(fields=["a"], max_instances=2)
+        MinimumOneExamplePerLabelRefiner(fields=["a"], max_instances=3)
         the resulting stream will be:
-        [{"a": 2, "b": 5}] + one of [{"a": 1, "b": 1},{"a": 1, "b": 2},{"a": 1, "b": 3},{"a": 1, "b": 4}]
+        [{'a': 1, 'b': 1}, {'a': 1, 'b': 2}, {'a': 2, 'b': 5}] (order may be different)
     """
 
     fields: List[str]
@@ -2015,8 +2015,9 @@ class MinimumOneExamplePerLabelRefiner(StreamRefiner):
         return str(tuple(dict_get(instance, field) for field in self.fields))
 
     def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generator:
-        for instance in stream:
-            yield instance
+        if self.max_instances is None:
+            for instance in stream:
+                yield instance
 
         counter = Counter()
         for instance in stream:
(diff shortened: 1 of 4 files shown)


src/unitxt/operators.py
```diff
@@ -1985,6 +1986,80 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
                 yield instance
 
 
+class MinimumOneExamplePerLabelRefiner(StreamRefiner):
+    """A class used to include at least one example per label.
+
+    For each instance, a signature is constructed from the values of the instance in specified input 'fields'.
```
yoavkatz, line 1992:
```suggestion
    For each instance, a signature value is constructed from the values of the instance in specified input 'fields'.
```

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,80 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+    By discarding instances from the input stream, MinimumOneExamplePerLabelRefiner maintains at least one
+    example per label. MinimumOneExamplePerLabelRefiner shuffles the results to avoid having one element
+    from each class first and then the rest . If max instance is not set, the original stream will be used
+
+    Attributes:
+        fields (List[str]): A list of field names to be used in producing the instance's signature.
+        max_instances (Optional, int)
```
yoavkatz, line 1999: I think this should be mandatory. What's the point of having this operator, if the entire stream is returned when max_instances=None

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,80 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+
+    Usage:
+        balancer = MinimumOneExamplePerLabelRefiner(fields=["field1", "field2"], max_instances=200)
+        balanced_stream = balancer.process(stream)
+
+    Example:
+        When input [{"a": 1, "b": 1},{"a": 1, "b": 2},{"a": 1},{"a": 1},{"a": 2}] is fed into
```
yoavkatz, line 2006: Bets to have all instructions have same fields (both "a", and "b" because this is the typical case in real cases)

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,80 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+
+        for instance in stream:
+            counter[self.signature(instance)] += 1
+        all_keys = counter.keys()
+        if len(counter) == 0:
+            return
+
+        if self.max_instances is not None and len(all_keys) > self.max_instances:
+            raise Exception(
+                f"max instances of {self.max_instances} is smaller than the number of different labels"
```
yoavkatz, line 2031:
```suggestion
                f"Can not generate a stream with at least one example per label, because the max instances requested  {self.max_instances} is smaller than the number of different labels {len(all_keys)}"
```

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,80 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
                 yield instance
 
 
+class MinimumOneExamplePerLabelRefiner(StreamRefiner):
+    """A class used to include at least one example per label.
+
+    For each instance, a signature is constructed from the values of the instance in specified input 'fields'.
+    By discarding instances from the input stream, MinimumOneExamplePerLabelRefiner maintains at least one
+    example per label. MinimumOneExamplePerLabelRefiner shuffles the results to avoid having one element
+    from each class first and then the rest . If max instance is not set, the original stream will be used
```
yoavkatz, line 1995: Need to mention the resulting stream is shuffled.

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,80 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+
+        counter = Counter()
+        used_indices = set()
+        selected_elements = []
+        # select at least one per class
+        for idx, instance in enumerate(stream):
+            sign = self.signature(instance)
+            if counter[sign] == 0:
+                counter[sign] += 1
+                used_indices.add(idx)
```
yoavkatz, line 2043: You always take the first example from each label.  Is this intended?

alonh: Yes, I didn't see an issue with that given that we shuffle after selection. Do you think it's a problem?

src/unitxt/operators.py
alonh: more than what I already wrote? 
``` 
...MinimumOneExamplePerLabelRefiner shuffles the results to avoid having one element from each class first and then the rest
```

src/unitxt/operators.py
alonh: I took this from another example, will fix.

src/unitxt/operators.py
alonh: 👍

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,79 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
                 yield instance
 
 
+class MinimumOneExamplePerLabelRefiner(StreamRefiner):
+    """A class used to include at least one example per label.
```
yoavkatz, line 1990:
```suggestion
    """A class used to return a specified number instances ensuring atleast one example  per label.
```

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,79 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+
+    For each instance, a signature value is constructed from the values of the instance in specified input 'fields'.
+    By discarding instances from the input stream, MinimumOneExamplePerLabelRefiner maintains at least one
```
yoavkatz, line 1993:
```suggestion
    MinimumOneExamplePerLabelRefiner takes first instance that appears from each label, and then adds more elements up to the max_instances limit.  
```

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,79 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+    example per label. MinimumOneExamplePerLabelRefiner shuffles the results to avoid having one element
+    from each class first and then the rest . If max instance is not set, the original stream will be used
+
+    Attributes:
+        fields (List[str]): A list of field names to be used in producing the instance's signature.
+        max_instances (Optional, int): Number of elements to select
```
yoavkatz, line 1999:
```suggestion
        max_instances (Optional, int): Number of elements to select. Note that max_instances of StreamRefiners that are passed to the recipe (e.g. 'train_refiner'. `test_refiner`) are overridden by the recipe parameters ( `max_train_instances`, `max_test_instances`) 
```

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,79 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
                 yield instance
 
 
+class MinimumOneExamplePerLabelRefiner(StreamRefiner):
+    """A class used to include at least one example per label.
+
+    For each instance, a signature value is constructed from the values of the instance in specified input 'fields'.
+    By discarding instances from the input stream, MinimumOneExamplePerLabelRefiner maintains at least one
+    example per label. MinimumOneExamplePerLabelRefiner shuffles the results to avoid having one element
```
yoavkatz, line 1994:
```suggestion
    MinimumOneExamplePerLabelRefiner then shuffles the results to avoid having one element
```

src/unitxt/operators.py
```diff
@@ -1985,6 +1986,79 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+
+    Usage:
+        balancer = MinimumOneExamplePerLabelRefiner(fields=["field1", "field2"], max_instances=200)
+        balanced_stream = balancer.process(stream)
+
+    Example:
+        When input [{"a": 1, "b": 1},{"a": 1, "b": 2},{"a": 1, "b": 3},{"a": 1, "b": 4},{"a": 2, "b": 5}] is fed into
+        MinimumOneExamplePerLabelRefiner(fields=["a"], max_instances=2)
+        the resulting stream will be:
+        [{"a": 2, "b": 5}] + one of [{"a": 1, "b": 1},{"a": 1, "b": 2},{"a": 1, "b": 3},{"a": 1, "b": 4}]
```
yoavkatz, line 2009: The code is deterministic - so it will return a particular subset (although the order may be different).

alonh: right.

alonh: right. will update.

PR #868, opened 2024-05-28
Modified recipe and dataset creation to support data classification policy

yoavkatz:
As part of #646  , we added data classification to loaders , metrics, and other artifacts.  However, the full pipeline did not support this.

Other changes:

1. Legacy StreamInstanceOperator removed because in any case  it was not imported where originally added (operators.py)  .
2. Removed legacy examples that are not current methodology.


src/unitxt/standard.py
```diff
@@ -131,11 +131,11 @@ def set_pipelines(self):
         )
         self.standardization = SequentialOperator()
         self.standardization.__description__ = (
-            "Standardizing the dataset fields to task definition."
+            "Standardizing the raw dataset fields to task field definition."
         )
         self.processing = SequentialOperator()
         self.processing.__description__ = (
-            "Processing the task fields (e.g. selecting demos per sample if needed)."
+            "Setting task fields (and add selecting demos per sample if needed)."
```
yoavkatz, line 138:
```suggestion
            "Setting task fields (and selecting demos per sample if needed)."
```

pawelknes (approved):
lgtm

Issue #871, opened 2024-05-30
Convert outputs to list by hyphen with a space ("- ")


marukaz:
There is a processor [to_list_by_comma](https://github.com/IBM/unitxt/blob/31f3abc2581d05221762a60d5e7168d9402fea99/prepare/processors/to_list_by_comma.py#L17), which converts a string like "dog,cat" to a list like `["dog", "cat"]`

However, It is also common to have LLM generate a hyphen-separated list such as

```
- Company A, INC
- Company B Co., Ltd.
```

It would be useful if there is a processor to deal with it.

PR #872, opened 2024-05-30
Add to_list_by_hyphen_space processor

marukaz:
close https://github.com/IBM/unitxt/issues/871

I ran `python -m unittest tests/library/test_postprocessors.py` it passed.


prepare/processors/to_list_by_hyphen.py
```diff
@@ -0,0 +1,39 @@
+from unitxt import add_to_catalog
+from unitxt.operator import SequentialOperator
+from unitxt.operators import RemoveValues
+from unitxt.processors import ToListByHyphenSpace
+
+add_to_catalog(
+    SequentialOperator(
+        steps=[
+            ToListByHyphenSpace(field="prediction", process_every_value=False),
```
elronbandel, line 9:
```suggestion
         SplitStrip(delimiter="- ", strip_every_element=True, process_every_value=False),
```
I think it it is cleaner to have it like this rather then define a new class.

elronbandel (commented):
Not sure about it but i think its easier to have all the logic defined in the prepare file.

prepare/processors/to_list_by_hyphen.py
marukaz: Thank you for the review! Yes, I agree, and in fact, when I personally used this processor, I implemented it in the same file. However, since the implementation of [to_list_by_comma](https://github.com/IBM/unitxt/blob/main/prepare/processors/to_list_by_comma.py) defined a class and separated the files, I followed that approach. I don't have a strong preference, so we could either consolidate this processor into one file, refactor to_list_by_comma, or keep it as is following the current implementation of to_list_by_comma. Which option do you prefer?

elronbandel:
Also, wouldn't be safer to split by "\n-"? What do you think @marukaz?

marukaz:
@elronbandel I think the leading hyphen will remain when splitting by '\n-'

elronbandel:
So just to be on the safe side how about:
```python
from string_operators import RegexSplit

RegexSplit(by="(?<=^|\n)- ")
```
Which will split by either `"- "` if it is at the beginning of the string or if its after `"\n"`
If you do want to go there we can close merge as it is.

marukaz:
hat sounds better! I have updated the code accordingly. Also, I followed [your comment](https://github.com/IBM/unitxt/pull/872#discussion_r1624108215) and removed the class, consolidating everything into the same file.

elronbandel (approved):
Neat!

PR #874, opened 2024-05-31
In card preparers, generate the tags with "singletons" rather than values paired with True

dafnapension:


yoavkatz:
How about calling "singletons" something like  "flags_tags" or just "flags"?
Maybe better ones, but "singletons" seems very technical.

elronbandel (changes requested):
I suggest to remove 'croissent' from that list and change its name to "others"

dafnapension:
No problem at all in replacing the tag name:  `singletons -> flags`.  Just waiting for OK from both Michals. 
Re. `croissant`: @elronbandel , do you want it out at all, as if never has been, or you want it still living as a separate tag, as before:  `croissant: True`  ?

dafnapension:
Hi @elronbandel , @yoavkatz , @michal-jacovi , and @gitMichal ,
I updated the `__tags__` contents so that the field with all the singletons is now named `flags` rather than `singletons`.
I also updated the `__description__`  fields of some of the cards, with contents that @michal-jacovi  provided, having reviewed and updated the contents that came from the HF tools.
I changed `catalog` to reflect the line breaks that Michal determined.

elronbandel (approved):

PR #875, opened 2024-06-02
Fix operator and metrics data bugs and add metric prefix

yoavkatz:
There were some very subtle bugs in treatment of operators.  

The issue was that if we assign a field to a field (E.g in CopyField operator) , and these are both fields that are dictionary.
then field the dictionary values of the source field were copied to the target field.  
If the one of these fields, was further modified - they effected on another.

CopyField(pred -> pred_orig) 
CopyField(pred/contexts -> refs)
ListFields[task/question -> pred)

Following this, pred_orig was equal to the task/question.

As a first solution - I a deep copied instances in instance operators, so they don't effect each other.

Another issue was that metric global scores were shared across instances, and this also caused problems following the above change. The code was simplified that each global score is a separate object.

Finally, added an option for metric to receive a score_prefix, that is added to all scores (except "score" and "score_name"). 
This allows running the same metric on two different fields of the task.

This is a currently branch over 1.9.0 to allow people who need to use it on a stable version before a release.

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -393,6 +393,11 @@ class InstanceFieldOperator(StreamInstanceOperator):
     def process(
         self, instance: Dict[str, Any], stream_name: Optional[str] = None
     ) -> Dict[str, Any]:
+        # Need to deep copy instance, because when assigning two dictionary fields,
+        # dict_set() the target field dictionary fields.
+        # This means that if this target field was assigned to another field before,
+        # the field is updated as well.
+        instance = deepcopy(instance)
         for from_field, to_field in self._field_to_field:
             try:
                 old_value = dict_get(
--- a/src/unitxt/test_utils/card.py
+++ b/src/unitxt/test_utils/card.py
@@ -205,7 +205,7 @@ def test_predictions(
         )
     logger.info("*" * 80)
     logger.info("Sample score output:")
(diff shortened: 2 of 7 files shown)


src/unitxt/operators.py
```diff
@@ -393,6 +393,11 @@ def prepare(self):
     def process(
         self, instance: Dict[str, Any], stream_name: Optional[str] = None
     ) -> Dict[str, Any]:
+        # Need to deep copy instance, because when assigning two dictionary fields,
+        # dict_set() the target field dictionary fields.
+        # This means that if this target field was assigned to another field before,
+        # the field is updated as well.
+        instance = deepcopy(instance)
```
dafnapension, line 400: I think that perhaps this was a feature, not a bug, and perhaps better make this `deepcopy` depend on a special flag arg: `use_deepcopy`

src/unitxt/test_utils/card.py
```diff
@@ -205,7 +205,7 @@ def test_predictions(
         )
     logger.info("*" * 80)
     logger.info("Sample score output:")
-    logger.info(json.dumps(results[0]["score"]["global"], sort_keys=True, indent=4))
+    logger.info(json.dumps(results[0]["score"], sort_keys=True, indent=4))
```
elronbandel, line 208: maybe we should use `construct_dict_str`?

elronbandel (approved):

PR #890, opened 2024-06-06
use aggregators as a uniform incarnation for the across-instances computation for both instance and global metrics

dafnapension:
Thus simplified and thereby extended grouping and filtering over to global metrics and to bulk-instance metrics

### DIFF
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -9,7 +9,6 @@ from abc import ABC, abstractmethod
 from collections import Counter, defaultdict
 from copy import deepcopy
 from dataclasses import field
-from operator import itemgetter
 from statistics import mean
 from typing import Any, Dict, Generator, List, Optional, Tuple, Union
 
@@ -17,14 +16,20 @@ import evaluate
 import numpy
 import numpy as np
 import pandas as pd
-from scipy.stats import bootstrap
 from scipy.stats._warnings_errors import DegenerateDataWarning
 
-from .artifact import Artifact, fetch_artifact
+from .aggregators import (
+    ControlComparisonAggregator,
+    FilterAggregator,
+    GrouperAggregator,
+    MaxAggregator,
+    MeanAggregator,
+)
+from .artifact import fetch_artifact
+from .base_metrics import BulkInstanceMetric, GlobalMetric, InstanceMetric, Metric
 from .dataclass import (
-    AbstractField,
+    Field,
     InternalField,
-    NonPositionalField,
     OptionalField,
 )
 from .deprecation_utils import deprecation
@@ -33,22 +38,16 @@ from .inference import HFPipelineBasedInferenceEngine, InferenceEngine
 from .logging_utils import get_logger
 from .metric_utils import InstanceInput, MetricRequest, MetricResponse
 from .operator import (
-    InstanceOperator,
     MultiStreamOperator,
     SequentialOperator,
     StreamingOperator,
     StreamOperator,
 )
-from .operators import Copy
-from .random_utils import get_seed
-from .settings_utils import get_settings
+from .operators import Copy, FilterByCondition
 from .stream import MultiStream, Stream
-from .type_utils import Type, isoftype, parse_type_string, to_type_string
+from .type_utils import isoftype, parse_type_string, to_type_string
 
 logger = get_logger()
-settings = get_settings()
-
-warnings.filterwarnings("ignore", category=DegenerateDataWarning)
 
 warnings.filterwarnings("ignore", category=DegenerateDataWarning)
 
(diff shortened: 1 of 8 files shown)


src/unitxt/metrics.py
```diff
@@ -224,13 +226,98 @@ def disable_confidence_interval_calculation(self):
         pass
 
 
+class Aggregator(Artifact):
+    @abstractmethod
+    def aggregate_one_group_score_named(
+        self, instances: List[Dict[str, Any]], score_names: List[str]
+    ) -> dict:
+        pass
```
elronbandel, line 234:
```suggestion
class Aggregator(Artifact):
    """
    Aggregate list of instances to a dictionary of scores.
    """"
    score_names: List[str]
    @abstractmethod
    def aggregate(
        self, instances: List[Dict[str, Any]]
    ) -> Dict[str, Any]:
        pass
```

src/unitxt/metrics.py
```diff
@@ -224,13 +226,98 @@ def disable_confidence_interval_calculation(self):
...
+        instances: list of dicts of each instance's instance scores.
+        score_name: score field names to compute the mean for.
+    """
+    return nan_max(
+        [instance["score"]["instance"][score_name] for instance in instances]
+    )
+
+
+class AverageItemsAggregator(SimpleAggregator):
+    aggregating_func = Field(default_factory=lambda: average_item_scores)
```
elronbandel, line 274:
```suggestion
class Mean(Aggregator):
    def aggregate(
        self, instances: List[Dict[str, Any]]
    ) -> dict:
        result = {}
        for score_name in self.score_names:
            result[score_name] = nan_mean(
          [instance["score"]["instance"][score_name] for instance in instances]
      )

class Max(Aggregator):
    def aggregate(
        self, instances: List[Dict[str, Any]]
    ) -> dict:
        result = {}
        for score_name in self.score_names:
            result[score_name] = nan_max(
          [instance["score"]["instance"][score_name] for instance in instances]
      )
```

src/unitxt/metrics.py
```diff
@@ -224,13 +226,98 @@ def disable_confidence_interval_calculation(self):
...
+    # in case of grouping by, the following boolean flag specifies whether resampling for CI
+    # should be done from the individual groups' scores (True), as if each group is represented by
+    # one instance whose instance["score"]["instance"][score_name] is the group's global score for score_name,
+    # Or from the whole stream (False), where each resample is then split to
+    # groups, the score of which is then computed, and finally averaged with the other groups' scores, as done
+    # here for the original whole stream.
+    ci_samples_from_groups_scores: bool = False
+
+    # the basic aggregation along the instances: no split to groups, no filtering
+    aggregator: Aggregator = Field(default_factory=lambda: AverageItemsAggregator())
```
elronbandel, line 317:
```suggestion
    aggregator: Aggregator = Mean()
```

src/unitxt/metrics.py
```diff
@@ -317,7 +380,7 @@ def score_based_confidence_interval(
             # if aggregation_func is None, we simply take the mean of the resampled instance scores
             # otherwise, the aggregation_func needs to be applied AFTER resampling the instances;
             #   that is, re-form the groups, calculate the function, and take the mean of the group scores
-            aggregation_func = self.average_item_scores
+            aggregation_func = AverageItemsAggregator().aggregate_one_group_score_named
```
elronbandel, line 383:
```suggestion
            aggregation_func = self.aggregator
```

src/unitxt/metrics.py
elronbandel:
```
class Metric:
     aggregator: Aggregator = Group(ConfigenceInterval(Mean(fields=["f1"]), group_by="group_id")
```
So every metric has one and only aggregator

elronbandel (changes requested):
See comments

src/unitxt/metrics.py
```diff
@@ -224,134 +253,207 @@ def disable_confidence_interval_calculation(self):
         pass
 
 
+class Aggregator(Artifact):
+    """Aggregates over a list of instances, updating the score/global field of each of them by a dictionary of scores."""
+
+    score_names: List[str] = None
+
+    # non trivial only for grouper_aggregator
+    aggregator_based_score_prefix = ""
```
elronbandel, line 262:
```suggestion
    score_prefix = ""
```
This should be an option to every aggregate (such as mean/ci) then it can add their prefix to every score given from a bottom level aggregator

src/unitxt/metrics.py
```diff
@@ -224,134 +253,207 @@ def disable_confidence_interval_calculation(self):
...
+            new_gs[full_prefix + score_name] = score
+
+        return new_gs
+
+    # trivial for all but grouperaggregator
+    def instances_to_sample_from_and_sample_aggregator(
+        self, instances: List[Dict[str, Any]]
+    ) -> Tuple[List[Dict[str, Any]], Any]:
+        return (instances, self)
+
```
elronbandel, line 298:
```suggestion
     def __call__(self, instances: List[Dict[str, Any]]):
          global_score = self.aggregate(instances)
          global_score = self.global_score(global_score)
          self.update_global_score(instacnces)
```
The aggregator should be in charge of all the affected raised from the aggregation

src/unitxt/metrics.py
dafnapension: Hi @elronbandel ,  recently, Yoav added this option of `score_prefix` for all **metrics**, that is applicable to instance score as well.
Here we are talking about global scores, and as you said: every **aggregator** has this field.
Currently, for backward compatibility, actual use is made only by `GrouperAggergator`,  but all aggregators have this field, as you see.

src/unitxt/metrics.py
dafnapension: It should also supply hints for the CI computer.  So it can not be simply encapsulated.

dafnapension:
Hi @elronbandel , I copy your important comments over to here:

Few comments on your PR:

1. why the global metric dosent use consume_stream?
     (there was a comment that a delicate issue of backward compatibility is hiding there, but I will look again to better understand)
2. Can we have both global metric and instance metric inhereting both from a BaseMetric that:
    a. consume the stream
    b. go over each instance with a function `_work_on_instance`
    c. Then operate on all stream with `_work_on_all_instances`
3. Then they each InstanceMetric and GlobalMetric change only the code inside this two functions
  (A neat idea. I wonder, though, whether, since Instance focuses mainly on the former (b) and Global mainly on the latter (c), this idea will just pump up each in the direction that is not of its concern)

dafnapension:
Maybe return to this after the war ends

dafnapension:
too complicated to maintain rebase-able. So closing for now.

PR #894, opened 2024-06-08
fixed the pitty diff in perplexity result

dafnapension:
Very small diff, smaller than what is noticable by git, but noticable on our dev envs, when running test_metrics


dafnapension:
Not sure why the test results on git are different from those in our dev env.

Issue #893, opened 2024-06-08
Add optional default values to task fields


yoavkatz:
For some task fields, like "text_type", It would good to have defaults so the user does not have to specify them in the card.  For example for text_type it could be "text".

I suggest adding another field to Task , called " defaults" of type Dict[str, Any].

It would add the default values for any instance that does not already have the field.

yoavkatz:
@pawelknes . If you are want to take this issue, it would be great. It continues the work on Tasks.

elronbandel:
My only concern about it is that mechanism does not support inheritance. 

For example if we want to create task for sentiment analysis ideally we would like to inherit most of the definition from the classification task and just set the `class_type=sentiment` as default. Yet our only way to do it and save it in the catalog is to duplicate the original task. Not to mention we dont have  a native way to specify the templates for classification are compatible with sentiment analysis (and ofcourse more can be added specfically just for this task)

If lets say we had a task of type `SubTask` it could inherit regular task and add defaults.
Then sentiment analysis could be define as `SubTask(task="task.classification.multi_class", defaults={"type_of_class":sentiment})`
Or ofcourse it can support: `SubTask(task=Task(...),  defaults={"type_of_class":sentiment})` and be compatible with your original suggestion. 

Another advantage of sub tasks is that it can yield an instruct lab like taxonomy tree where you have sub tasks like:  `SubTask(task="tasks.generation", defaults={"input_type":"email"})` and then `SubTask(task="tasks.generation.from_email", defaults={"output_type":"summary"}) `and  `SubTask(task="tasks.generation.from_email", defaults={"output_type":"bullet points"})`

yoavkatz:
Hi. I agree that SubTask is a useful extension.  Actually we have this syntax:

tasks.generation[defaults={"input_type":"email"}]

The difference is that you want to defaults to be added and not replaced, but this is allow true for metrics

tasks.generation[metrics=["metrics.my_metric"]]

we may want to add the metric and not replace all the existing metrics.

pawelknes:
> @pawelknes . If you are want to take this issue, it would be great. It continues the work on Tasks.

Sure, I can work on this

PR #908, opened 2024-06-14
SubTask

pawelknes:
~Added a new field to the `Task` class named `defaults` which sets default values of an instance for chosen fields in both `inputs` and `outputs`.~

Added a new class `SubTask` which allows creating a new task in order to inherit attributes (`metrics`, `inputs` etc.) from other tasks, and possibly add new fields or overwrite others. 
`prediction_type` must be consistent across all combined tasks. Each following defined sub-task has priority over previous ones since its fields may overwrite values for the same keys defined in previous tasks, while any optional parameter defined for the `SubTask` itself has the top priority.

For #893


src/unitxt/task.py
```diff
@@ -99,9 +105,46 @@ def check_metrics_type(self) -> None:
...
+                )
+
+                assert isoftype(default_value, parse_type_string(val_type)), (
+                    f"The value of '{default_name}' from the 'defaults' must be of "
+                    f"type '{val_type}', however, it is of type '{type(default_value)}'."
+                )
+
+    def set_default_values(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        if self.defaults:
+            instance.update(self.defaults)
```
yoavkatz, line 140: What happens if the value is already set in the instance? In this case, we don't want to value to be updated.

pawelknes: Alright, I thought it should be overwritten anyway. So I'll change that

src/unitxt/task.py
```diff
@@ -119,3 +162,94 @@ def process(
 
 class FormTask(Task):
     pass
+
+
+class SubTask(Task):
```
yoavkatz, line 167: Personally, I thinking this kind of multiple inheritance is to complex for users to understand.  Unitxt is already too complicated for most users.   Even in python it's complex, and people will not be able to debug this.   
We are getting a lot of feedback that unitxt becomes too complicated.

I suggest splitting between these changes, the default one is pretty straight forward - and will save people effort (e.g. not specifying text_type in classification, if they don't need to use it in their templates).  

 I think we should discuss the SubTask again - as we already have methods to modify an existing instance using task.qa[metrics=[mymetric]].

elronbandel:
I think it would be easier to comprehend if instead of having SubTask,  every task would have optional `parent_task`. When `parent_task` is assigned all the fields are populated by the task and its parent. So the inputs would be the inputs of the parent task and in addition the inputs of the current task. Similarly for the the defaults field. That way for instance you could have:
```python
add_to_catalog(Task(
    inputs= { "text" : "str", "target_language" : "str"},
    outputs= {"translation" : "str"},
    prediction_type="str",
    metrics=["metrics.bleu"],
), "tasks.translation")

add_to_catalog(Task(
    parent_task="tasks.translation",
    inputs= { "source_language" : "str"},
), "tasks.translation.specific_language")

add_to_catalog(Task(
    parent_task="tasks.translation.specific_language",
    defaults= { "source_language" : "English"},
), "tasks.translation.from_english")
```
I think its simpler mechanism which is easier to understand and can be useful. what do you think @yoavkatz @pawelknes ?

yoavkatz:
We discussed with Elron, and we think the parent task complicates the API, so for now we will not add it.

pawelknes:
> We discussed with Elron, and we think the parent task complicates the API, so for now we will not add it.

Alright

yoavkatz:
> > We discussed with Elron, and we think the parent task complicates the API, so for now we will not add it.
> 
> Alright

So we are closing this for now.

PR #909, opened 2024-06-16
Add LogProbInferenceEngines API and implement for OpenAI

lilacheden:
- add logprobs option
- add data_classification_policy 
- fix return type to a list of strings as in other inference engines


arielge (approved):

src/unitxt/inference.py
```diff
@@ -183,8 +190,9 @@ def prepare(self):
         self.client = OpenAI(api_key=api_key)
 
     def _infer(self, dataset):
```
OfirArviv, line 192: please add typing with full interface, as its non trivial now

OfirArviv (changes requested):

src/unitxt/inference.py
OfirArviv: also if you return log prob, you don't have to return a list of strings. Just be clear in the return type and in the documentation

OfirArviv: currently the only thing that uses the inference engine is the llamaj, so we can adapt the interface as needed, if its better.
I think we can consider having a different return type, or even a different function.

lilacheden: I can add helper functions for readability, but I still think that this is the most convenient interface. We currently use it from fm-eval, and the logprobs postprocessors expect strings (same as they get from bam inference and python inference in fm-eval run_text2text), so I don't see a reason to change that.

yoavkatz: We added a separate interface for inference of logics.

src/unitxt/inference.py
```diff
@@ -15,11 +17,27 @@ def _infer(self, dataset):
         """Perform inference on the input dataset."""
         pass
 
+    def _infer_log_probs(self, dataset):
```
OfirArviv, line 20: make abstract method like _infer

src/unitxt/inference.py
```diff
@@ -15,11 +17,27 @@ def _infer(self, dataset):
...
+        raise NotImplementedError(
+            f"{self.__name__} does not support inference returning log probabilities"
+        )
+
     def infer(self, dataset):
         """Verifies instances of a dataset and performs inference."""
         [self.verify_instance(instance) for instance in dataset]
         return self._infer(dataset)
 
+    def infer_log_probs(self, dataset):
```
OfirArviv, line 31: return type?

yoavkatz: Added.

src/unitxt/inference.py
yoavkatz: I create a LogProbInferenceEngine API , so there is no need for all classes to implement this method.

PR #911, opened 2024-06-16
Added trust_remote_code to HF dataset query operations

yoavkatz:


prepare/cards/mlsum.py
```diff
@@ -6,9 +6,14 @@
     TaskCard,
 )
 from unitxt.catalog import add_to_catalog
+from unitxt.settings_utils import get_settings
 from unitxt.test_utils.card import test_card
 
-langs = get_dataset_config_names("mlsum")  # the languages
+settings = get_settings()
+
+langs = get_dataset_config_names(
```
OfirArviv, line 14: shouldn't this be inside the function? Why pass it every time ?

PR #916, opened 2024-06-18
feat(inference operator): allow inference engine to be used as operator

jezekra1:
Provide option to run inference in a stream:

```
multi_stream = inference_engine(multi_stream)
```
Will enhance instances by `prediction` and `generation_info` fields

### DIFF
--- a/src/unitxt/inference.py
+++ b/src/unitxt/inference.py
@@ -1,15 +1,19 @@
 import abc
 import os
 from dataclasses import field
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, Generator, List, Literal, Optional, Union
 
 from tqdm import tqdm
 
+from . import get_logger
 from .artifact import Artifact
-from .operator import PackageRequirementsMixin
+from .deprecation_utils import deprecation
+from .operator import PackageRequirementsMixin, PagedStreamOperator
+
+logger = get_logger()
 
 
-class InferenceEngine(abc.ABC, Artifact):
+class InferenceEngine(abc.ABC, PagedStreamOperator):
     """Abstract base class for inference."""
 
     @abc.abstractmethod
@@ -25,8 +29,10 @@ class InferenceEngine(abc.ABC, Artifact):
 
     def infer(self, dataset):
         """Verifies instances of a dataset and performs inference."""
-        [self.verify_instance(instance) for instance in dataset]
-        return self._infer(dataset)
+        for instance in dataset:
+            self.verify_instance(instance)
+        outputs, _metadata = self._infer(dataset)
+        return outputs
 
     def infer_log_probs(self, dataset):
         """Verifies instances of a dataset and performs inference that returns log probabilities of top tokens.
@@ -35,8 +41,15 @@ class InferenceEngine(abc.ABC, Artifact):
         [ "top_tokens": [ "text": ..., "logprob": ...} , ... ]
 
         """
-        [self.verify_instance(instance) for instance in dataset]
+        for instance in dataset:
+            self.verify_instance(instance)
         return self._infer_log_probs(dataset)
+
+    def process(self, page: List[Dict], stream_name: Optional[str] = None) -> Generator:
+        for instance, output, metadata in zip(page, *self._infer(page)):
+            instance["prediction"] = output
+            instance["generation_info"] = metadata
+            yield instance
 
 
 class HFPipelineBasedInferenceEngine(InferenceEngine, PackageRequirementsMixin):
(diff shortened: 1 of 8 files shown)


src/unitxt/inference.py
```diff
@@ -106,6 +121,7 @@ def _infer(self, dataset):
         return ["[[10]]" for instance in dataset]
 
 
+@deprecation(version="1.11.0", alternative="Plain dictionary")
```
jezekra1, line 124: This interface does not contain all parameters and it's errorprone because it has to be updated with every change in GenAI sdk. We can just use plain dict instead and validate it using the `TextGenerationParameters` model from SDK. I suggest a similar approach with other models (openAI parameters)

OfirArviv: The problem is most user don't know what the params are. The definition is for the user who do not know what their names are. I prefer it to be fully defined and set the version of the sdk.

src/unitxt/inference.py
```diff
@@ -35,9 +41,16 @@ def infer_log_probs(self, dataset):
         [ "top_tokens": [ "text": ..., "logprob": ...} , ... ]
 
         """
-        [self.verify_instance(instance) for instance in dataset]
+        for instance in dataset:
+            self.verify_instance(instance)
         return self._infer_log_probs(dataset)
 
+    def process(self, page: List[Dict], stream_name: Optional[str] = None) -> Generator:
```
yoavkatz, line 48: Right now, the process calls the regular inference call and not the log_prob inference.  
Perhaps it would be clearer and more flexible if we used composition , so we would have two classes:

InferenceEngineOperator(engine: InferenceEngine )
LogProbInferenceEngineOperator(engine: InferenceEngine )

which would implement the process() method ?

src/unitxt/inference.py
yoavkatz: Yes.   It's a tradeoff :
   type API (that is self including documentation) - that needs to be updated with each version of SDK
vs 
   non typed  API - that requires the user to go to the external API to see the params and assume it's compatible
   
@OfirArviv - Not that today - we don't require specific version of any dependent  library in unitxt, so people can choose whatever fit for them given their other requirements - we assume the level of API we use does not change.   This is a tradeoff in itself - fexlibility for user, vs risk on mismatch between version.

src/unitxt/inference.py
jezekra1: Makes sense, I this more too, please mention it in the [unmerged PR](https://github.com/IBM/unitxt/pull/909).

src/unitxt/inference.py
jezekra1: Would it be better if the Artifact class is dynamically generated from the pydantic type (I can look into this) or do you still prefer to have it explicitly written in code so that people can inspect it?

src/unitxt/inference.py
yoavkatz: I actually think there are two options.  

1) Create two types of inference engine:

InferenceEngine
LogProbInferenceEngine

Then each can have a single infer method.  They can also be derived from an Operator with  a  "process" statement.

2) Keep  one type of inference engine and have two separator operators (as we discussed before)
 
Somehow, I now tend to the first option, but I'm not sure.
 
[BTW, there is also a request for adding EmbeddingEngine, which embeds each of the input ] 
 
 What do you think ? @jezekra1 , @OfirArviv , @elronbandel , @lilacheden ?

jezekra1: > Somehow, I now tend to the first option, but I'm not sure.

I like it too because it's obvious which model supports which method instead of raising `NotImplementedError`

src/unitxt/inference.py
yoavkatz: I think at this point, I'd prefer explicating typing, so it's clear documented.   This helps the users understand what they can and can not control.    I realize it's not a clear cut case, but it's more consistent with other places in the library.

src/unitxt/inference.py
yoavkatz: I talked with Ofir  -  other libraries have one class for inference engine.  [e.g LM-eval harness  https://github.com/EleutherAI/lm-evaluation-harness/blob/main/lm_eval/api/model.py ]   It will also not require a separate class for input parameters when used in the different modes.

yoavkatz: After much deliberations, we think that there should be a separate Operator class.

class Infer(PagedStreamOperator)):  
   inference_engine: InferenceEngine
   to_field : str ="prediction"
   to_metadata_field : str = "generate_info"
   
   def process(self, page: List[Dict], stream_name: Optional[str] = None) -> Generator:
          for instance, output, metadata in zip(page, *self._infer(page)):
             instance[self.to_field] = output
             instance[self.to_metadata_field] = metadata
             yield instance

src/unitxt/inference.py
```diff
@@ -142,30 +159,42 @@ def prepare(self):
...
+        metadata = []
+
+        for resp in self.client.text.generation.create(
+            inputs=prompts, model_id=self.model_name, parameters=self.parameters
+        ):
+            for result in resp.results:
+                predictions.append(result.generated_text)
+                metadata.append(
+                    {
+                        "meta": resp.model_dump(exclude={"results"}, exclude_none=True),
```
yoavkatz, line 189: @jezekra1 - what do you need the meta data for? If each inference engine, returns different meta data, then the code that uses it, can not do much with it except record it.

PR #920, opened 2024-06-19
feat(metric): support adhoc metric

jezekra1:
This will allow to specify metric ad-hoc in recipe, i.e.:
```
      "task": {
        "type": "form_task",
        "inputs": [
          "text_type",
          "classes",
          "input"
        ],
        "outputs": {
          "output": "str"
        },
        "metrics": [
          "metrics.f1_micro",
          "metrics.accuracy",
          {
            "type": "llm_as_judge",
            "name" "metrics.my_custom_metric_defined_ad_hoc"
            "inference_model": {
              "type": "ibm_gen_ai_inference_engine",
              "model_name": "meta-llama/llama-3-70b-instruct",
              "parameters": {
                "max_new_tokens": 252
              }
            },
            "template": "templates.safety.unsafe_content",
            "task": "rating.single_turn",
            "format": "formats.llama3_chat",
            "main_score": "llama_3_70b_instruct_ibm_genai_template_unsafe_content"
          }
        ]
      }
```

The evaluation logic does not take this into account now and would need to be adjusted, however I think it should not be forbidden on the task definition level.

We have our own evaluation function which I can provide as well, however I don't have insight into all the low level details in `MetricOperator` and other places where evaluation can occur so this would be quite time consuming for me to dive into, I'd appreciate any help on this front :)

### DIFF
--- a/src/unitxt/llm_as_judge.py
+++ b/src/unitxt/llm_as_judge.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, List, Literal, Optional
 
 from .api import evaluate, produce
+from .artifact import Artifact
 from .inference import InferenceEngine, OpenAiInferenceEngine
 from .metrics import BulkInstanceMetric
 from .operator import SequentialOperator
@@ -38,6 +39,8 @@ class LLMAsJudge(BulkInstanceMetric):
             instances = []
             for task_data_instance in task_data:
                 template = task_data_instance["metadata"]["template"]
+                if Artifact.is_artifact_dict(template):
+                    template = Artifact.from_dict(template)
                 instance = SequentialOperator(
                     steps=[template, "formats.empty"]
                 ).process_instance(
--- a/src/unitxt/artifact.py
+++ b/src/unitxt/artifact.py
@@ -11,7 +11,6 @@ from .dataclass import (
     AbstractField,
     Dataclass,
     Field,
-    InternalField,
     NonPositionalField,
     fields,
 )
@@ -127,7 +126,9 @@ class Artifact(Dataclass):
     __tags__: Dict[str, str] = NonPositionalField(
         default_factory=dict, required=False, also_positional=False
     )
-    __id__: str = InternalField(default=None, required=False, also_positional=False)
+    __id__: str = NonPositionalField(
+        default=None, required=False, also_positional=False
+    )
 
     data_classification_policy: List[str] = NonPositionalField(
         default=None, required=False, also_positional=False
--- a/src/unitxt/task.py
+++ b/src/unitxt/task.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, List, Optional, Union
 
 from .artifact import fetch_artifact
 from .logging_utils import get_logger
+from .metrics import Metric
 from .operator import InstanceOperator
 from .type_utils import (
     get_args,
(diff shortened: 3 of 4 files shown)


src/unitxt/llm_as_judge.py
```diff
@@ -38,6 +39,8 @@ def _get_input_instances(self, task_data: List[Dict]) -> List:
             instances = []
             for task_data_instance in task_data:
                 template = task_data_instance["metadata"]["template"]
+                if Artifact.is_artifact_dict(template):
+                    template = Artifact.from_dict(template)
```
jezekra1, line 43: I also found a bug in `llm_as_judge` where a template was defined in-place instead of a string identifier from catalog.

yoavkatz:
This is a good change. We actually probably want to allow this also for post processors.

src/unitxt/metrics.py
```diff
@@ -102,6 +101,8 @@ class Metric(Artifact):
     #
     score_prefix: str = ""
 
+    name: Optional[str] = None
```
yoavkatz, line 104: What about using the __id__ field?  Can it work? Prefer not to add another identifier.

jezekra1: I tried it, but it's not possible to set it from outside:
```
Value error, Invalid unitxt artifact: Unexpected keyword argument(s) {'__id__': 'test'} for class LLMAsJudge.
Should be one of: ['type', '__description__', '__tags__', '__id__', 'data_classification_policy', ...]
```
It's probably due to the double undescores which are [masked in python](https://docs.python.org/3/tutorial/classes.html#private-variables). I tried prefixing with classname, but that also doesn't work: `"_LLMAsJudge__id__": "test"`

Maybe we can make `__id__` public and overwritable?

jezekra1: Hmm, actually it's enough to change it from `InternalField` to `NonPositionalField`, seems it's handled differently in dataclasses. I changed it in `Artifact`

yoavkatz: I understand. Let's see if  @elronbandel has any suggestion.

yoavkatz: But I'd like @elronbandel to give his feedback, because it seems (also according to the test), that is was intentional.  Maybe __id__ was only for artifacts in the catalog.

jezekra1: This might need to be changed as well:
```python
    @staticmethod
    def artifact_to_jsonable(artifact):
        if artifact.__id__ is None:
            return artifact.to_dict()
        return artifact.__id__
```

https://github.com/IBM/unitxt/blob/31f3abc2581d05221762a60d5e7168d9402fea99/src/unitxt/schema.py#L30

elronbandel:
I think we need to test in an end to end test with evaluation using Huggingface load_dataset. The reason for that is that if we add metrics as they are we need to transfer them to textual format when using the Huggingface dataset, then at evaluation time recover them back to the metric python classes instances.

What we can do is to add an operator at the end of the recipe that will transfer all the metrics to strings by their catalog name in case they are saved in the catalog or to their stringified json representation otherwise. The at evaluation time to go over the metrics list and make the json regular dicts so the evaluation pipeline can recover them back to python classes instances. 

Then finally the argument `metrics` of the task can be defined simply as : `metrics: List[Metric]` which will cover all cases.

yoavkatz:
@jezekra1 - We are taking these ideas and implementing in a more fundamental way in #922.

jezekra1:
Awesome, thank you! :)

yoavkatz:
Closing as handled in #922 (merged to main)

PR #921, opened 2024-06-19
Added defaults to tasks

pawelknes:
Added a new field to the `Task` class named `defaults` which sets default values of an instance for chosen fields in both `inputs` and `outputs`.

For #893 issue and discussions in #908


src/unitxt/task.py
```diff
@@ -99,9 +106,46 @@ def check_metrics_type(self) -> None:
...
+                    f"which is of type '{type(self.defaults)}'."
+                )
+
+            for default_name, default_value in self.defaults.items():
+                assert isinstance(default_name, str), (
+                    f"If specified, all keys of the 'defaults' must be strings, "
+                    f"however, the key '{default_name}' is of type '{type(default_name)}'."
+                )
+
+                val_type = self.inputs.get(default_name) or self.outputs.get(
```
yoavkatz, line 124: This assumes the task fields  are in the new format (dictionary) and not in the old list format (still used by some people).  Perhaps add a check , that inputs and outputs are dict and fail with "default params are only supported when input and output fields are defined as { "field_name1" : "field_type1" , ....} and not [ "field_name1" ... ] .

pawelknes: But this happens at the end of the `verify` method which converts both `inputs` and `outputs` to directories anyway

yoavkatz: I see.  IF so, please ignore this.

yoavkatz (commented):
@pawelknes - can you  also make some changes to actual tasks so they work with only needed data (such as "text_type" should default to "text",   "text_a_type" should default to "first text"  ,  "text_a_type" should default to "second text")

Then you can also remove the default from the actual cards.

unitxt/prepare/tasks/classification.py 
unitxt/prepare/tasks/generation.py
unitxt/prepare/tasks/rewriting.py

pawelknes:
> @pawelknes - can you also make some changes to actual tasks so they work with only needed data (such as "text_type" should default to "text", "text_a_type" should default to "first text" , "text_a_type" should default to "second text")
> 
> Then you can also remove the default from the actual cards.
> 
> unitxt/prepare/tasks/classification.py unitxt/prepare/tasks/generation.py unitxt/prepare/tasks/rewriting.py

I've added default values you mentioned as well as some others which I found quite common across different cards

prepare/cards/20newsgroups_sklearn.py
```diff
@@ -42,13 +42,7 @@
         ),
         RenameFields(field_to_field={"data": "text", "target": "label"}),
         MapInstanceValues(mappers={"label": map_labels}),
-        Set(
-            fields={
-                "classes": list(map_labels.values()),
-                "text_type": "text",
-                "type_of_class": "topic",
```
yoavkatz, line 49: While topic classification is a common task, it can not be assumed as a default for "tasks.classification.multi_class", because many other (e.g .sentiment) would get a wrong default.
 Thus I would recommend removing "type_of_class": "topic" from the  "tasks.classification.multi_class" and create a new "tasks.classification.multi_class.topic_classification" which that default.

prepare/tasks/classification.py
```diff
@@ -48,6 +50,7 @@
             "metrics.f1_macro_multi_label",
         ],
         augmentable_inputs=["text"],
+        defaults={"text_type": "text", "type_of_classes": "topics"},
```
yoavkatz, line 53: Here too, I would remove "type_of_classes": "topics"

elronbandel (approved):
Looks good to me.

prepare/tasks/regression/tasks.py
```diff
@@ -13,6 +13,7 @@
         prediction_type="Any",
         metrics=["metrics.spearman"],
         augmentable_inputs=["text"],
+        defaults={"attribute_name": "similarity"},
```
yoavkatz, line 16: While similarity is a common regression, we can assume it's the default, because there are others..

yoavkatz (changes requested):
Hi @pawelknes - I think in some cases , we can not add defaults although they are the common values (defaults, should make sense in all cases).  See details in the comments below.

prepare/tasks/regression/tasks.py
elronbandel: I think that if we want to assume that we should specify "similarity" in the task name

yoavkatz: We can "tasks.regression.two_texts.similarity".

prepare/tasks/classification.py
```diff
@@ -103,7 +108,26 @@
         prediction_type="str",
         metrics=["metrics.f1_micro", "metrics.accuracy", "metrics.f1_macro"],
         augmentable_inputs=["text"],
+        defaults={"text_type": "text", "type_of_class": "topic"},
```
yoavkatz, line 111:
```suggestion
        defaults={"text_type": "text"},
```

yoavkatz: This is only used in legal_bench.py

yoavkatz (approved):

PR #922, opened 2024-06-19
Allow add-hoc metrics and template (and Add first version of standalone example of dataset with LLM as a judge )

eladven:
… metric

### DIFF
--- a/examples/standalone_evaluation_llm_as_judge.py
+++ b/examples/standalone_evaluation_llm_as_judge.py
@@ -1,12 +1,11 @@
 from unitxt import get_logger
-from unitxt.api import evaluate
+from unitxt.api import evaluate, load_dataset
 from unitxt.blocks import Task, TaskCard
 from unitxt.inference import (
     HFPipelineBasedInferenceEngine,
 )
 from unitxt.llm_as_judge import LLMAsJudge
 from unitxt.loaders import LoadFromDictionary
-from unitxt.standard import StandardRecipe
 from unitxt.templates import InputOutputTemplate, TemplatesDict
 
 logger = get_logger()
@@ -75,9 +74,7 @@ card = TaskCard(
 )
 
 # Convert card to a dataset
-recipe = StandardRecipe(card=card, template_card_index="simple")
-stream = recipe()
-dataset = stream.to_dataset()
+dataset = load_dataset(card=card, template_card_index="simple")
 test_dataset = dataset["test"]
 
 # Infer a model to get predictions.
@@ -88,17 +85,18 @@ inference_model = HFPipelineBasedInferenceEngine(
 predictions = inference_model.infer(test_dataset)
 
 # Evaluate the predictions using the defined metric.
-dataset_with_scores = evaluate(predictions=predictions, data=test_dataset)
+evaluated_dataset = evaluate(predictions=predictions, data=test_dataset)
 
-for sample, prediction in zip(dataset_with_scores, predictions):
+for instance in evaluated_dataset:
     logger.info("*" * 80)
-    logger.info(f"Model input:\n{sample['source']}")
-    logger.info(f"Model prediction (as returned by the model):\n{prediction}")
-    logger.info(f"Model prediction (after post processing):\n{sample['prediction']}")
-    logger.info(f"References:\n{sample['references']}")
-    score_name = sample["score"]["instance"]["score_name"]
-    score = sample["score"]["instance"]["score"]
+    logger.info(f"Model input:\n{instance['source']}")
+    logger.info(
+        f"Model prediction (as returned by the model):\n{instance['prediction']}"
+    )
+    logger.info(f"References:\n{instance['references']}")
+    score_name = instance["score"]["instance"]["score_name"]
+    score = instance["score"]["instance"]["score"]
     logger.info(f"Sample score ({score_name}) : {score}")
-global_score = dataset_with_scores[0]["score"]["global"]["score"]
+global_score = evaluated_dataset[0]["score"]["global"]["score"]
 logger.info("*" * 80)
 logger.info(f"Aggregated score ({score_name}) : {global_score}")
--- a/examples/standalone_evaluation_llm_as_judge.py
+++ b/examples/standalone_evaluation_llm_as_judge.py
@@ -11,13 +11,14 @@ from unitxt.templates import InputOutputTemplate, TemplatesDict
 
 logger = get_logger()
 
+# First, we define the examples data.
 data = {
     "test": [
         {"question": "What is the capital of Texas?", "answer": ""},
         {"question": "What is the color of the sky?", "answer": ""},
     ]
 }
-
+# Second, We define the prompt we show to the judge.
 judge_correctness_template = InputOutputTemplate(
     instruction="Please act as an impartial judge and evaluate if the answer of the assistant is correct."
     "Rate the response on a scale of 1 to 10, where 1 means totally wrong, and 10 means totally correct,"
@@ -29,18 +30,17 @@ judge_correctness_template = InputOutputTemplate(
         r"processors.extract_mt_bench_rating_judgment",
     ],
 )
-
+# Third, We define the metric as LLM as a judge, with the desired platform and model.
 llm_judge_metric = LLMAsJudge(
     inference_model=HFPipelineBasedInferenceEngine(
         model_name="google/flan-t5-large", max_new_tokens=256, use_fp16=True
     ),
     template=judge_correctness_template,
     task="rating.single_turn",
-    format="formats.models.mistral.instruction",
     main_score="flan-t5-large_huggingface",
     strip_system_prompt_and_format_from_inputs=False,
 )
-
+# we wrapped all ingredients in a task card.
 card = TaskCard(
     loader=LoadFromDictionary(data=data),
     task=Task(
@@ -61,16 +61,20 @@ card = TaskCard(
     ),
 )
 
+# Convert card to a dataset
 recipe = StandardRecipe(card=card, template_card_index="simple")
 stream = recipe()
 dataset = stream.to_dataset()
 test_dataset = dataset["test"]
 
+# Infer a model to get predictions.
 model_name = "google/flan-t5-base"
 inference_model = HFPipelineBasedInferenceEngine(
     model_name=model_name, max_new_tokens=32
 )
 predictions = inference_model.infer(test_dataset)
+
+# Evaluate the predictions using the defined metric.
 dataset_with_scores = evaluate(predictions=predictions, data=test_dataset)
 
 for sample, prediction in zip(dataset_with_scores, predictions):
--- a/examples/standalone_evaluation_llm_as_judge.py
+++ b/examples/standalone_evaluation_llm_as_judge.py
@@ -20,9 +20,10 @@ data = {
 }
 # Second, We define the prompt we show to the judge.
 judge_correctness_template = InputOutputTemplate(
-    instruction="Please act as an impartial judge and evaluate if the answer of the assistant is correct."
-    "Rate the response on a scale of 1 to 10, where 1 means totally wrong, and 10 means totally correct,"
-    ' by strictly following this format: "[[rating]]"'
+    instruction="Please act as an impartial judge and evaluate if the assistant's answer is correct."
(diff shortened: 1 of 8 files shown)


examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,97 @@
...
+)
+
+add_to_catalog(
+    rating_template,
+    "templates.rating_template",
+    overwrite=True,
+)
+
+rating_metric = LLMAsJudge(
+    inference_model=HFPipelineBasedInferenceEngine(
```
yoavkatz, line 40: I would use a  strong large mode in service  IBMGenAI (or OpenAI) in the example.

examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,97 @@
...
+        model_name="google/flan-t5-large", max_new_tokens=256, use_fp16=True
+    ),
+    template="templates.rating_template",
+    task="rating.single_turn",
+    format="formats.models.mistral.instruction",
+    main_score="flan-t5-large_huggingface",
+    strip_system_prompt_and_format_from_inputs=False,
```
yoavkatz, line 47: Why is this required?

examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,97 @@
...
+
+add_to_catalog(
+    rating_metric,
+    "metrics.rating_metric",
+    overwrite=True,
+)
+
+card = TaskCard(
+    loader=LoadFromDictionary(data=data),
+    task=Task(
```
yoavkatz, line 58: I would move in this example to use the catalog (https://github.com/IBM/unitxt/blob/03ef303b280fdc0453b23db8620a1f0482a9a8b0/examples/qa_evaluation.py#L21) and override the metic.

This will make it shorter, and also demonstrate how one overrides the metric.

examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,97 @@
...
+    ' by strictly following this format: "[[rating]]"'
+    ".\n\n",
+    input_format="[Question]\n{question}\n\n" "[Assistant's Answer]\n{answer}\n",
+    output_format="[[{rating}]]",
+    postprocessors=[
+        r"processors.extract_mt_bench_rating_judgment",
+    ],
+)
+
+add_to_catalog(
```
yoavkatz, line 33: Best if we can avoid storing in the catalog - and just pass as object.  Note it requires a fix as in #920.

src/unitxt/artifact.py
```diff
@@ -404,13 +407,27 @@ def __str__(self):
...
+def fetch_artifact(artifact_rep):
+    if isinstance(artifact_rep, Artifact):
+        return artifact_rep, None
+    if Artifact.is_artifact_file(artifact_rep):
+        return Artifact.load(artifact_rep), None
 
-    artifactory, name, args = get_artifactory_name_and_args(name=name)
-
-    return artifactory.get_with_overwrite(name, overwrite_args=args), artifactory
+    try:
```
yoavkatz, line 416: Instead of try catch, use the method that checks legal catalog artifact name.

src/unitxt/artifact.py
```diff
@@ -279,9 +279,12 @@ def __post_init__(self):
     def _to_raw_dict(self):
         return {"type": self.type, **self._init_dict}
 
-    def save(self, path):
+    def to_json(self):
         data = self.to_dict()
-        save_json(path, data)
+        return json_dump(data)
+
```
yoavkatz, line 285: def serialize(self):
    if ( __id___) return __id___ else self.to_json()

src/unitxt/artifact.py
yoavkatz: Take 413-430 become Artifact.deserialize(str)

yoavkatz:
Elad - when completed also update unitxt/tests/library/test_api.py evaluate tests, which currently does not check serialization of post processors, because it's not done correctly.

examples/standalone_evaluation_llm_as_judge.py
yoavkatz: Why should it be false.  Doesn't it work with 'true'?

examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,104 @@
...
+    input_format="[Question]\n{question}\n\n" "[Assistant's Answer]\n{answer}\n",
+    output_format="[[{rating}]]",
+    postprocessors=[
+        r"processors.extract_mt_bench_rating_judgment",
+    ],
+)
+
+platform = "hf"
+model_name = "google/flan-t5-large"
+inference_model = HFPipelineBasedInferenceEngine(
```
elronbandel, line 37:
```suggestion
inference_model = HFPipelineInferenceEngine(
```

examples/standalone_evaluation_llm_as_judge.py
eladven: I added Lamma (in commented out section

examples/standalone_evaluation_llm_as_judge.py
eladven: As we discussed, currently the code supports only rating.single_turn (and rating.single_turn_with_reference), The main score name has been changed. 
Regarding strip_system_prompt_and_format_from_inputs. Currently setting it to true caused the run to crash. I think we should solve it, but in a separate PR.

src/unitxt/artifact.py
eladven: Done

examples/standalone_evaluation_llm_as_judge.py
yoavkatz: It's a tradeoff - we do want people to use catalog tasks - and we already have an example like this.  We don't want to repeat this on every example and we want to show new capabilities .  Still,   If you feel strong about it though, I'm fine.

examples/standalone_evaluation_llm_as_judge.py
yoavkatz: Ok.

examples/standalone_evaluation_llm_as_judge.py
yoavkatz: Best  be in a different PR.

examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,104 @@
...
+                input_format="{question}",
+                output_format="{answer}",
+                postprocessors=["processors.lower_case"],
+            )
+        }
+    ),
+)
+
+# Convert card to a dataset
+recipe = StandardRecipe(card=card, template_card_index="simple")
```
yoavkatz, line 78: These 3 lines can now be done with load_dataset(card=card, template_card_index="simple").

examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,104 @@
...
+
+# Infer a model to get predictions.
+model_name = "google/flan-t5-base"
+inference_model = HFPipelineBasedInferenceEngine(
+    model_name=model_name, max_new_tokens=32
+)
+predictions = inference_model.infer(test_dataset)
+
+# Evaluate the predictions using the defined metric.
+dataset_with_scores = evaluate(predictions=predictions, data=test_dataset)
```
yoavkatz, line 91: Now that predictions and processed predictions are returns.  These lines can be set done via:

```
evaluated_dataset = evaluate(predictions=predictions, data=test_dataset)

# Print results
for instance in evaluated_dataset:
    logger.info("*" * 80)
    logger.info(f"Model input:\n{instance['source']}")
    logger.info(
        f"Model prediction (as returned by the model):\n{instance['prediction']}"
    )
    logger.info(
        f"Model prediction (after post processing):\n{instance['processed_prediction']}"
    )
    logger.info(f"References:\n{instance['references']}")
    logger.info(
        f"References (after post processing):\n{instance['processed_references']}"
    )
    score_name = instance["score"]["instance"]["score_name"]
    score = instance["score"]["instance"]["score"]
    logger.info(f"Sample score ({score_name}) : {score}")
global_score = evaluated_dataset[0]["score"]["global"]["score"]
logger.info("*" * 80)
logger.info(f"Aggregated score ({score_name}) : {global_score}")
```

This is simpler.

src/unitxt/dataset_utils.py
```diff
@@ -20,13 +21,18 @@ def parse(query: str):
...
     register_all_artifacts()
-    recipe = fetch(dataset_str)
+    recipe = fetch(dataset)
     if recipe is None:
-        args = parse(dataset_str)
-        if "__type__" not in args:
-            args["__type__"] = settings.default_recipe
+        args = parse(dataset)
+        if "type" not in args:
+            args["type"] = settings.default_recipe
```
elronbandel, line 36: Has changed on main to `__type__`

src/unitxt/llm_as_judge.py
```diff
@@ -121,22 +122,22 @@ def compute(
...
-            recipe = f"{recipe},system_prompt={self.system_prompt}"
+            recipe_args["system_prompt"] = self.system_prompt
         if self.format:
-            recipe = f"{recipe},format={self.format}"
-
+            recipe_args["format"] = self.format
+        recipe = Artifact.from_dict(recipe_args)
         dataset = produce(instances, recipe)
         verdicts = self.inference_model.infer(dataset)
         meta_scores = evaluate(predictions=verdicts, data=dataset)
```
elronbandel, line 139: Why is this necessary? Just for the postprocessors utilisation?

elronbandel (commented):
See my comments. In general I think we should have post_process function like produce that can do so without calculating metrics (for no reason). Also it would be useful for others to have a function like infer(instances, recipe, inference_engine) that can bring all this utility in one line (this was requested by a team using paraphrasing for augmentations and would love to utilise all our tools for that end to end.

examples/standalone_evaluation_llm_as_judge.py
eladven: Done. Indeed much clearer

examples/standalone_evaluation_llm_as_judge.py
eladven: Done. It is indeed simpler, but please note that the evaluated_dataset doesn't contain 'processed_prediction' field. Therefore we avoid printing the prediction before the processing (I am not sure, but I think that the prediction field is after the postprocessing).

src/unitxt/dataset_utils.py
eladven: Done. Thanks for letting me know.

src/unitxt/llm_as_judge.py
eladven: The code here is using unitxt to build a dataset for the judge inference. The old code used a representation of string, that represents dict, that represents a recipe.  In addition to the fact that this way is much less clear and bug-prone, it doesn't allow to use ad-hoc template.

elronbandel: Agree, the question is about the use of `evaluate`, is it just for the post-processing aspect?

yoavkatz: I believe so.  Right now, there is no simple way to run the post processors without executing the metrics.

examples/standalone_evaluation_llm_as_judge.py
```diff
@@ -0,0 +1,105 @@
...
+model_name = "google/flan-t5-base"
+inference_model = HFPipelineBasedInferenceEngine(
+    model_name=model_name, max_new_tokens=32
+)
+predictions = inference_model.infer(test_dataset)
+
+# Evaluate the predictions using the defined metric.
+evaluated_dataset = evaluate(predictions=predictions, data=test_dataset)
+
+for instance in evaluated_dataset:
```
yoavkatz, line 90: Simplify with new printouts format using print_dict()

yoavkatz (approved):

yoavkatz (changes requested):
Also - don't forget to update the example documentation.

examples/standalone_evaluation_llm_as_judge.py
eladven: done

yoavkatz (approved):

PR #923, opened 2024-06-19
Update example intro and simplified WNLI cards

yoavkatz:
since WNLI is used in the demos -  simplified it , so each processing effects one field, and it used explicit splitting.

### DIFF
--- a/docs/docs/examples.rst
+++ b/docs/docs/examples.rst
@@ -3,27 +3,29 @@
 Examples ✨
 ==============
 
-Here you find complete examples of using Unitxt
+Here you find complete examples showing how to perform different tasks using Unitxt. 
+Each example is self contained python file that you can run and later modify.
+
 
 .. list-table:: 
    :widths: 50 50 50 50
    :header-rows: 1
 
-   * - Example
+   * - What do you want to do?
      - Description
      - Link to code
      - Related documentation
-   * - Evaluate existing dataset 
+   * - Evaluate an existing dataset from the Unitxt catalog
      - Demonstrates how to evaluate an existing entailment dataset (wnli) using Huggingface 
        datasets and evaluate APIs, with no installation required.  
      - `code <https://github.com/IBM/unitxt/blob/main/examples/evaluate_existing_dataset_no_install.py>`_
      - :ref:`Evaluating datasets <evaluating_datasets>`.  
-   * - Evaluate a user define open QA dataset 
+   * - Evaluate your question-answering dataset 
      - Demonstrates how to evaluate a user QA answering dataset in a standalone file using a user defined task and template.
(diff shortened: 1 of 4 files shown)


docs/docs/examples.rst
```diff
@@ -3,27 +3,31 @@
 Examples ✨
 ==============
 
-Here you find complete examples of using Unitxt
+Here you find complete examples showing how to perform different tasks using Unitxt. 
+Each example is self contained python file that you can run and later modify.
```
yoavkatz, line 7:
```suggestion
Each example is a self contained python file that you can run and later modify.
```

docs/docs/examples.rst
```diff
@@ -3,27 +3,31 @@
...
+       | :ref:`Relation template in catalog <catalog.templates.classification.multi_class.relation.default>`.
+   * - Evaluate your question-answering dataset 
      - Demonstrates how to evaluate a user QA answering dataset in a standalone file using a user defined task and template.
      - `code <https://github.com/IBM/unitxt/blob/main/examples/standalone_qa_evaluation.py>`_
      - :ref:`Add new dataset tutorial <adding_dataset>`.
-   * - Evaluate a user defined open QA dataset - reusing existing catalog assets
-     - Demonstrates how to evaluate a user QA answering dataset using the predefined open qa task and templates.
+   * - Evaluate your question-answering dataset  - reusing existing catalog assets
+     - Demonstrates how to evaluate a user QA dataset using the predefined open qa task and templates.
        It also shows how to use preprocessing steps to align the raw input of the dataset with the predefined task field.
```
yoavkatz, line 31:
```suggestion
       It also shows how to use preprocessing steps to align the raw input of the dataset with the predefined task fields.
```

prepare/cards/wnli.py
```diff
@@ -11,24 +12,18 @@
...
     loader=LoadHF(path="glue", name="wnli"),
     preprocess_steps=[
-        "splitters.small_no_test",
-        RenameFields(
-            field_to_field={
-                "sentence1": "text_a",
-                "sentence2": "text_b",
-            }
+        SplitRandomMix(
+            {"train": "train[95%]", "validation": "train[5%]", "test": "validation"}
```
eladven, line 16: Isn't it the same as splitters.small_no_test? Why not using it?

eladven:
Hi. I understand the one-field processing. However I thought that the best practice is to use predefined artifacts if exists

prepare/cards/wnli.py
yoavkatz: See this issue #925.    I think these catalog splitters are not clearly named  and hence not understood by users , especially in initial examples.

Issue #924, opened 2024-06-20
Support for load_dataset for cards not in catalog


yoavkatz:
That when I have define a card , I need to use this to load it:
dataset = StandardRecipe(card=card, template_card_index=“simple”)().to_dataset()
but if it’s in the catalog I use
datasets = load_dataset(“card=cards.mycard,template_card_index=simple”)
which is much clearer.

I think it would be good to have an option to support both
datasets = load_dataset(card=card, template_card_index=“simple”)
and
datasets = load_dataset(“card=cards.mycard,template_card_index=simple”)

(After adding this ability, should also update all the examples)

PR #929, opened 2024-06-21
load_dataset supports loading cards not present in local catalog

pawelknes:
`unitxt.api.load_dataset` now also supports loading cards which are not present in local catalog

For #924


elronbandel (approved):
Neat!

Issue #912, opened 2024-06-17
New  HF datasets version causes unit test failures  when using  filtering_lambda


yoavkatz:
Localized an temporarily disabled test in test_loaders.

@unittest.skip("Currently this fails from datasets 2.20")
    def test_load_from_HF_multiple_innvocation_with_filter(self):
        loader = LoadHF(
            path="CohereForAI/aya_evaluation_suite",
            name="aya_human_annotated",
            filtering_lambda='lambda instance: instance["language"]=="eng"',
        )
        ms = loader.process()
        dataset = ms.to_dataset()
        self.assertEqual(
            list(dataset.keys()), ["test"]
        )  # that HF dataset only has the 'test' split
        self.assertEqual(dataset["test"][0]["language"], "eng")

        ms = loader.process()
        dataset = ms.to_dataset()
        self.assertEqual(
            list(dataset.keys()), ["test"]
        )  # that HF dataset only has the 'test' split
        self.assertEqual(dataset["test"][0]["language"], "eng")

dafnapension:
Hi, I am not sure I got the issue right, so I commented out line 210:
`# @unittest.skip("Currently this fails from datasets 2.20")`
and yet, the whole of `test_loaders.py` completed successfully:

```
(virtual39) dafna@DESKTOP-GM8R3J7:~/workspaces/unitxt$ git diff
diff --git a/tests/library/test_loaders.py b/tests/library/test_loaders.py
index ebc586c3..7b71c513 100644
--- a/tests/library/test_loaders.py
+++ b/tests/library/test_loaders.py
@@ -207,7 +207,7 @@ class TestLoaders(UnitxtTestCase):
         )  # that HF dataset only has the 'test' split
         self.assertEqual(dataset["test"][0]["language"], "arb")

-    @unittest.skip("Currently this fails from datasets 2.20")
+    # @unittest.skip("Currently this fails from datasets 2.20")
     def test_load_from_HF_multiple_innvocation_with_filter(self):
         loader = LoadHF(
             path="CohereForAI/aya_evaluation_suite",
(virtual39) dafna@DESKTOP-GM8R3J7:~/workspaces/unitxt$ python -m unittest tests.library.test_loaders
[Unitxt|INFO|loaders.py:123] 2024-06-23 10:18:02,115 >> LoadCSV sets 'data_classification_policy' to ['proprietary'] by default when loading from local files.
To use a different value or remove this message, explicitly set the `data_classification_policy` attribute of the loader.

[...]

PR #930, opened 2024-06-23
adapt filtering lambda to datasets 2.20

dafnapension:
Per issue #912


dafnapension:
Hi @elronbandel ,  I do not manage to reproduce in my dev env. Even after cleaning my whole huggingface cache.   That is: all my local runs end successfully.
What do I miss?

yoavkatz:
Maybe it was a local issue - I had?

dafnapension:
Thanks, @yoavkatz. Yes, that was my puzzle: why is the regression failing, whereas my dev-env is laughing healthily all the way through?  And douze points!  my hf is indeed `datasets==2.19.0`.  I must comment, though, that `base.eqr` reads:  `datasets>=2.16.0`
Trying now with `2.20`, as your comment for the skip reads.

dafnapension:
Hi @yoavkatz , did you notice this while upgrading to 2.20? perhaps this is one source of troubles?:

```
Installing collected packages: tqdm, requests, datasets
  Attempting uninstall: tqdm
    Found existing installation: tqdm 4.66.2
    Uninstalling tqdm-4.66.2:
      Successfully uninstalled tqdm-4.66.2
  Attempting uninstall: requests
    Found existing installation: requests 2.31.0
    Uninstalling requests-2.31.0:
      Successfully uninstalled requests-2.31.0
  Attempting uninstall: datasets
    Found existing installation: datasets 2.19.0
    Uninstalling datasets-2.19.0:
      Successfully uninstalled datasets-2.19.0
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
ibm-cos-sdk-core 2.13.4 requires urllib3<1.27,>=1.26.18; python_version < "3.10", but you have urllib3 2.2.1 which is incompatible.
Successfully installed datasets-2.20.0 requests-2.32.3 tqdm-4.66.4
```

src/unitxt/loaders.py
```diff
@@ -177,18 +177,13 @@ def verify(self):
...
                 f"{self.__class__.__name__} cannot run use filtering_lambda expression without setting unitxt.settings.allow_unverified_code=True or by setting environment variable: UNITXT_ALLOW_UNVERIFIED_CODE."
             )
         logger.info(f"\nLoading filtered by: {self.filtering_lambda};")
-        return MultiStream(
-            {
-                name: dataset[name].filter(eval(self.filtering_lambda))
-                for name in dataset
-            }
-        )
+        return dataset.filter(eval(self.filtering_lambda))
```
elronbandel, line 186: Should this line be protected by `allow_unverified_code` (https://github.com/IBM/unitxt/blob/6b1d93d2914e2551668c1803f59392411cee05c6/src/unitxt/settings_utils.py#L121)
@yoavkatz  can you advise?

elronbandel (changes requested):

src/unitxt/loaders.py
dafnapension: Hi @elronbandel , it is well protected, in line 181, a few lines above the line you commented about

elronbandel (approved):

PR #935, opened 2024-06-24
Add MultiTemplate

elronbandel:


src/unitxt/templates.py
```diff
@@ -122,7 +128,32 @@ def apply_formatting(
             ) from e
 
 
-class InputOutputTemplate(Template):
+class MultiTemplate(Template):
+    templates: List[BaseTemplate]
```
yoavkatz, line 132: Looks good. Please add inline documentation and please add short description to all templates typed in 

https://unitxt.readthedocs.io/en/latest/docs/adding_template.html

Issue #945, opened 2024-06-24
Fix Llama3 format


elronbandel:

oktie:
I removed the formats.llama3_chat_with_system_prompt system prompt that was not being used anywhere, and also wasn't right (e.g. had \n between parts instead of \n\n).

I've updated formats.llama3_chat definition so:

    the demos also have the user/assistant prompts. I believe that is how we are supposed to provide examples.
    added a default DEFAULT_SYSTEM_PROMPT that is one that is being suggested/used for llama2, e.g. see: https://developer.ibm.com/tutorials/awb-prompt-engineering-llama-2/

I've added a formats.llama3_instruct that I believe more accurately reflects the official guidance https://huggingface.co/blog/llama3#how-to-prompt-llama-3 which states

    The Instruct versions use the following conversation structure:

<|begin_of_text|><|start_header_id|>system<|end_header_id|>

{{ system_prompt }}<|eot_id|><|start_header_id|>user<|end_header_id|>

{{ user_msg_1 }}<|eot_id|><|start_header_id|>assistant<|end_header_id|>

{{ model_answer_1 }}<|eot_id|>

    This format has to be exactly reproduced for effective use.

Note: the instruction is not a part of the demos. I believe a demo example-specific instruction can improve the prompt, but I don't think it's possible the way unitxt templates work now (instruction cannot be used in demo_format).

PR #950, opened 2024-06-25
llama3 instruct and chat system prompts

oktie:
Issue #945 

I removed the `formats.llama3_chat_with_system_prompt` system prompt that was not being used anywhere, and also wasn't right (e.g. had `\n` between parts instead of `\n\n`).

I've updated `formats.llama3_chat` definition so:
1) the demos also have the `user`/`assistant` prompts. I believe that is how we are supposed to provide examples.
2) added a default `DEFAULT_SYSTEM_PROMPT` that is one that is being suggested/used for llama2, e.g. see: https://developer.ibm.com/tutorials/awb-prompt-engineering-llama-2/

I've added a `formats.llama3_instruct` that I believe more accurately reflects the official guidance https://huggingface.co/blog/llama3#how-to-prompt-llama-3 which states 
>The Instruct versions use the following conversation structure:
```
<|begin_of_text|><|start_header_id|>system<|end_header_id|>

{{ system_prompt }}<|eot_id|><|start_header_id|>user<|end_header_id|>

{{ user_msg_1 }}<|eot_id|><|start_header_id|>assistant<|end_header_id|>

{{ model_answer_1 }}<|eot_id|>
```
>This format has to be exactly reproduced for effective use.

Note: the `instruction` is not a part of the demos. I believe a demo example-specific instruction can improve the prompt, but I don't think it's possible the way unitxt templates work now (`instruction` cannot be used in `demo_format`).


eladven (commented):
Elron, The PR looks fine. Please add the json files and it will allow the consistency test to pass

yoavkatz (changes requested):
To be consistent. The system prompt needs to be defined outside. the format

prepare/formats/models/llama3.py
```diff
@@ -2,25 +2,50 @@
...
 format = SystemFormat(
-    demo_format="{source}\n\n{target_prefix}{target}\n\n",
-    model_input_format="<|begin_of_text|><|eot_id|><|start_header_id|>user<|end_header_id|>\n\n"
-    "{instruction}\\N{demos}{source}<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
-    "{target_prefix}",
+    demo_format="<|start_header_id|>user<|end_header_id|>\n\n"
+    "{instruction}{source}<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
+    "{target_prefix}{target}<|eot_id|>",
+    model_input_format="<|begin_of_text|><|start_header_id|>system<|end_header_id|>\n\n"
+    "{DEFAULT_SYSTEM_PRO#MPT}<|eot_id|>{demos}<|start_header_id|>user<|end_header_id|>"
```
yoavkatz, line 28: system_prompt should be taken from {system_prompt}, so it can be configured from the outside.

oktie: @yoavkatz yes, this is how it is for `llama3_instruct`. For `llama3_chat` it was defined without a system prompt in the past, and according to the guidelines that is not a good idea. So I thought it makes sense to add a default system prompt so if the user does not define one, we at least have a generic one. Does this make sense?

oktie: should we just put the system prompt there instead of using this `DEFAULT_SYSTEM_PROMPT` constant? I tested it on my side, and it worked, but I am not sure if it will work if unitxt is installed as a library.

yoavkatz: If the recommended approach is to have system_prompt, then llama3_chat should have system prompt, but this should be configurable via {system_prompt}.

By default, it should be format=formats.llama3_chat,system_prompt=system_prompt.llama3. 

 The users can always pass system_prompt.empty if they  don't want system prompt.
 
If there is any need to format without |system| special token, there should be a format which is llama3_chat_without_system_prompt.

yoavkatz: BTW, formats.llama3_chat_with_system_prompt was used in the internal project that uses unitxt.

yoavkatz: I think this should be the chat version:

{
    "__type__": "system_format",
    "demo_format": "<|start_header_id|>user<|end_header_id|>\n\n{instruction}{source}<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n{target_prefix}{target}<|eot_id|>",
    "model_input_format": "<|begin_of_text|><|start_header_id|>system<|end_header_id|>\n\n{system_prompt}<|eot_id|>{demos}<|start_header_id|>user<|end_header_id|>{source}<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n{target_prefix}"
}

In this case the demos really part of the chat.

and this should be the instruction version:

 {
    "__type__": "system_format",
    "demo_format": "{source}\n\n{target_prefix}{target}",
    "model_input_format": "<|begin_of_text|><|start_header_id|>system<|end_header_id|>\n\n{system_prompt}<|eot_id|><|start_header_id|>user<|end_header_id|>{demos}{source}<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n{target_prefix}"

In this case the demos are part of the single user input.  Instruct models don't support multiple turns, as far as I understand.

This is actually a good place to add an example to unitxt - on the different type of formatting.
https://unitxt.readthedocs.io/en/latest/docs/examples.html

yoavkatz: Do you agree? @oktie , @eladven , @elronbandel ?

oktie: @yoavkatz I think examples should include <user> and <assistant> roles. I couldn't find much resources online about this for llama3, but this guide for example says that roles should be specified: https://www.promptingguide.ai/models/code-llama#few-shot-prompting-with-code-llama

I also got better scores on my task with this prompt, but it was a small test on just one task. I can run more tests and report them here.

yoavkatz: Did you run against the instruct model or chat model or both?

If few shots are placed as user/assistant also in instruct models, what's the difference between instruct model prompting and chat model prompting?

oktie: @yoavkatz Llama 3 only has instruct models. There is no separate chat model. This is another reference for the prompt template for all llama3 models: https://github.com/meta-llama/llama-recipes?tab=readme-ov-file#llama-recipes-examples-to-get-started-using-the-llama-models-from-meta

I did a bit more testing and made a few changes based on my observations. I've pushed my changes on this branch, and I think it helped resolve some test errors as well.

For my new tests, instead of using my own task as I did before, I randomly picked an existing task, `cards.boolq.classification` and tried with and without the system prompt and llama3 format. Without a specific `system_prompt` for the task, the result of using the format is actually worse because the model acts more like a chatbot and could explain a yes/no answer or say "absolutely" instead of "yes". I added a new system prompt `system_prompts.boolqa` to guide the model to generate only yes/no, and with that, the results are better than not using the right prompt and format. Of course one could modify the `instruction` instead of modifying the system prompt, but again for llama3, the recommended way of doing it seems to be in the system prompt, and repeating some in `instruction` can further improve the outcome based on my (limited) results.

This is an area where I'm hoping the community can chime in later on and improve the formats as well as system prompts for different tasks.

arielge: @oktie @yoavkatz It seems that now the `instruction` was completely removed from the format. This is a mistake, right?

yoavkatz: I've added the instruction and default behavior to the system prompt.  I created under examples , simple code that we can all run and review the output:

python examples/evaluate_different_formats.py

```
source (str):
    <|begin_of_text|><|start_header_id|>system<|end_header_id|>
    
    You are a helpful, respectful and honest assistant. Always answer as helpfully as possible, while being safe.  Your answers should not include any harmful, unethica
    l, racist, sexist, toxic, dangerous, or illegal content. Please ensure that your responses are socially unbiased and positive in nature.
    
    
    If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a quest
    ion, please don't share false information.
    Answer the following questions in one word.<|eot_id|><|start_header_id|>user<|end_header_id|>
    
    Where is Spain?<|eot_id|><|start_header_id|>assistant<|end_header_id|>
    
    Europe<|eot_id|><|start_header_id|>user<|end_header_id|>
    
    When was IBM founded?<|eot_id|><|start_header_id|>assistant<|end_header_id|>
    
    1911<|eot_id|><|start_header_id|>user<|end_header_id|>
    
    What is the capital of Texas?<|eot_id|><|start_header_id|>assistant<|end_header_id|>

```

yoavkatz: template = InputOutputTemplate(
    instruction="Answer the following questions in one word.",
    input_format="{question}",
    output_format="{answers}",
)

dataset = load_dataset(
    card=card,
    template=template,
    format="formats.llama3_chat",
    system_prompt="system_prompts.models.llama2",
    num_demos=2,
    demos_pool_size=3,
)

oktie: Thanks @yoavkatz I was also working on a fix and testing it, but yours looks good to me, and I just tested it on a task and it works well (with llama3-70b-instruct I get a much better score with llama3 format).

The only thing is that now `llama3_chat` and `llama3_instruct` are the same. I suggest removing `llama3_chat` and updating all the references to it. Do you want me to do it?

yoavkatz: Yes. If they are the same we can keep just one. I just wonder about about the alternative format where the instruction and demos are in the same user prompt as the question.  Granite recommended icl format is this way.

oktie: @yoavkatz okay I just replaced `llama_chat` with `llama_instruct` anywhere it was used.

I also experimented with two alternatives. An `alt1` option where the instruction goes to the user role prompt, and an `alt2` option where the demos all go into one user prompt, as you had suggested. Numbers I get on two example tasks, each with `llama-3-70b-instruct` and 3-shot prompts:
| Task  | Option | Score |
| -------- | ------- | ------- |
| `cards.boolq.classification` |  `formats.llama3_instruct`    |  0.9119  |
| `cards.boolq.classification`  |   `formats.llama3_instruct_alt1`  |  0.9081 |
| `cards.boolq.classification`  |  `formats.llama3_instruct_alt2`   |  0.89 |
| `cards.summarize_from_human_feedback` |  `formats.llama3_instruct`    |  0.68 |
| `cards.summarize_from_human_feedback`  |   `formats.llama3_instruct_alt1`  |  0.67 |
| `cards.summarize_from_human_feedback`  |  `formats.llama3_instruct_alt2`   |  0.56  |

So it seems like `alt2` is not making much of a difference, but `alt2` is worse. I believe we have the right prompt now. I suggest approving this PR and merging for now. I'll be doing more testing and will let you know if I find ways to improve the format,

yoavkatz (approved):
I approved.  Make sure the old jsons are deleted. 
Also, given the results maybe we don't need alt1 and alt2 in the catalog.  If we keep then, let's give them descriptive names (e.g formats.llama3_instruct_all_demos_in_a_single_turn".)

PR #953, opened 2024-06-26
adding the end2end rag task

assaftibm:


prepare/metrics/rag.py
```diff
@@ -416,3 +416,144 @@
     add_to_catalog(
         metric, f"metrics.rag.response_generation.{axis}.{base_metric}", overwrite=True
     )
+
+#############################################
```
matanor, line 420: i think it would be nice to have any new answer correctness and context correctness metrics in these modules:
https://github.com/IBM/unitxt/blob/main/prepare/metrics/rag_answer_correctness.py
https://github.com/IBM/unitxt/blob/main/prepare/metrics/rag_context_correctness.py

splitting up the metrics across modules makes things easier to follow, and allows reuse of the testing code that exists in the above modules.

prepare/metrics/rag.py
```diff
@@ -416,3 +416,144 @@
...
+#### End to end rag metrics #################
+#############################################
+
+end_to_end_artifact_names = [
+    TaskRagEndToEndConstants.METRICS_RAG_END_TO_END_ANSWER_CORRECTNESS,
```
matanor, line 425: Can these names be shortened?

For example, consider:
`RagEndToEnd.ANSWER_CORRECTNESS`
so `RagEndToEnd` is not repeated twice, being a constant is implied from the use of `UPPER_CASE`, `Task` is implied from `RagEndToEnd` and `Metric` is implied from `ANSWER_CORRECTNESS`.

Would makes things much more readable IMO.. WDYT?

yoavkatz:
Closing this issue, as we handled in different PRs.

PR #955, opened 2024-06-26
Update adding_metric.rst

yoavkatz:
Added description of handling multiple metrics.


elronbandel (approved):
Important

yoavkatz:
> Important

Yes. I think we should turn every user question to a doc update when it's not available.

PR #957, opened 2024-06-26
Fix inference/production api to support invocation by task without specifying a card

elronbandel:
From now on you will be able to use `produce` and `infer` to process data by specifying only the task your data points complying to. 

This means you will be able to use any task in the Unitxt catalog as an inference function. 

Key contributions:

1. No need to assign values to output fields in infer/produce functions (only to input fields as the task specify)
2. No need to specify a card unless few shot examples are needed. You can specify the task and it will be enough. 

See the full production.rst for details


docs/docs/production.rst
```diff
@@ -63,6 +63,18 @@ Then, produce the model-ready input data with the `produce` function:
 
   result = produce(instance, recipe)
 
+If we want to add to the data few shot examples, we should specify recipe with a data card:
+
+.. code-block:: python
+
+  recipe = "card=cards.wnli,template=templates.classification.multi_class.relation.default,demos_pool_size=5,num_demos=2"
```
OfirArviv, line 70: I don't understand this example. 
Why are you giving the card of wnli? Will it use the example from there?

prepare/processors/to_string.py
```diff
@@ -1,25 +1,18 @@
 from unitxt import add_to_catalog
 from unitxt.blocks import ToString, ToStringStripped
-from unitxt.operator import SequentialOperator
+from unitxt.processors import PostProcess
+from unitxt.settings_utils import get_constants
+
+constants = get_constants()
```
OfirArviv, line 6: where is this used? It's in all the files but never get used

src/unitxt/api.py
```diff
@@ -92,7 +92,7 @@ def evaluate(predictions, data) -> List[Dict[str, Any]]:
 
 
 def post_process(predictions, data) -> List[Dict[str, Any]]:
```
OfirArviv, line 94: what is this function aim? It's not clear what it does and how the change reflect it

OfirArviv: is it only for inference? What does it means?

docs/docs/production.rst
elronbandel: yes the few shot examples come from that card.

prepare/processors/to_string.py
elronbandel: what is what used?

src/unitxt/api.py
elronbandel: This function is for the api to use only post processing on a data without the rest of the pipeline.

src/unitxt/processors.py
```diff
@@ -114,7 +141,7 @@ def process_value(self, text: Any) -> Any:
         return non_empty_line
 
 
-class LowerCase(FieldOperator):
+class Lower(FieldOperator):
```
eladven, line 144: This is a breaking change. We should use gentle deprecation.

src/unitxt/schema.py
```diff
@@ -22,6 +25,38 @@
...
+    {
+        "source": Value("string"),
+        "metrics": Sequence(Value("string")),
+        "groups": Sequence(Value("string")),
+        "subset": Sequence(Value("string")),
+        "postprocessors": Sequence(Value("string")),
+        "task_data": Value(dtype="string"),
+        "data_classification_policy": Sequence(Value("string")),
+    }
+)
```
eladven, line 51:
```suggestion
```

src/unitxt/settings_utils.py
```diff
@@ -87,6 +88,17 @@ def get_all_environment_variables(self):
...
+    def context(self, **kwargs):
+        old_values = {key: self._settings.get(key, None) for key in kwargs}
+        try:
+            for key, value in kwargs.items():
+                self.__setattr__(key, value)
+            yield
+        finally:
+            for key, value in old_values.items():
+                self.__setattr__(key, value)
+
```
eladven, line 101:
```suggestion
```

It should be in a separate PR

eladven (approved):

PR #961, opened 2024-06-28
Added support for ibm-watsonx-ai inference

pawelknes:
Added support for WML inference using `ibm-watsonx-ai`


pawelknes:
I had to remove `ibm-watsonx-ai` from requirements as the lowest available release supports Python version only down to 3.9 and it causes conflicts with some other packages during installation anyway. So the viable solution for a user would be to install the package on their own and use Python >=3.10 (I've included that information in `_requirement` of `WMLInferenceEngine`)

yoavkatz (approved):

PR #965, opened 2024-06-30
Add example of using LLM as a judge for summarization dataset.

eladven:


elronbandel (approved):
Overall looks good. Yet do we really want users to write templates for each task being evaluated? I think a better model is that we will guide them to use something like:
"card=cards.xsum,metrics=[metrics.llm_as_judge]"

yoavkatz:
> Overall looks good. Yet do we really want users to write templates for each task being evaluated? I think a better model is that we will guide them to use something like: "card=cards.xsum,metrics=[metrics.llm_as_judge]"

Yes. I agree. We already have an example of adding an llm metric definition and using it.  Here , I think we should use a simple predefined metric.  We should show one metric that uses the reference answr and ibe that does not.

PR #968, opened 2024-06-30
Add simple LLM as a judge example, of using it without installaiotn

eladven:


docs/docs/examples.rst
```diff
@@ -33,13 +33,19 @@ Each example is a self contained python file that you can run and later modify.
      - | :ref:`Add new dataset tutorial <adding_dataset>`.  
        | :ref:`Open QA task in catalog <catalog.tasks.qa.open>`.
        | :ref:`Open QA template in catalog <catalog.templates.qa.open.title>`.
+   * - Evaluate an existing question-answering dataset from the Unitxt catalog, and evaluate it - using LLM as a judge
```
yoavkatz, line 36: Not sure if we want to recommend the no_installation required option too much, because it slower (every load of a dataset requires installation).  Also, it's harder to extend it with new capabilities.

Issue #958, opened 2024-06-27
Fix MRR wiring


assaftibm:
The underlying implementation has changed, and now it requires the reference to be provided as List[List[str]] rather than just List[str].
We want to expose the metric in its standard definition, where the reference is List[str]. 
@benjaminsznajder noticed this and can give more details.

matanor:
Suggested fix from @yoavkatz :

```
 from unitxt.collections_operators import Wrap
metric = MetricPipeline(
        main_score=“score”,
        preprocess_steps=[
            Copy(field=“context_ids”, to_field=“prediction”),
            Wrap(field=“ground_truths_context_ids”, inside=“list”, to_field=“references”),
        ],
        metric=f”metrics.{metric_name}“,
    )
```

PR #969, opened 2024-06-30
fix MRR wiring

matanor:
Closes https://github.com/IBM/unitxt/issues/958

- Fix wiring of `context_correctness` RAG metrics (MRR and MAP): allow the `context_ids` to be a list of strings, instead of a `list[list[str]]`. This allows directly passing the list of predicted context ids, as was done in unitxt version 1.7. 
- Added corresponding tests.

Details:
- Moved the preparation of the `context_correctness` metrics to a new file `rag_context_correctness.py`. I hope that's ok. It makes development and testing easier, and these metrics are independent of the other rag metrics defined in `rag.py`.
- The actual fix is based on @yoavkatz 's suggestion here: https://github.com/IBM/unitxt/issues/958#issuecomment-2198482606.
- Added tests (there were none):
  - A test over the `evaluate` API, for that added a new util `test_evaluate`, and also refactored the scores comparison code into a separate method `check_scores`.
  - A test using the usual `test_metric` function.


yoavkatz:
@eladven will try to help tomorrow.

PR #971, opened 2024-06-30
add LlamaIndex faithfulness metric

arielge:


src/unitxt/metrics.py
```diff
@@ -2254,6 +2251,53 @@ def compute(
         }
 
 
+class LlamaIndexFaithfulness(LlamaIndexCorrectness):
```
yoavkatz, line 2254: There is a lot of duplicated code between the metrics. Can you please create a common base class  (e.g. LlamaIndexMetric), and more the code there?

yoavkatz (changes requested):
Once change requested regarding refactoring common code - see details in the comment.

src/unitxt/metrics.py
arielge: done

src/unitxt/metrics.py
```diff
@@ -2247,11 +2248,37 @@ def compute(
...
-            # "feedback": result.feedback, # removed since this cannot be tested
-        }
+        return {self.main_score: result / 5}
+
+
+class LlamaIndexFaithfulness(LlamaIndexLLMMetric):
+    """LlamaIndex based metric class for evaluating faithfulness."""
+
+    metric_name_prefix = "faithfulness"
+    reference_field = "contexts"  # metric doesn't require reference answers
```
yoavkatz, line 2258: Is this needed? I don't see it being used.

arielge: As this is a referenceless metric, this is there to prevent failures of `InstanceMetric` where no `"references"` field was provided

src/unitxt/metrics.py
```diff
@@ -2134,11 +2134,10 @@ def compute(
         return self.pipe(predictions, batch_size=self.batch_size)
 
 
-class LlamaIndexCorrectness(InstanceMetric):
-    """LlamaIndex based metric class for evaluating correctness."""
-
+class LlamaIndexLLMMetric(InstanceMetric):
     model_name: str = ""
     main_score: str = ""
+    metric_name_prefix: str = ""
```
yoavkatz, line 2140: There is now a generic score_prefix in Metric base class.  Maybe it can be used, instead of adding this new name ?

#
    # Used to add a prefix to all score, except the "score_name" and "score" fields.
    # This is used to distinguish two scores of the same metrics, operating on different fields of the task
    #
    score_prefix: str = ""

src/unitxt/metrics.py
arielge: (https://github.com/IBM/unitxt/blob/1242b022919d43716b3f26d67e3c75f262485900/src/unitxt/metrics.py#L877)

src/unitxt/metrics.py
arielge: Sure, I wasn't aware of it

src/unitxt/metrics.py
yoavkatz: Thanks for the quick replies on the CR comments. I see.  But a "references" are always passed in unitxt - even for reference-less metrics.  How did you get a failure?

arielge: Right, that is a different discussion that is outside the scope of this pull request... So if you want I can remove this

yoavkatz: Yes. For now, please do (it will avoid confusion).

arielge: No problem. I will say on a more general note that as unitxt provides the possibility for users to use just specific parts of the library (in this case, maybe I want to use a unitxt metric without using the full unitxt pipeline), these sorts of questions might need more thought (cc @elronbandel)

yoavkatz (approved):

yoavkatz:
I approved - waiting for test to pass and auto merge.

PR #974, opened 2024-07-01
added num_proc parameter for parallel data loading

csrajmohan:
This adds support to pass an optional `num_proc` parameter in Unitxt loader so that it can be passed onto `hf_load_dataset` method of Huggingface Datasets library for parallel dataset loading.

p.s.: The num_proc parameter is used to specify the number of processes to use for multiprocessing in the Hugging Face datasets library. This can speed up operations while loading large datasets.
(Tested with wikitq dataset)


csrajmohan:
Added to `LoadHF` right now, but can also be added to few other Unitxt loaders like `LoadFromIBMCloud, LoadFromSklearn, LoadFromKaggle` that invoke `hf_load_dataset` if needed.

elronbandel (approved):

elronbandel (approved):

PR #991, opened 2024-07-03
try to deep as little as possible

dafnapension:
Until that metric that modifies 'predictions' is crucified, an attempt to reduce the damage it causes for performance.
My hope is that the removal of `instace=deepcopy(instance)` from `FieldOperator`, which is executed also for each `Rename`, for example, will bear significant fruits, in terms of performance.

### DIFF
--- a/src/unitxt/dict_utils.py
+++ b/src/unitxt/dict_utils.py
@@ -1,4 +1,5 @@
 import re
+from copy import deepcopy
 from typing import Any, List, Tuple
 
 from .text_utils import construct_dict_str
@@ -225,11 +226,11 @@ def dict_delete(
 # if query includes * then return a list of values reached by all paths that match the query
 # flake8: noqa: C901
 def get_values(
-    current_element: Any, query: List[str], index_into_query: int
+    current_element: Any, query: List[str], index_into_query: int, deep_copy: bool
 ) -> Tuple[bool, Any]:
     # going down from current_element through query[index_into_query].
     if index_into_query == 0:
-        return (True, current_element)
+        return (True, current_element if not deep_copy else deepcopy(current_element))
 
     # index_into_query < 0
     component = query[index_into_query]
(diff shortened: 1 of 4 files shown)


yoavkatz:
I think indeed the problem is not with copy operator (which did a deep copy even before), but with other operators (e.g. ListFieldValues) which do not do a copy.  I need to recreate the original scenario (it happened in RAG), but in general, I think that if we assign a field to a field , it deep copy the value (unless it's immutable like a string or number) - otherwise a user will not be able to follow and debug this (it's hard enough to deal with this in pure python code, but following unitxt stream flow, is much harder).

We  don't need to copy the full instance as done now in main branch - which is wasteful.

dafnapension:
Hi @yoavkatz , do you refer to [this rag](https://github.com/IBM/unitxt/blob/main/prepare/metrics/rag.py)?

May I use the opportunity..:
Why do you copy the scores at the end, and 'apologize' that copy does not remove the source? why not use `Raname` operator instead?

And I see the problem in ListFieldValues, indeed. This is a copy that we do ourselves, not the user, and we should be careful. 

And where is the point that **RAG actually changes  predictions**?

yoavkatz:
@dafnapension - to fully recreate it in real scenario we have a run example in internal code.  @elronbandel will look at it and then map to the unitxt fix.

dafnapension:
Hi @yoavkatz , is there still a performance issue, that you would like to look into (issue 987)?
I think that the `deepcopy` that is carried out on each and every `FieldOperator` , **on the whole input stream** (not only on those instances that go to metrics) is a high tax to pay for a bandage around (unlawful) metrics.  Are we ready to remove that `deepcopy` (of a whole instance) from the `FieldOperator`?

dafnapension:
covered by https://github.com/IBM/unitxt/pull/1087  which fixed all needed for the violating metrics, without deepcopy of every instance passing through every FieldInstanceOperator, and without deepcopying in between metrics, in ApplyMetrics

PR #993, opened 2024-07-04
1.10.3

eladven:


elronbandel (changes requested):
Can you change the name of the python classes of those metrics rather then reverting this change? 
Safety -> SafetyMetric
Regard -> RegardMetric

Issue #938, opened 2024-06-24
Replace confusing names of inputs and outputs of Task


yoavkatz:
Today tasks have "inputs" and "outputs".  This is confusing many users, when actually it should be named:

"input_fields" - used to create the input 
"reference_fields" - used to create references

PR #994, opened 2024-07-04
Chore/rename task fields

luisaadanttas:
This pull request addresses issue #938


yoavkatz:
Thanks for your contribution Maria! The changes look right.  However,   the proposed change is not backward compatible (require changes to user code). Since we don't want to make abrupt changes, we should keep the original fields in Task , but add them as deprecated.  Then we should check in the Task. prepare() that either the old fields or new fields are set (but not both).  It's good that you changed all the existing documentation and examples, so people will follow the new approach from now on. Finally, we should add one test that uses the old fields and see it continues works.

Do you think you can proceed with these changes.

luisaadanttas:
Yes, that makes a lot of sense. Thank you!

I can implement those changes :)

elronbandel:
@luisaadanttas thanks again! notice you mark the old fields with DeprectedField.
See this example for reference:
https://github.com/IBM/unitxt/blob/279412f6650eb69a4a9b7769100b65a31453ed7f/src/unitxt/operators.py#L257

src/unitxt/task.py
```diff
@@ -38,20 +39,49 @@ class Task(InstanceOperator):
...
+
+        self.input_fields = (
+            self.input_fields if self.input_fields is not None else self.inputs
+        )
+        self.reference_fields = (
+            self.reference_fields if self.reference_fields is not None else self.outputs
+        )
+
     def verify(self):
         for io_type in ["input_fields", "reference_fields"]:
```
yoavkatz, line 78: We should check that input_fields and reference_fields are not None (because they are optional fields).
Best to check each on separately:

 if (self.input_fields is   None):
            raise ValueError(
                "Missing attribute in task attributes 'input_fields' not set."
            )
 if (self.reference_fields is   None):
            raise ValueError(
                "Missing attribute: 'reference_fields' not set."
            )

src/unitxt/task.py
```diff
@@ -38,20 +39,49 @@ class Task(InstanceOperator):
...
+        },
+    )
     metrics: List[str]
     prediction_type: Optional[str] = None
     augmentable_inputs: List[str] = []
     defaults: Optional[Dict[str, Any]] = None
 
+    def prepare(self):
+        super().prepare()
+        if (self.input_fields is not None and self.reference_fields is not None) and (
```
yoavkatz, line 63: I suggest separating into two checks, because people can mix only one pair of fields.

if (self.input_fields is not None and self.inputs is not None):
  aise ValueError(
                "Conflicting attributes: 'input_fields'  cannot be set simultaneously with 'inputs'. Use only 'input_fields'"
            )

yoavkatz (changes requested):
Thanks for making the changes. I added two minor comments regarding error checking - and then I think we are good to merge.

luisaadanttas:
> Thanks for making the changes. I added two minor comments regarding error checking - and then I think we are good to merge.

Thanks for your review and suggestions! I've incorporated your feedback on the field checks and updated the tests accordingly :)

yoavkatz (approved):

yoavkatz:
Thanks for making the changes.  I think you need to run:
utils/prepare_all_artifacts.py to update a few json files that were not converted.

yoavkatz:
Changes are merged.

Thank you Maria for your contribution!

PR #1000, opened 2024-07-07
draft:  indicating that Helm uses code from its own cache rather than the PR-ed code. trying to fix 11.0 errors

dafnapension:
hi @elronbandel , it seems that helm uses code from its cache, and does not test the new code that is PR-ed


elronbandel:
I don't see how your changes can solve it? To me the problem might be that we do not install unitxt in the helm test which eliminate any cache

dafnapension:
My code fixed the several CopyFields->Copy, and yet, test Helm found one of them, in operators.py that it used from its own cache as clearly stated on its error logs. One that I did fix to Copy.

dafnapension:
The cause of the PR - fixed CopyField->Field was already solved otherwise. This PR was only kept to draw the attention of @elronbandel to the fact that on Git tests, Helm uses code (operators.py) from its own cache, rather than the PR-ed code.

PR #1005, opened 2024-07-08
Solve problem with striping format at LLM as a judge code.

eladven:


src/unitxt/artifact.py
```diff
@@ -305,8 +305,9 @@ def save(self, path):
 
     @classmethod
     def deserialize(cls, artifact_rep):
```
yoavkatz, line 307: Worthwhile to document"

"Recieves a representation (either json string or dictionary)  of an Artifact and return the object."

(Maybe with while to check that it's not already an Artifact)

src/unitxt/artifact.py
```diff
@@ -432,17 +433,18 @@ def __str__(self):
 def fetch_artifact(artifact_rep) -> Tuple[Artifact, Union[Artifactory, None]]:
     if isinstance(artifact_rep, Artifact):
         return artifact_rep, None
-    if Artifact.is_artifact_file(artifact_rep):
+    if isinstance(artifact_rep, str) and Artifact.is_artifact_file(artifact_rep):
```
yoavkatz, line 436: IIUC, this code is used to load an artifact from string name "tasks.classification.multi_class" and the code below is if it's "tasks.classification.multi_class[metrics=[....]].  Can remove this line and let the code below handle both cases?

Issue #1008, opened 2024-07-09
Exponential time complexity when confidence interval calculation is enabled


jezekra1:
The code below will call a full-dataset metric computation for each row - i.e. the work is duplicated num_row times.

This is a big issue for huggingface metrics, such as `metrics.meteor` which will run for 3 hours instead of 10 seconds.

https://github.com/IBM/unitxt/blob/cd3e483e33d1bb36b602a9ebdcc9ccf3030395d8/src/unitxt/metrics.py#L420

You can replicating it by running `metrics.meteor` on a large dataset with confidence interval computation enabled.

elronbandel:
@dafnapension can you have a look if you know how to fix?

dafnapension:
Hi @elronbandel , and thanks for assigning this to me. An interesting issues.
I have a few questions..
(1) When we tried to be "iber_khukhem", and compute global metrics by hand, aiming at one pass over the instances, @yoavkatz claimed that we are talking a few hundred instances, and this is why we can easily accommodate them all, conveniently in main memory, and whats the big deal. That the number of instances for which metric is computed is dramatically smaller than the number of instances started their way on the unitxt pipeline.
**Is this assumption still valid?**   How many instances participate in `meteor` 's **metric computation**? Can you tell? 
(2) by definition, confidence interval is computed by generating `n_resamples` , each the length of the number of instances over which the metric is computed (and hence assumed to be a few hundred, by above),  as clearly explained in lines 404-405 [https://github.com/IBM/unitxt/blob/cd3e483e33d1bb36b602a9ebdcc9ccf3030395d8/src/unitxt/metrics.py#L404](https://github.com/IBM/unitxt/blob/cd3e483e33d1bb36b602a9ebdcc9ccf3030395d8/src/unitxt/metrics.py#L404)
https://github.com/IBM/unitxt/blob/cd3e483e33d1bb36b602a9ebdcc9ccf3030395d8/src/unitxt/metrics.py#L404

yoavkatz:
I'm not sure there is any bug in the code.   I believe the problem is that when a a single metric calculation of global metrics takes a few seconds, it is multiplied by 100 when using confidence internal.

I'm not sure why Meteor slow,I think the Rouge reason it is slow , is that in itself it does bootstrapping in the code:

https://huggingface.co/spaces/evaluate-metric/rouge/blob/main/rouge.py#L134

I think the solution is to avoid use aggregation, which I'm working on.

[...]

PR #1011, opened 2024-07-09
Refactor Rouge and Meteor to InstanceMetric for faster score computation

yoavkatz:
added a new metric with interval calculations

### DIFF
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -1734,7 +1734,7 @@ class F1MacroMultiLabel(F1MultiLabel):
     average = None
 
 
-class Rouge(HuggingfaceBulkMetric):
+class Rouge(HuggingfaceInstanceMetric):
     hf_metric_name = "rouge"
     main_score = "rougeL"
     scale = 1.0
@@ -1765,17 +1765,22 @@ class Rouge(HuggingfaceBulkMetric):
         nltk.download("punkt")
         self.sent_tokenize = nltk.sent_tokenize
 
-    def compute(self, references, predictions, task_data: List[Dict]):
+    def compute(self, references, prediction, task_data: List[Dict]):
+        # for a single instance, prediction is of type str, and references: list of str
         if self.sent_split_newline:
-            predictions = [
-                "\n".join(self.sent_tokenize(prediction.strip()))
-                for prediction in predictions
-            ]
+            prediction = "\n".join(self.sent_tokenize(prediction.strip()))
+
             references = [
-                ["\n".join(self.sent_tokenize(r.strip())) for r in reference]
+                "\n".join(self.sent_tokenize(reference.strip()))
                 for reference in references
             ]
-        return super().compute(references, predictions, task_data)
+
+        hf_score = super().compute(references, prediction, task_data)
+        for metric_field in self.hf_metric_fields:
+            if isinstance(hf_score[metric_field], list):
+                assert len(hf_score[metric_field]) == 1
+                hf_score[metric_field] = hf_score[metric_field][0]
+        return hf_score
 
 
 # Computes char edit distance, ignoring whitespace
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -1699,7 +1699,6 @@ class Rouge(HuggingfaceMetric):
     prediction_type = "str"
     single_reference_per_prediction = False  # multiple references allowed
 
-    use_aggregator: bool = True
     rouge_types: List[str] = ["rouge1", "rouge2", "rougeL", "rougeLsum"]
 
     sent_split_newline: bool = True
@@ -1709,8 +1708,10 @@ class Rouge(HuggingfaceMetric):
     def prepare(self):
         super().prepare()
 
+        # We don't use the aggregation, to avoid running bootstrapping by the
+        # internal library (which is costly) and done by Unitxt in any case.
         self.hf_compute_args.update(
-            {"use_aggregator": self.use_aggregator, "rouge_types": self.rouge_types}
+            {"use_aggregator": False, "rouge_types": self.rouge_types}
         )
 
         import nltk
@@ -1728,7 +1729,10 @@ class Rouge(HuggingfaceMetric):
                 ["\n".join(self.sent_tokenize(r.strip())) for r in reference]
                 for reference in references
             ]
-        return super().compute(references, predictions, task_data)
+        results = super().compute(references, predictions, task_data)
+        for rouge_type in self.rouge_types:
+            results[rouge_type] = mean(results[rouge_type])
+        return results
 
 
 # Computes char edit distance, ignoring whitespace
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -328,10 +328,6 @@ class MetricWithConfidenceInterval(Metric):
             #   that is, re-form the groups, calculate the function, and take the mean of the group scores
             aggregation_func = self.average_item_scores
 
-        method = "BCa"
-        if len(instances) > 100:
-            method = "percentile"
-
         for score_name in score_names:
             # If all computed instance level scores are the same, there is no point in computing
             # confidence intervals. So skip to the next score.
@@ -357,7 +353,6 @@ class MetricWithConfidenceInterval(Metric):
                 statistic=statistic,
                 n_resamples=self.n_resamples,
                 confidence_level=self.confidence_level,
-                method=method,
                 random_state=self.new_random_generator(),
             ).confidence_interval
             full_score_name = ci_score_prefix + score_name
@@ -443,10 +438,6 @@ class MetricWithConfidenceInterval(Metric):
         if self._can_compute_confidence_intervals(num_predictions=num_predictions):
             identifiers = list(range(num_predictions))
 
-            method = "BCa"
-            if num_predictions > 100:
-                method = "percentile"
-
             with warnings.catch_warnings():
                 # Avoid RuntimeWarning in bootstrap computation. This happens on small datasets where
                 # the value of the computed global metric is the same on all resamplings.
@@ -456,7 +447,6 @@ class MetricWithConfidenceInterval(Metric):
                     statistic=statistic,
                     n_resamples=self.n_resamples,
                     confidence_level=self.confidence_level,
-                    method=method,
                     random_state=random_gen,
                 ).confidence_interval
             result["score_ci_low"] = ci.low
@@ -1342,16 +1332,46 @@ class HuggingfaceInstanceMetric(InstanceMetric):
         return score
 
 
-class Meteor(HuggingfaceInstanceMetric):
+class Meteor(InstanceMetric):
     main_score = "meteor"
     ci_scores = ["meteor"]
     reduction_map = {"mean": ["meteor"]}
     prediction_type = "str"
 
-    hf_metric_name = "meteor"
-    hf_metric_fields = ["meteor"]
-    hf_compute_args = {"alpha": 0.9, "beta": 3, "gamma": 0.5}
-    # the defaults in hf's Meteor
+    _requirements_list: List[str] = ["nltk"]
+    alpha: float = 0.9
+    beta: int = 3
+    gamma: float = 0.5
+    # unitxt uses nltk version >= 3.8
+
+    def prepare(self):
+        import nltk
+
+        nltk.download("wordnet")
+        nltk.download("omw-1.4")
+
+    def verify(self):
+        import importlib.metadata as importlib_metadata
+
+        from datasets.config import version
+
+        nltk_version = version.parse(importlib_metadata.version("nltk"))
+        assert nltk_version >= version.Version(
+            "3.6.6"
+        ), "nltk version must be at least 3.6.6"
+
+    def compute(self, references, prediction, task_data):
+        from nltk import word_tokenize
+        from nltk.translate import meteor_score
+
+        score = meteor_score.meteor_score(
+            [word_tokenize(ref) for ref in references],
+            word_tokenize(prediction),
+            alpha=self.alpha,
+            beta=self.beta,
+            gamma=self.gamma,
+        )
+        return {"meteor": score}
 
 
 class F1(GlobalMetric):
@@ -1745,7 +1765,46 @@ class F1MacroMultiLabel(F1MultiLabel):
     average = None
 
 
-class Rouge(HuggingfaceInstanceMetric):
+class Rouge(InstanceMetric):
+    main_score = "rougeL"
+    prediction_type = "str"
+    single_reference_per_prediction = False  # multiple references allowed
+    rouge_types: List[str] = ["rouge1", "rouge2", "rougeL", "rougeLsum"]
+    reduction_map = {"mean": ["rouge1", "rouge2", "rougeL", "rougeLsum"]}
+    ci_scores = ["rouge1", "rouge2", "rougeL", "rougeLsum"]
+
+    sent_split_newline: bool = True
+    _requirements_list: List[str] = ["nltk", "rouge_score"]
+
+    def compute(self, references: List[Any], prediction: Any, task_data: Dict) -> dict:
+        import nltk
+        from rouge_score import rouge_scorer
+
+        nltk.download("punkt")
+        self.sent_tokenize = nltk.sent_tokenize
+
+        # for a single instance, prediction is of type str, and references: list of str
+        if self.sent_split_newline:
+            prediction = "\n".join(self.sent_tokenize(prediction.strip()))
+
+            references = [
+                "\n".join(self.sent_tokenize(reference.strip()))
+                for reference in references
+            ]
+
+        # the following is taken from HF rouge, using the defaults:
+        # use_aggregator=True, use_stemmer=False, tokenizer=None
+        scorer = rouge_scorer.RougeScorer(
+            rouge_types=self.rouge_types, use_stemmer=False, tokenizer=None
+        )
+        # with Unitxt, references is a list
+        score = scorer.score_multi(references, prediction)
+        for key in score:
+            score[key] = score[key].fmeasure
+        return score
+
+
+class RougeHF(HuggingfaceInstanceMetric):
     hf_metric_name = "rouge"
     main_score = "rougeL"
     scale = 1.0
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -1347,8 +1347,13 @@ class Meteor(InstanceMetric):
     def prepare(self):
         import nltk
 
-        nltk.download("wordnet")
-        nltk.download("omw-1.4")
+        nltk.download("wordnet", quiet=True)
+        nltk.download("omw-1.4", quiet=True)
+        from nltk import word_tokenize
+        from nltk.translate import meteor_score
+
+        self.word_tokenize = word_tokenize
+        self.meteor_score = meteor_score
 
     def verify(self):
         import importlib.metadata as importlib_metadata
@@ -1361,12 +1366,9 @@ class Meteor(InstanceMetric):
         ), "nltk version must be at least 3.6.6"
 
     def compute(self, references, prediction, task_data):
-        from nltk import word_tokenize
-        from nltk.translate import meteor_score
-
-        score = meteor_score.meteor_score(
-            [word_tokenize(ref) for ref in references],
-            word_tokenize(prediction),
+        score = self.meteor_score.meteor_score(
+            [self.word_tokenize(ref) for ref in references],
+            self.word_tokenize(prediction),
             alpha=self.alpha,
             beta=self.beta,
             gamma=self.gamma,
@@ -1776,13 +1778,16 @@ class Rouge(InstanceMetric):
     sent_split_newline: bool = True
     _requirements_list: List[str] = ["nltk", "rouge_score"]
 
-    def compute(self, references: List[Any], prediction: Any, task_data: Dict) -> dict:
+    def prepare(self):
         import nltk
         from rouge_score import rouge_scorer
 
-        nltk.download("punkt")
+        self.rouge_scorer = rouge_scorer
+
+        nltk.download("punkt", quiet=True)
         self.sent_tokenize = nltk.sent_tokenize
 
+    def compute(self, references: List[Any], prediction: Any, task_data: Dict) -> dict:
         # for a single instance, prediction is of type str, and references: list of str
         if self.sent_split_newline:
             prediction = "\n".join(self.sent_tokenize(prediction.strip()))
@@ -1794,7 +1799,7 @@ class Rouge(InstanceMetric):
 
         # the following is taken from HF rouge, using the defaults:
         # use_aggregator=True, use_stemmer=False, tokenizer=None
-        scorer = rouge_scorer.RougeScorer(
+        scorer = self.rouge_scorer.RougeScorer(
             rouge_types=self.rouge_types, use_stemmer=False, tokenizer=None
         )
         # with Unitxt, references is a list
@@ -1832,7 +1837,7 @@ class RougeHF(HuggingfaceInstanceMetric):
 
         import nltk
 
-        nltk.download("punkt")
+        nltk.download("punkt", quiet=True)
         self.sent_tokenize = nltk.sent_tokenize
 
     def compute(self, references, prediction, task_data: List[Dict]):
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -1300,6 +1300,49 @@ class HuggingfaceBulkMetric(BulkInstanceMetric):
         return results
 
 
+class HuggingfaceInstanceMetric(InstanceMetric):
+    hf_metric_name: str
+
+    hf_metric_fields: List[str]
+    hf_compute_args: dict = {}
+
+    def prepare(self):
+        super().prepare()
+        self.metric = evaluate.load(
+            self.hf_metric_name, experiment_id=str(uuid.uuid4())
+        )
+
+    def compute(self, references: List[Any], prediction: Any, task_data: Dict) -> dict:
+        # invokes  module.compute, which invokes, e.g., meteor's _compute
+
+        try:
+            score = self.metric.compute(
+                predictions=[prediction],
+                references=[references],
+                **self.hf_compute_args,
+            )
+        except:
+            score = {self.main_score: np.nan}
+
+        if self.hf_metric_fields is not None and len(self.hf_metric_fields) > 0:
+            to_ret = {field: score[field] for field in self.hf_metric_fields}
+            score = to_ret
+
+        return score
+
+
+class Meteor(HuggingfaceInstanceMetric):
+    main_score = "meteor"
+    ci_scores = ["meteor"]
+    reduction_map = {"mean": ["meteor"]}
+    prediction_type = "str"
+
+    hf_metric_name = "meteor"
+    hf_metric_fields = ["meteor"]
+    hf_compute_args = {"alpha": 0.9, "beta": 3, "gamma": 0.5}
+    # the defaults in hf's Meteor
+
+
 class F1(GlobalMetric):
     _metric = None
     main_score = "f1_macro"
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -327,6 +327,11 @@ class MetricWithConfidenceInterval(Metric):
             # otherwise, the aggregation_func needs to be applied AFTER resampling the instances;
             #   that is, re-form the groups, calculate the function, and take the mean of the group scores
             aggregation_func = self.average_item_scores
+
+        method = "BCa"
+        if len(instances) > 100:
+            method = "percentile"
+
         for score_name in score_names:
             # If all computed instance level scores are the same, there is no point in computing
             # confidence intervals. So skip to the next score.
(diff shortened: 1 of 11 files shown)


OfirArviv:
@yoavkatz  maybe change it in the class itself? Not only in the catalog? That way if people fetch the metric as I did they still have the default

yoavkatz:
> @yoavkatz maybe change it in the class itself? Not only in the catalog? That way if people fetch the metric as I did they still have the default


Meteor (unlike rouge) does not have a class, and only is an instantiation of Huggingface metric.

I think the Rouge reason it is slow , is that in itself it does bootstrapping in the code:

https://huggingface.co/spaces/evaluate-metric/rouge/blob/main/rouge.py#L134

I think the solution is to avoid use aggregation.

dafnapension:
I think that the problem with the use of unitxt of HF Rouge and Meteor is making them Global, and even leaving the default arg of GlobalMetric: `process_single_instances = True
` in place.
In such a case, HF, that by itself invokes heavy language models, is invoked first for each and every instance (from the beginning of GlovalMetric.process()), and then for the whole stream (from the second part of that process()).  When HF is invoked for the whole stream, it again processes each and every instance.
Now if this happens with CI - the processing of each and every instance (which is heavy language model) occurs 2 * number of resamples.
If these HF metric are incarnated as InstanceMetric (and there is no reason why not), then each instance would have been processed just once!  the CI  uses existing scores, if there are, and there will be if so incarnated as InstanceMetric.

dafnapension (changes requested):
Please see my comment before you merge.  I think it is not a good idea to run away from CI when this is not necessary.

yoavkatz:
> @eladven , @yoavkatz , I think that the problem with the use of unitxt of HF Rouge and Meteor (and others) is in making them Global, and even leaving the default arg of GlobalMetric: `process_single_instances = True` in place. In such a case, HF, that by itself invokes heavy language models, is invoked first for each and every instance (from the beginning of GlovalMetric.process()), and then for the whole stream (from the second part of that process()). When HF is invoked for the whole stream, it again processes each and every instance. Now if this happens with CI - the processing of each and every instance (which is heavy language model) repeats 2 * number of resamples. If these HF metric are incarnated as InstanceMetric (and there is no reason why not), then each instance would have been processed just once! the CI uses existing scores, if there are, it does not compute them again, and there will be if so incarnated as InstanceMetric.
> 
> I think it is better to fix that issue and not flood the catalog with cards of yes/no CI.

I agree. I changed the code of Rouge to be HuggingfaceBulkMetric and it solved the issue for Rouge.

Please review and let me know what you think.

Regarding Meteor - can you do the appropriate thing?

dafnapension:
yes, coming up

dafnapension:
Hi @yoavkatz , I added Meteor as an HuggingFaceInstance,  to see how this works. With ci.
Also, I thought about rouge: if it is bulk, then each bulk gets the global score as the average of the bulk, not the average of the whole stream. Is that OK?
So I changed Rouge to inherit from this new HuggingFaceInstance.  Nothing else changed.
I also fixed a small bug in test_metric_utils, and things seem to be fine now, here, on my little laptop.
I hope that this new incarnation will survive also out in the wild fields of fm_eval..

yoavkatz:
> Hi @yoavkatz , I added Meteor as an HuggingFaceInstance, to see how this works. With ci. Also, I thought about rouge: if it is bulk, then each bulk gets the global score as the average of the bulk, not the average of the whole stream. Is that OK? So I changed Rouge to inherit from this new HuggingFaceInstance. Nothing else changed. I also fixed a small bug in test_metric_utils, and things seem to be fine now, here, on my little laptop. I hope that this new incarnation will survive also out in the wild fields of fm_eval..


In general BulkInstanceMetric does not perform aggregation, and assume the results returned per instance. It just a matter of optimization, for example in LLM based metrics we send a batch of requests at a time , instead of 1-1 which is much slower.

In Rouge, I removed the use_aggregation=True to the metric returns for each instance the rouge score, which is then averaged by the generic BulkInstanceMetric code.

yoavkatz:
@dafnapension - it would be if we add  a unit test that checks that if we wrap a metric (like Rouge) in HuggingfaceInstanceMetric or HuggingfaceBulkInstanceMetric - we would get the same results.

dafnapension:
A good idea, coming up!

dafnapension:
Added a comparison of Meteor to its previous global HF implementation, upon preparation of the metric (in Meteor.py).
Now looking into Rouge.

dafnapension:
Hi @yoavkatz ,  I added a test comparing old and new rouge - to show they return the same result.
Also, per a good advice from @arielge and @assaftibm , in case of more than 100 instances, I changed the method of bootstrap from "BCa" (the default) to "percentile".

src/unitxt/metrics.py
```diff
@@ -327,6 +327,11 @@ def score_based_confidence_interval(
             # otherwise, the aggregation_func needs to be applied AFTER resampling the instances;
             #   that is, re-form the groups, calculate the function, and take the mean of the group scores
             aggregation_func = self.average_item_scores
+
+        method = "BCa"
+        if len(instances) > 100:
```
yoavkatz, line 332: What the run time difference and accuracy between BCa and percentile?

dafnapension: I do not really know.. I took the advice of @arielge  and @assaftibm , here:
https://github.com/IBM/unitxt/issues/1008#issuecomment-2220794399
and
https://github.com/IBM/unitxt/issues/1008#issuecomment-2222678876

dafnapension:
Hi @yoavkatz ,  looking at these lines in `HuggingfaceBulkMetric`:
https://github.com/IBM/unitxt/blob/2381c4e4211074f8dee3e6fd14a3319b1bed3c0c/src/unitxt/metrics.py#L1305
it seems like you assume that every score that is a list, means one component per instance.  It is true for Rouge,  but not so in [bleu](https://huggingface.co/spaces/evaluate-metric/bleu), for example, where a score named `precisions` is an array whose length equals the input argument `max_order`.
I think that this breakdown of list scores to instance score should be done in the individual metric level, and not in the `BulkMetric` level.

yoavkatz:
> Hi @yoavkatz , looking at these lines in `HuggingfaceBulkMetric`:
> 
> https://github.com/IBM/unitxt/blob/2381c4e4211074f8dee3e6fd14a3319b1bed3c0c/src/unitxt/metrics.py#L1305
> 
> 
> it seems like you assume that every score that is a list, means one component per instance. It is true for Rouge, but not so in [bleu](https://huggingface.co/spaces/evaluate-metric/bleu), for example, where a score named `precisions` is an array whose length equals the input argument `max_order`.
> I think that this breakdown of list scores to instance score should be done in the individual metric level, and not in the `BulkMetric` level.

I suggest we complete this PR on Rouge and Metric, and then think about Bleu and generalizing this.

yoavkatz:
Note that there is HuggingfaceInstanceMetric which is used by Rouge and Meteor and HuggingfaceBulkMetric, which is only used by BertScore as far as I understand.

dafnapension:
Hi @yoavkatz , indeed I was referring to a generalization (for a future PR), trying to address your comment that performance-wise, if possible, better approach HF in batches.

src/unitxt/metrics.py
dafnapension: I found information about percentile and BCa [here](https://blogs.sas.com/content/iml/2017/07/12/bootstrap-bca-interval.html).
From my understanding (@assaftibm , @arielge , please correct my mistakes..),  
'percentile' works by the following intuitive notion:  make n_resamples of the population [the stream fed to metric],  each such resample is of the size of the population and is built by repeating the following process: independently, and identically distributed, select an item [instance] from the population, with replacement [return of a selected item back to the population].  
Then, given one resample, compute the statistics (the metric) on it, the same way the metric is computed on the original population. 
Then, given n_resamples scores (one per re_sample), return the 0.05 and 0.95 percentile ("quintile" of 100 rather than 5) as the ci borders.

BCa continues the above result of n_resamples scores, by correcting skewness and biases thereof, observed when comparing to the original population.

yoavkatz: I'm not sure it matters, but I saw somewhere the BCa is the one that is commonly used.   If the difference in runtime is small (I would think the main cost is recalculating the metric and not the BCa vs percentile), then I would keep it simple and not change behavior on some threshold.

So - can you try running twice (with BCA and with percentile) on say 200 instances, and see the difference in runtime?

assaftibm: Hi @dafnapension!

Yes - I think your understanding of Percentile bootstrap is correct. 

I would suggest the following:
1. When it's an InstanceMetric, use BCa bootstrap.
2. When it's a GlobalMetric, use Basic bootstrap (or Percentile, see discussion below). 

In 2, I wouldn't condition the type of bootstrap on the size of the dataset and switch between BCa and other lighter bootstrap because it can confuse people that test a full dataset vs. smaller samples of it - they will suddenly get different CIs due to a switch in the bootstrap type. I think it's better to stick to one for consistency.

Regarding Basic vs. Percentile bootstrap.
There is a description [here](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)#Methods_for_bootstrap_confidence_intervals).  I have to admit that I don't fully understand the computation of the Basic bootstrap. However, note that [Scipy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.bootstrap.html) includes this comment:
> While the 'percentile' method is the most intuitive, it is rarely used in practice. Two more common methods are available, 'basic' (‘reverse percentile’) and 'BCa' (‘bias-corrected and accelerated’); they differ in how step 3 is performed.

So this is the main reason why I would lean towards Basic over Percentile.  It would be easier if we could stick to BCa also in GlobalMetrics but unfortunately that doesn't seem practical.

As a side note, I think it makes sense to ask about the slowdown with BCa on StackOverflow - showing a short self-contained sample code that reproduces  the problem using Scipy's implementation and asking if this is as expected. It's possible that the implementation in Scipy is not optimal.

yoavkatz: @assaftibm - Are we sure the slow down is due to BcA vs basic (or percentile)?  I think the main slowdown is that if the metric is slow to compute, then all method will call the metric n_samples times.  If n_samples is 100, and the metric takes 1 second, it will take 100 second.

assaftibm: About the last note, there's an interesting discussion [here](https://stackoverflow.com/questions/74437089/time-consumption-of-scipys-bootstrap-as-a-function-of-the-number-of-resamples). In our short investigation, @arielge and I also noticed that it is the `jackknife` computation that dominates in BCa. 

In Scipy code, we can see [here](https://github.com/scipy/scipy/blob/87c46641a8b3b5b47b81de44c07b840468f7ebe7/scipy/stats/_resampling.py#L121) an outer for-loop over the dataset, and then [here](https://github.com/scipy/scipy/blob/87c46641a8b3b5b47b81de44c07b840468f7ebe7/scipy/stats/_resampling.py#L129) an inner for-loop over jackknife samples.  The statistic is computed therein.

arielge: @yoavkatz To clarify, the meaning is that because of the CI computation method (and specifically the `jackknife`), the metric is being called additional times. So it is still a function of the metric being slow, just that with BCa the metric is called many more than `n_resample` times.

dafnapension: > I'm not sure it matters, but I saw somewhere the BCa is the one that is commonly used. If the difference in runtime is small (I would think the main cost is recalculating the metric and not the BCa vs percentile), then I would keep it simple and not change behavior on some threshold.
> 
> So - can you try running twice (with BCA and with percentile) on say 200 instances, and see the difference in runtime?

Yes, @yoavkatz , you saw this comment at the documentation of scipy's bootstrap that we use.
Yes, I can try. Running on my own computer, so there is no question of loads by other processes.
I will use the `python -m cProfile -s cumtime xx.py` that I learned from you.

assaftibm: @yoavkatz  - to add more info:
* You are right that with global metrics, we need to compute the metric for each resample of the data. So it is `O(n_resamples)` and subject to noticeable slow down when the metric is slow to compute. But this is true no matter what the bootstrap is. So global metrics are slower than instance metrics by definition  as long as CIs are concerned. 
* The new thing (to me) that @arielge saw is that BCa adds a significant slow down by itself.  This is because it attempts to correct for biases in the data, and  part of it is the `jackknife` computation, which computes the metric for each entry in the dataset. It adds at least `O(len(dataset))` calls to the metric (I don't understand from the code whether the number is even higher, it could also be `O(len(dataset)^2)`, see the links I added in the reply to @dafnapension). So even if you reduce the value of `n_resamples` , if your dataset has 10k examples, it will compute the metric 10k times. (or even more) 

So I think that a practical solution would be:
* Instance metrics, use `n_resamples=1k` and `BCa` (or maybe keep it at the default of 10k if the diff is small)
* Global metrics, use `n_resmples=100` and `Basic`
*

dafnapension: I wrote a small recipe (I think I am not sufficiently familiar with the new structure thereof, and did several steps "manually":

```python
recipe = StandardRecipe(card="cards.numeric_nlg[task=tasks.generation[metrics=[metrics.meteor[n_resamples=100]]]]",num_demos=0,template_card_index=0,
                        loader_limit=300)
ms = recipe()
ms = {"train" : ms["train"]}
copyapp = Copy(field="target", to_field="prediction")
ms = copyapp(ms)
applypp = ApplyOperatorsField(operators_field="postprocessors")
ms = applypp(ms)
applymetric = ApplyMetric(metric_field="metrics", calc_confidence_intervals=True)
ms = applymetric(ms)
ms_dict = {key: list(ms[key]) for key in ms}
print({key: len(ms_dict[key]) for key in ms_dict})
print(ms_dict["train"][0]["score"])
```
for sanity check: the printed lines are:
```
{'train': 300}
{'global': {'meteor': 0.8932977564065374, 'score': 0.8932977564065374, 'score_name': 'meteor', 'meteor_ci_low': 0.8633536857531038, 'meteor_ci_high': 0.9123895754319261, 'score_ci_low': 0.8633536857531038, 'score_ci_high': 0.9123895754319261}, 'instance': {'meteor': 0.9998177842565598, 'score': 0.9998177842565598, 'score_name': 'meteor'}}
```

I then ran the above script through `cProfile`, in this new branch, having just turned off both conditional additions:
```
        method = "BCa"
        # if len(instances) > 100:
        #     method = "percentile"
```
so that first, only BCa is used, and got the following results:

```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      301    0.000    0.000   17.068    0.057 metrics.py:805(process)
        1    0.007    0.007   16.953   16.953 metrics.py:887(compute_instance_scores)
      300    0.005    0.000    3.760    0.013 metrics.py:1326(compute)
        2    0.000    0.000    3.027    1.514 metrics.py:1320(prepare)
        1    0.000    0.000    1.805    1.805 metrics.py:1(<module>)
        1    0.000    0.000    0.114    0.114 metrics.py:295(score_based_confidence_interval)
        3    0.000    0.000    0.100    0.033 metrics.py:342(statistic)
      402    0.001    0.000    0.096    0.000 metrics.py:260(average_item_scores)
      401    0.000    0.000    0.096    0.000 metrics.py:346(<lambda>)
      402    0.003    0.000    0.053    0.000 metrics.py:62(nan_mean)
      402    0.041    0.000    0.041    0.000 metrics.py:269(<listcomp>)
      300    0.002    0.000    0.008    0.000 metrics.py:159(_validate_reference)
      300    0.001    0.000    0.004    0.000 metrics.py:153(_validate_prediction)
      300    0.001    0.000    0.002    0.000 metrics.py:121(_add_score_prefixes_to_score_dict)
      601    0.001    0.000    0.001    0.000 metrics.py:174(get_prediction_type)
      900    0.001    0.000    0.001    0.000 metrics.py:114(_add_score_prefix)
       40    0.000    0.000    0.000    0.000 metrics.py:692(<lambda>)
       34    0.000    0.000    0.000    0.000 metrics.py:478(<lambda>)
        1    0.000    0.000    0.000    0.000 metrics.py:2711(Perplexity)
      300    0.000    0.000    0.000    0.000 metrics.py:1339(<dictcomp>)
        1    0.000    0.000    0.000    0.000 metrics.py:2798(Template)
        1    0.000    0.000    0.000    0.000 metrics.py:284(_all_instance_scores_equal)
        1    0.000    0.000    0.000    0.000 metrics.py:243(new_random_generator)
        3    0.000    0.000    0.000    0.000 metrics.py:371(resample_from_non_nan)
        6    0.000    0.000    0.000    0.000 metrics.py:1186(<lambda>)
        1    0.000    0.000    0.000    0.000 metrics.py:286(<listcomp>)
        1    0.000    0.000    0.000    0.000 metrics.py:1122(MetricPipeline)
        9    0.000    0.000    0.000    0.000 metrics.py:584(<lambda>)
```

dafnapension: Repeating all the above, now with the old meteor implementation (as a global HF metric):
First, with BCa  (what we now have in main branch):

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      301    0.006    0.000   58.748    0.195 metrics.py:473(process)
      702    0.004    0.000   48.132    0.069 metrics.py:540(_compute)
      702    0.032    0.000   48.128    0.069 metrics.py:1195(compute)
        1    0.000    0.000   45.006   45.006 metrics.py:397(compute_global_confidence_intervals)
        3    0.000    0.000   45.003   15.001 metrics.py:403(statistic)
      401    0.011    0.000   44.991    0.112 metrics.py:421(<lambda>)
      401    0.002    0.000   44.929    0.112 metrics.py:406(metric)
        2    0.000    0.000    3.350    1.675 metrics.py:1189(prepare)
        1    0.000    0.000    1.790    1.790 metrics.py:1(<module>)
      401    0.019    0.000    0.019    0.000 metrics.py:422(<listcomp>)
      401    0.017    0.000    0.017    0.000 metrics.py:424(<listcomp>)
      401    0.015    0.000    0.015    0.000 metrics.py:423(<listcomp>)
        1    0.000    0.000    0.003    0.003 metrics.py:130(_validate_references_and_prediction)
      301    0.002    0.000    0.003    0.000 metrics.py:121(_add_score_prefixes_to_score_dict)
      300    0.000    0.000    0.002    0.000 metrics.py:159(_validate_reference)
      903    0.001    0.000    0.001    0.000 metrics.py:114(_add_score_prefix)
      601    0.000    0.000    0.001    0.000 metrics.py:174(get_prediction_type)
       35    0.000    0.000    0.000    0.000 metrics.py:467(<lambda>)
       37    0.000    0.000    0.000    0.000 metrics.py:681(<lambda>)
      300    0.000    0.000    0.000    0.000 metrics.py:153(_validate_prediction)
        1    0.000    0.000    0.000    0.000 metrics.py:2648(Perplexity)
        2    0.000    0.000    0.000    0.000 metrics.py:1177(verify)
        9    0.000    0.000    0.000    0.000 metrics.py:1175(<lambda>)
        1    0.000    0.000    0.000    0.000 metrics.py:2735(Template)
        1    0.000    0.000    0.000    0.000 metrics.py:243(new_random_generator)
        1    0.000    0.000    0.000    0.000 metrics.py:1111(MetricPipeline)
        9    0.000    0.000    0.000    0.000 metrics.py:573(<lambda>)
        3    0.000    0.000    0.000    0.000 metrics.py:365(resample_from_non_nan)

```

dafnapension: So, a (unnecessarily) global - is clearly out of consideration.  I next checked the difference between both instance-implementation, across BCa vs percentile, when we increase n_resamples from 100 (as above) to 1000.

dafnapension: And last, increased the number of instances from 300 to 1200.
First, BCa:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1201    0.003    0.000   53.394    0.044 metrics.py:805(process)
        1    0.030    0.030   48.052   48.052 metrics.py:887(compute_instance_scores)
     1200    0.026    0.000   14.935    0.012 metrics.py:1326(compute)
        1    0.003    0.003    5.336    5.336 metrics.py:295(score_based_confidence_interval)
        3    0.000    0.000    5.153    1.718 metrics.py:342(statistic)
        1    0.000    0.000    5.079    5.079 metrics.py:1(<module>)
     2201    0.007    0.000    5.079    0.002 metrics.py:346(<lambda>)
     2202    0.062    0.000    5.075    0.002 metrics.py:260(average_item_scores)
     2202    3.603    0.002    3.603    0.002 metrics.py:269(<listcomp>)
        2    0.000    0.000    3.516    1.758 metrics.py:1320(prepare)
     2202    0.112    0.000    1.410    0.001 metrics.py:62(nan_mean)
     1200    0.008    0.000    0.045    0.000 metrics.py:159(_validate_reference)
     1200    0.006    0.000    0.015    0.000 metrics.py:153(_validate_prediction)
     1200    0.010    0.000    0.013    0.000 metrics.py:121(_add_score_prefixes_to_score_dict)
        1    0.000    0.000    0.006    0.006 metrics.py:284(_all_instance_scores_equal)
        1    0.004    0.004    0.004    0.004 metrics.py:289(<listcomp>)
     2401    0.003    0.000    0.003    0.000 metrics.py:174(get_prediction_type)
        1    0.003    0.003    0.003    0.003 metrics.py:286(<listcomp>)
     3600    0.002    0.000    0.002    0.000 metrics.py:114(_add_score_prefix)
     1200    0.001    0.000    0.001    0.000 metrics.py:1339(<dictcomp>)
       34    0.000    0.000    0.001    0.000 metrics.py:478(<lambda>)
       40    0.000    0.000    0.001    0.000 metrics.py:692(<lambda>)
        1    0.000    0.000    0.001    0.001 metrics.py:2711(Perplexity)
        1    0.000    0.000    0.001    0.001 metrics.py:2798(Template)
        1    0.000    0.000    0.000    0.000 metrics.py:243(new_random_generator)
        6    0.000    0.000    0.000    0.000 metrics.py:1186(<lambda>)
        3    0.000    0.000    0.000    0.000 metrics.py:371(resample_from_non_nan)
        9    0.000    0.000    0.000    0.000 metrics.py:584(<lambda>)
```
and then percentile:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1201    0.002    0.000   60.528    0.050 metrics.py:805(process)
        1    0.041    0.041   58.954   58.954 metrics.py:887(compute_instance_scores)
     1200    0.030    0.000   18.488    0.015 metrics.py:1326(compute)
        2    0.000    0.000    3.005    1.502 metrics.py:1320(prepare)
        1    0.000    0.000    1.900    1.900 metrics.py:1(<module>)
        1    0.004    0.004    1.570    1.570 metrics.py:295(score_based_confidence_interval)
        1    0.000    0.000    1.490    1.490 metrics.py:342(statistic)
     1000    0.002    0.000    1.469    0.001 metrics.py:346(<lambda>)
     1001    0.016    0.000    1.469    0.001 metrics.py:260(average_item_scores)
     1001    1.101    0.001    1.101    0.001 metrics.py:269(<listcomp>)
     1001    0.022    0.000    0.352    0.000 metrics.py:62(nan_mean)
     1200    0.010    0.000    0.045    0.000 metrics.py:159(_validate_reference)
     1200    0.007    0.000    0.017    0.000 metrics.py:153(_validate_prediction)
     1200    0.008    0.000    0.015    0.000 metrics.py:121(_add_score_prefixes_to_score_dict)
     3600    0.005    0.000    0.005    0.000 metrics.py:114(_add_score_prefix)
     2401    0.002    0.000    0.003    0.000 metrics.py:174(get_prediction_type)
        1    0.000    0.000    0.002    0.002 metrics.py:284(_all_instance_scores_equal)
     1200    0.002    0.000    0.002    0.000 metrics.py:1339(<dictcomp>)
        1    0.001    0.001    0.001    0.001 metrics.py:286(<listcomp>)
        1    0.001    0.001    0.001    0.001 metrics.py:289(<listcomp>)
       40    0.000    0.000    0.000    0.000 metrics.py:692(<lambda>)
       34    0.000    0.000    0.000    0.000 metrics.py:478(<lambda>)
        1    0.000    0.000    0.000    0.000 metrics.py:243(new_random_generator)
        1    0.000    0.000    0.000    0.000 metrics.py:2711(Perplexity)
        6    0.000    0.000    0.000    0.000 metrics.py:1186(<lambda>)
        1    0.000    0.000    0.000    0.000 metrics.py:2798(Template)
        9    0.000    0.000    0.000    0.000 metrics.py:584(<lambda>)
```

dafnapension: In all my experiments, because of lack of real data, I used prediction=target (first reference). I think it only change a constant time of computing each instance score, so I allowed myself..
In my humble opinion, in the numbers of instances and resamples that I see in the cards, we can stick to the said to be most commonly used BCa.   Just remember to check again if we grow dramatically.

yoavkatz: Thanks for the detailed analysis.   BCa seems to increase the time about 10%-20% in instance metrics, but more than doubles the time in global metrics.

yoavkatz: So do we agree to keep BCa always? I don't think we have many more global metrics that take a long time - and it's the simplest approach.

dafnapension: I, too, think so. Leave the change for a future consideration, if and when needed.
@elronbandel also asked that I simply copy over the code from what we find in HF to an in-house instance metric, for a further speedup.

src/unitxt/metrics.py
```diff
@@ -1709,26 +1824,33 @@ class Rouge(HuggingfaceMetric):
...
+        # We don't use the aggregation, to avoid running bootstrapping by the
+        # internal library (which is costly) and done by Unitxt in any case.
         self.hf_compute_args.update(
-            {"use_aggregator": self.use_aggregator, "rouge_types": self.rouge_types}
+            {"use_aggregator": False, "rouge_types": self.rouge_types}
         )
 
         import nltk
 
         nltk.download("punkt")
```
elronbandel, line 1835:
```suggestion
        nltk.download("punkt", quiet=True)
```

tests/library/test_metrics.py
```diff
@@ -799,19 +801,54 @@ def test_rouge(self):
...
+                self.hf_compute_args.update(
+                    {
+                        "use_aggregator": self.use_aggregator,
+                        "rouge_types": self.rouge_types,
+                    }
+                )
+
+                import nltk
+
+                nltk.download("punkt")
```
elronbandel, line 832:
```suggestion
                nltk.download("punkt", quiet=True)
```

src/unitxt/metrics.py
```diff
@@ -1300,6 +1301,79 @@ def compute(
...
+    alpha: float = 0.9
+    beta: int = 3
+    gamma: float = 0.5
+    # unitxt uses nltk version >= 3.8
+
+    def prepare(self):
+        import nltk
+
+        nltk.download("wordnet")
+        nltk.download("omw-1.4")
```
elronbandel, line 1351:
```suggestion
        nltk.download("wordnet", quiet=True)
        nltk.download("omw-1.4", quiet=True)
```

src/unitxt/metrics.py
```diff
@@ -1300,6 +1301,79 @@ def compute(
...
+        from datasets.config import version
+
+        nltk_version = version.parse(importlib_metadata.version("nltk"))
+        assert nltk_version >= version.Version(
+            "3.6.6"
+        ), "nltk version must be at least 3.6.6"
+
+    def compute(self, references, prediction, task_data):
+        from nltk import word_tokenize
+        from nltk.translate import meteor_score
```
elronbandel, line 1365: move to prepare so import is performed only once

src/unitxt/metrics.py
```diff
@@ -1691,16 +1765,57 @@ class F1MacroMultiLabel(F1MultiLabel):
...
+
+    sent_split_newline: bool = True
+    _requirements_list: List[str] = ["nltk", "rouge_score"]
+
+    def compute(self, references: List[Any], prediction: Any, task_data: Dict) -> dict:
+        import nltk
+        from rouge_score import rouge_scorer
+
+        nltk.download("punkt")
+        self.sent_tokenize = nltk.sent_tokenize
```
elronbandel, line 1784: move to prepare

elronbandel (changes requested):
@dafnapension looks great. only need to move imports to prepare() and make nltk downloads quiet ( see my comments)

Also consider making new PR only with your changes since there is this .secrets.basline that seem to be affected by wrong merging of the main

src/unitxt/metrics.py
dafnapension: thanks, @elronbandel , at last I learned (from you, as always) how to avoid those "already up to date" running on my screen.

If I import in `prepare`,  `process` says it does not recognize `nltk`, for example.
Is it only my VS code? or the truth?

src/unitxt/metrics.py
dafnapension: image

dafnapension: Hi @elronbandel , the above happens if I move to prepare.  is it אזעקת שוא?

elronbandel: You need to save it under self.scorer then use it in the compute

dafnapension: Hi @elronbandel , Done.

src/unitxt/metrics.py
assaftibm: I think the problem will appear with more complex global metrics like CorpusBLEU, but I agree it can be dealt with in  a separate PR.

src/unitxt/metrics.py
yoavkatz: @elronbandel @dafnapension - can we merge these?

PR #1013, opened 2024-07-10
WML Inference Engine fix

pawelknes:
Fixes to `WMLInferenceEngine`:
* changed `_requirements_list` to search for `ibm_watsonx_ai` package (which is the name of import used in Python)
* added typing `Any` for both `client` and `credentials` (without typing, there was an issue when trying to pass values to these attributes when using unitxt as package in other repository)


src/unitxt/inference.py
```diff
@@ -351,8 +351,8 @@ class WMLInferenceEngine(InferenceEngine, PackageRequirementsMixin):
         results = wml_inference.infer(dataset["test"])
     """
 
-    client = None
-    credentials = None
+    client: Any = None
```
yoavkatz, line 354: Typing is important to allow override (that is an odd python behavior), but can we use stricter types than Any?

pawelknes: Initially I wanted to use types from `ibm_watsonx_ai` but it was causing errors since the package needs to be installed. And not everyone may want to install it as well as the package is incompatible with older Python versions. Having said this, I described both variables in the documentation

yoavkatz: Ok. Make sense.

yoavkatz (approved):

PR #1019, opened 2024-07-11
Additional inference parameters for openai and genai

pawelknes:
Added additional inference parameters for both `IbmGenAiInferenceEngine` and `OpenAiInferenceEngine`

### DIFF
--- a/src/unitxt/artifact.py
+++ b/src/unitxt/artifact.py
@@ -557,16 +557,3 @@ def get_artifacts_data_classification(artifact: str) -> Optional[List[str]]:
         return None
 
     return data_classification.get(artifact)
-
-
-def return_artifact_specific_kwargs(
-    artifact: Artifact, keep_empty: bool = True
-) -> Dict[str, Any]:
-    attributes = artifact.__annotations__.keys()
-    kwargs = {}
-    for attribute in attributes:
-        value = getattr(artifact, attribute)
-        if not keep_empty and value is None:
-            continue
-        kwargs[attribute] = value
-    return kwargs
--- a/src/unitxt/artifact.py
+++ b/src/unitxt/artifact.py
@@ -557,3 +557,16 @@ def get_artifacts_data_classification(artifact: str) -> Optional[List[str]]:
         return None
 
     return data_classification.get(artifact)
+
+
+def return_artifact_specific_kwargs(
+    artifact: Artifact, keep_empty: bool = True
+) -> Dict[str, Any]:
+    attributes = artifact.__annotations__.keys()
+    kwargs = {}
+    for attribute in attributes:
+        value = getattr(artifact, attribute)
+        if not keep_empty and value is None:
+            continue
+        kwargs[attribute] = value
+    return kwargs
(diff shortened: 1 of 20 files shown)


pawelknes:
@yoavkatz

elronbandel (changes requested):
Is it necessary on top of `to_dict()` inherited from Dataclass?

pawelknes:
> Is it necessary on top of `to_dict()` inherited from Dataclass?

`to_dict` also returns attributes of the `Artifact` class, so there would be need to filter these parameters each time anyway

elronbandel:
So it is just the immediate class parameters without its ancestor?

pawelknes:
> So it is just the immediate class parameters without its ancestor?

Yup

src/unitxt/artifact.py
```diff
@@ -557,3 +557,16 @@ def get_artifacts_data_classification(artifact: str) -> Optional[List[str]]:
...
+    artifact: Artifact, keep_empty: bool = True
+) -> Dict[str, Any]:
+    attributes = artifact.__annotations__.keys()
+    kwargs = {}
+    for attribute in attributes:
+        value = getattr(artifact, attribute)
+        if not keep_empty and value is None:
+            continue
+        kwargs[attribute] = value
+    return kwargs
```
elronbandel, line 572: Given this is the functionality and the fact it uses Dataclass internals I think it shuold be a function of Dataclass.
For the benefit of future users it might be good if it was a functionality of `to_dict` something like `to_dict(without_parent=True)`. What do you think @pawelknes ?

elronbandel (changes requested):
One minor comment, other than that its good, yet I think the real solution to this mass parameters passing is turning Inference Engine to Functors like any other class in unitxt and having the GenerationParams inherited as Mixin.

src/unitxt/artifact.py
yoavkatz: @elronbandel - do you mean that it would be a function of Artifact called to_dict()?

pawelknes: Yup, that's good idea

pawelknes: I've moved the method to `Dataclass` and also modified it so that it is now able to return parameters of chosen specified parent classes (maybe that will be useful in future)

And I've also made `InferenceEngineParams` classes to be Mixin now as you suggested

elronbandel (approved):
To me this is much more neat! Thanks @pawelknes! The solution of getting arguments given parent class is really cool!

yoavkatz (changes requested):
I like the changes - but note they are not backward compatible.  We need to make the changes gradually.

pawelknes:
> I like the changes - but note they are not backward compatible. We need to make the changes gradually.

So should we revert changes to main?

yoavkatz:
Do we have a solution to make it backward compatible? There are pro's and con's in keeping old code.  

At the minimum, we should keep  IbmGenAiInferenceEngineParams  as a deprecated class, so  there will be a clear error message.

yoavkatz:
So I suggest we revert for and make these changes.

PR #1020, opened 2024-07-11
Add a guide for using unitxt with lm-evaluation-harness

elronbandel:


docs/docs/lm_eval.rst
```diff
@@ -0,0 +1,51 @@
...
+
+If you're unsure about your choice, you can use the :ref:`Explore Unitxt <demo>` tool for an interactive recipe exploration UI. After making your selection, click on Generate Prompts, and then click on the Code tab. You will see a code snippet such as the following:
+
+.. code-block:: python
+
+    dataset = load_dataset('unitxt/data', 'card=cards.wnli,template=templates.classification.multi_class.relation.default,max_train_instances=5', split='train')
+
+The second string parameter to `load_dataset()` is the recipe. Note that you will might want to remove `max_train_instances=5` from the recipe before using it, as the `max_train_instances`. If you wish to use few-shot in-context learning, you should configure this using the `num_demos` and `demos_pool_size` parameters instead e.g. `num_demos=5,demos_pool_size=10`.
+
+Once you have your Unitxt recipe set go and save it in the lm-evaluation-harness `tasks/unitxt` directory.
```
yoavkatz, line 29: If i install lm eval via pip,  where is this directory?if I understand right ,  In some internal conda or virtual env directory. We should allow it to work from a yaml file in the current directory.

elronbandel: Is it supported by lm eval? Can you create new task yamls anywhere?

yoavkatz: Yes. There is an include path option

--include_path : Accepts a path to a folder. If passed, then all YAML files containing lm-eval compatible task configurations will be added to the task registry as available tasks. Used for when one is writing config files for their own task in a folder other than lm_eval/tasks/.

yoavkatz: I'm not sure oif

Include: unitxt 

will work.  May need to online it's content.

yoavkatz (approved):

PR #1024, opened 2024-07-15
Backward compatibility for inference engines

pawelknes:
Provided backward compatibility for inference engines (support for passing inference parameters as a class to an engine)

For #1019


pawelknes:
@yoavkatz

yoavkatz (approved):
Thanks for adding this.

Issue #406, opened 2023-12-18
Bug: Dataclass not overriding properties in inheritance


elronbandel:
see: https://github.com/IBM/unitxt/pull/403

elronbandel:
https://github.com/IBM/unitxt/pull/403#issue-2046548367

PR #1027, opened 2024-07-17
Fix bug in data classes and add support for field overriding in fields containing types or functions

elronbandel:
Closes: https://github.com/IBM/unitxt/issues/406


tests/library/test_dataclass.py
```diff
@@ -276,6 +278,30 @@ class Child(Mixin, Parent1):
...
+
+        class GrandChild(Child):
+            c = 7
+            pass
+
+        child = GrandChild(b=2, c=3)
+
+        self.assertEqual(child.t(), 5)
+        self.assertEqual(child.b, 2)
+        self.assertEqual(child.c, 3)
```
elronbandel, line 303: This test would fail without the fix

PR #1030, opened 2024-07-18
changed input and output of templates to "input_fields" and "reference_ fields" - Non backward compatible

yoavkatz:
This is to continues  the work done on tasks.

Non backward compatible
Template methods changes:

```
inputs_to_instruction_and_target_prefix(self, inputs) ->    input_fields_to_instruction_and_target_prefix(self, input_fields)

inputs_to_source(self, inputs: Dict[str, object]) ->  input_fields_to_source(self, input_fields: Dict[str, object])

outputs_to_target_and_references(self, outputs: Dict[str, object]) -> reference_fields_to_target_and_references(self, reference_fields: Dict[str, object])

preprocess_inputs_and_outputs(   self, inputs: Dict[str, Any], outputs: Dict[str, Any] --> preprocess_input_and_reference_fields(self, input_fields: Dict[str, Any], reference_fields: Dict[str, Any]
   
```

### DIFF
--- a/docs/docs/adding_template.rst
+++ b/docs/docs/adding_template.rst
@@ -77,30 +77,32 @@ Making Your Custom Template
 ----------------------------
 
 In order to make your own template, you need to create a class inheriting from `Template` and
-implementing its two abstract methods:
+implementing its abstract methods:
 
 .. code-block:: python
 
-    @abstractmethod
-    def inputs_to_source(self, inputs: Dict[str, object]) -> Tuple[str, str]:
+     @abstractmethod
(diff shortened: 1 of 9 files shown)


elronbandel:
I think we could make it backward compatible by allowing user to either override the old name or the new name. not sure how necessary it is. What do you think @yoavkatz ?

yoavkatz:
I don't think many people created new template types - so keeping both methods may cause more confusion than not.  Just a few in FM-eval.

We just need to clearly not it in the release it.

 @eladven - what do you think?

elronbandel (approved):

Issue #1006, opened 2024-07-09
Current Formats do support different target_prefix in demos


yoavkatz:
Today the code in formats.py assumes the target prefix for the demos is taken from the instance.  This may cause subtle issue when the target_prefix uses input field for formatting (e.g. target_prefix="The answer for {num1} + {num2} is")

elronbandel:
can you explain why it causes issue in that case? or in what cases it will fail?

yoavkatz:
Since we take the target_prefix from the main instance and not the demos, if the instance is verbalized the calculation of 1+2 and the demo need to verbalize 4+5, it will use the instance target_prefix verbalization in the demos: 


User: calculate 4+5?
Assistant: The answer of 1+2 is 9 <--- demo (note the wrong target_prefix)

User: calculate 1+2?
Assistant: The answer of 1+2 is

dafnapension:
Resolved by PR #1031

PR #1031, opened 2024-07-18
demo's target prefix is now taken from demo instance

dafnapension:
for issue #1006


tests/library/test_formats.py
```diff
@@ -312,6 +320,93 @@ def test_system_format(self):
...
+        template = MultipleChoiceTemplate(
+            input_format="The following are multiple choice questions (with answers) about {topic}.\n{question}\nPotential Answers:\n{choices}\nAnswer:",
+            target_field="answer",
+            postprocessors=["processors.first_character"],
+            choices_separator="\n",
+            target_prefix="The {core_of_question} is: ",
+        )
+
+        card = TaskCard(
+            loader=LoadHF(path="cais/mmlu", name="elementary_mathematics"),
```
yoavkatz, line 340: Loading from HF is slow, so we should avoid it in unit test.  Instead you can use LoadFromDict to load a few examples in memory.

tests/library/test_formats.py
```diff
@@ -312,6 +320,93 @@ def test_system_format(self):
...
+                    expression='question[12+question.lower().index("what is the "):].replace("?","")',
+                ),
+            ],
+            # task="tasks.qa.multiple_choice.with_topic",
+            task=task,
+            # templates=["templates.qa.multiple_choice.with_topic.mmlu"],
+            templates=[template],
+        )
+
+        recipe = StandardRecipe(
```
yoavkatz, line 356: load_dataset() is simpler to use.

dafnapension:
Thanks, @yoavkatz , I will change to something small and local. Just want to understand first the eagering-related bug that popped up on this test.

yoavkatz (approved):

PR #1032, opened 2024-07-20
surfacing (and suggesting a solution for) a problem: eagering mode selects different demo cases than non-eagering

dafnapension:
Born [here.](https://github.com/IBM/unitxt/pull/1031#issuecomment-2238891519)


yoavkatz (changes requested):
This PR disables the ability for samplers that do anything more complex than random selection of instances from the demo pool.  There is already sampler that balance the labels of the returned samples, and a new PR #1034 adds ability to sample demo instances based on the similarity of the current instance.


Therefore, I don't think this is the solution for any indeterministic behavior.

yoavkatz:
Hi Dafna.  Per my comment above, I think there is overlap between this pr and #1034 .

Issue #948, opened 2024-06-25
OLMES Standard Compliance


elronbandel:
OLMES includes the following elements, justified in detail above:
- [ ]  Use test set when available, otherwise validation. Sample 1000 instances if more than 1500
- [x] Use specified, exact prompt format (Section 3.1)
- [ ]  Use fixed, curated 5-shot examples (Section 3.2)
   - [x] balanced
   - [ ] verified
- [ ] Evaluate with both MCF and CF, use the best result (Section 3.4)
   - [x] add CF template (choices specified)
   - [ ] add MCF template  (choices and numerals are not specified)
   - [ ] choose the best between CF and MCF at system level
- [ ]  Follow recommendations for all other evaluation details:

  - [x] For MMLU: use macro average (over 57 tasks) rather than micro average (over 14042 instances), following AI@Meta (2024). This better represents the diversity of fields in the dataset, although in practice it does not generally make a big difference (see Figure 7).
  - [x] When a model requires it, make sure to add the appropriate <bos> token at start of prompt (e.g., Gemma (Gemma Team et al., 2024)).
  - [ ] When using the “character” normalization forCF, include the leading space in the calculation of answer length.
  - [ ] Restrict all inputs (with completions) to 2048 tokens for consistency across models
  - [ ] Use the default model precision when evaluating (i.e., avoid options like load_in_8bit
unless it produces identical results)
  - [ ] OLMES uses the standard approach of two newlines to separate each in-context example.
      - [ ] (set unitxt format for olmes?)
  - [x] Other than the original instruction line for MMLU Hendrycks et al. (2021), we do not add
any extra instructions. This is in view of previous work finding the subject information from instructions makes little changes to model ranking (Alzahrani et al., 2024), and to reduce additional sources of variation in the prompt.

PR #1034, opened 2024-07-21
Add CloseTextSampler and FixedIndicesSampler

yoavkatz:
CloseTextSampler -  returns demos that are textually close to the current instance.
FixedIndicesSampler - return fixed demos based on indices in the train pool

Also fixed a bug in reproducibility of demos by making sure randomization is based on instance, and being global (so it is shared with ther instances)

### DIFF
--- a/src/unitxt/splitters.py
+++ b/src/unitxt/splitters.py
@@ -2,7 +2,6 @@ import itertools
 from abc import abstractmethod
 from copy import deepcopy
 from difflib import get_close_matches
-from random import Random
 from typing import Dict, List, Optional
 
 from .artifact import Artifact
@@ -111,7 +110,6 @@ class SliceSplit(Splitter):
 
 class Sampler(Artifact):
     sample_size: int = None
-    random_generator: Random = new_random_generator(sub_seed="Sampler")
 
     def prepare(self):
         super().prepare()
@@ -124,11 +122,6 @@ class Sampler(Artifact):
             ), f"sample_size must be a natural number, got {self.sample_size}"
             size = int(size)
         self.sample_size = size
-
-    def init_new_random_generator(self):
-        self.random_generator = new_random_generator(
-            sub_seed="init_new_random_generator"
-        )
 
     @abstractmethod
     def sample(
@@ -161,7 +154,8 @@ class RandomSampler(Sampler):
         instance: Optional[Dict[str, object]] = None,
     ) -> List[Dict[str, object]]:
         instances_pool = list(instances_pool)
-        return self.random_generator.sample(instances_pool, self.sample_size)
+        random_generator = new_random_generator(sub_seed=str(instance))
+        return random_generator.sample(instances_pool, self.sample_size)
 
 
 class FixedIndicesSampler(Sampler):
@@ -218,7 +212,8 @@ class CloseTextSampler(Sampler):
             for instance_in_pool in instances_pool
             if dict_get(instance_in_pool, field) in closest_matches
         ]
-        return self.random_generator.sample(instances_pool, self.sample_size)
+        random_generator = new_random_generator(sub_seed=str(instance))
+        return random_generator.sample(instances_pool, self.sample_size)
 
 
 class DiverseLabelsSampler(Sampler):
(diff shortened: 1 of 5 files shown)


src/unitxt/splitters.py
```diff
@@ -151,13 +153,52 @@ def filter_source_by_instance(
...
+        )
+        # Randmly select 'sample_size' instances that are from the closest matches text
+        # (There may be multiple instance with same text in the given field, and the order returned is
+        # is also randomized )
+        instances_pool = [
+            instance_in_pool
+            for instance_in_pool in instances_pool
+            if dict_get(instance_in_pool, field) in closest_matches
+        ]
+        return self.random_generator.sample(instances_pool, self.sample_size)
```
elronbandel, line 199: Why sampling is needed? isn't a list of size sample_size by now?

src/unitxt/splitters.py
```diff
@@ -151,13 +153,52 @@ def filter_source_by_instance(
...
 
 
+class CloseTextSampler(Sampler):
+    """Selects the samples of instances which are the closest textual match to the given instance.
+
+    Comparison is done based on a given field in the instance.
+
+    """
+
+    field: str
```
elronbandel, line 174: Maybe have a None default that uses just the serialization of the inputs?

elronbandel (changes requested):
See comments

src/unitxt/splitters.py
yoavkatz: The reason is noted above:
# (There may be multiple instance with same text in the given field, and the order returned is
# is also randomized )

src/unitxt/splitters.py
yoavkatz: This seems bit of an odd default.  Consider inputs such as "text_type" or "classes" , which you don't want to take part in the the calculation.

elronbandel: The do not harm either. Anyways just a suggestion

elronbandel (approved):

elronbandel:
Relevant to https://github.com/IBM/unitxt/issues/948

src/unitxt/splitters.py
```diff
@@ -151,11 +146,74 @@ def filter_source_by_instance(
...
+
     def sample(
-        self, instances_pool: List[Dict[str, object]]
+        self,
+        instances_pool: List[Dict[str, object]],
+        instance: Optional[Dict[str, object]] = None,
     ) -> List[Dict[str, object]]:
         instances_pool = list(instances_pool)
-        return self.random_generator.sample(instances_pool, self.sample_size)
+        random_generator = new_random_generator(sub_seed=str(instance))
```
elronbandel, line 157:
```suggestion
        random_generator = new_random_generator(sub_seed={**instance["inputs"], **instance["outputs"]})
```
Make it more robust to future changes in the instance structure (such as classification policy metadata etc that does not change the essence of the instance)

elronbandel (dismissed):
Looks like really important changes. Should we add to the metadata the indices of chosen demos in all the cases? It can help people reproduce or just locate good demos in the future.

src/unitxt/splitters.py
yoavkatz: Thanks.  I changed to code to use "input_fields" (references are not always available, and should not effect randomization if the input fields are the same).

yoavkatz:
> Looks like really important changes. Should we add to the metadata the indices of chosen demos in all the cases? It can help people reproduce or just locate good demos in the future.

I'm not sure this is a required use case.    For most uses, there are different demos per example, so when you average the score, you average over different indices.  You don't really need to know which examples instance indices were chosen for each text.

dafnapension:
Hi @yoavkatz  and @elronbandel ,  I think this is very nice, and understand developed per needs raised, and certainly solves the inconsistency problem.

In my branch, I used the occasion to remove a rather strange class, I think, `InstanceOperatorWithMultiStreamAccess`  which was only used for `SpreadSplit`,  and implemented `SpreadSplit` as a `MultiStreamOperator`.  Would you like to add this change also here?

yoavkatz:
> Hi @yoavkatz and @elronbandel , I think this is very nice, and understand developed per needs raised, and certainly solves the inconsistency problem.
> 
> In my branch, I used the occasion to remove a rather strange class, I think, `InstanceOperatorWithMultiStreamAccess` which was only used for `SpreadSplit`, and implemented `SpreadSplit` as a `MultiStreamOperator`. Would you like to add this change also here?

I think while right now this is the SpreadSplit is the only class that uses InstanceOperatorWithMultiStreamAccess , it may be useful for other classes in the future.

PR #1035, opened 2024-07-21
replace type and __type__ in type error

perlitz:


elronbandel (approved):
Thanks for that catch! I hope there are no tests relying on this verbatim

PR #1038, opened 2024-07-22
add validation to tldr, remove shuffle from billsum

alonh:
(shuffled by the SplitRandomMix)


perlitz (approved):
Changes has consistency issue
But it is OK

PR #1043, opened 2024-07-22
Added dedicated Unitxt warning and error classes  to link online documentaion

yoavkatz:


src/unitxt/doc_utils.py
```diff
@@ -0,0 +1,7 @@
+DOCUMENTATION_URL = "https://www.unitxt.ai/en/latest/"
+DOCUMENTATION_HUGGINGFACE_METRICS = "docs/adding_metric.html#adding-a-hugginface-metric"
+DOCUMENTATION_ADDING_TASK = "docs/adding_task.html"
```
elronbandel, line 3: Maybe better in the constants in settings_utils?

yoavkatz: There idea is to add many such links in the error messages of unitxt, so I think a dedicated file makes sense.

yoavkatz:
@elronbandel , @eladven  - this changes fails, probably because this file is not added to HF repo:

"/home/runner/.cache/huggingface/modules/datasets_modules/datasets/dataset/97bc70e0c4ed17863d1923cc78380a6ca44ca9a6251da45bb0b69c50ba0b1c90/task.py", line 6, in <module>
    from .doc_utils import DOCUMENTATION_ADDING_TASK, additional_info
ModuleNotFoundError: No module named 'datasets_modules.datasets.dataset.97bc70e0c4ed17863d1923cc78380a6ca44ca9a6251da45bb0b69c50ba0b1c90.doc_utils'

======================================================================
FAIL: test_dataset_imports (test_hf_artifacts.HFTests)
----------------------------------------------------------------------
  File "/home/runner/work/unitxt/unitxt/tests/library/test_hf_artifacts.py", line 18, in test_dataset_imports
    self.assertEqual(missing_imports, [])
AssertionError: Lists differ: ['doc_utils'] != []

First list contains 1 additional elements.
First extra element 0:
'doc_utils'

- ['doc_utils']
+ []

dafnapension (approved):

elronbandel (changes requested):
Thinking about this again I think it will be better if we had a basic UnitxtError class which can get docs link in its constructor. This will allow further modifications in the future and better control.

src/unitxt/error_utils.py
```diff
@@ -0,0 +1,44 @@
+from typing import Optional
+
+from .logging_utils import get_logger
+
+logger = get_logger()
+
+DOCUMENTATION_URL = "https://www.unitxt.ai/en/latest/"
+DOCUMENTATION_HUGGINGFACE_METRICS = "docs/adding_metric.html#adding-a-hugginface-metric"
+DOCUMENTATION_ADDING_TASK = "docs/adding_task.html"
```
elronbandel, line 9:
```suggestion
class Documentation:
    base = "https://www.unitxt.ai/en/latest/"
    HUGGINGFACE_METRICS = "docs/adding_metric.html#adding-a-hugginface-metric"
    ADDING_TASK = "docs/adding_task.html"
```
Maybe more neat and useful in @benjaminsznajder constants class style?

Then you have auto complete and only one import per file

elronbandel (approved):
lgtm, see minor comment

src/unitxt/error_utils.py
yoavkatz: Good idea. Done.

PR #1045, opened 2024-07-23
Real types in tasks and metrics

elronbandel:


prepare/cards/coedit.py
```diff
@@ -120,7 +120,8 @@
                 "input_type": "sentence",
             }
         ),
-        RenameFields(field_to_field={"src": "input", "tgt": "output_choice"}),
+        RenameFields(field="src", to_field="input"),
+        IndexOf(search_in="choices", index_of="tgt", to_field="output_choice"),
```
yoavkatz, line 124: Is this a bug fix? Should be noted in change log.

src/unitxt/operator.py
```diff
@@ -135,8 +135,12 @@ class MultiStreamOperator(StreamingOperator):
 
     caching: bool = NonPositionalField(default=None)
 
-    def __call__(self, multi_stream: Optional[MultiStream] = None) -> MultiStream:
+    def __call__(
+        self, multi_stream: Optional[MultiStream] = None, **kwargs
```
yoavkatz, line 139: How is this change related? What does it do? Not clear.

elronbandel: Completely unrelated will delete that

elronbandel: done

yoavkatz (approved):

pawelknes (approved):
L

PR #1054, opened 2024-07-24
Scigen

ShirApp:
llm_as_judge class expects question, answer, reference_answer fields so we could not use our own task
@elronbandel


elronbandel:
@eladven @OfirArviv can you advise how to get this done?

OfirArviv:
what is the task? I'm not sure I understand. Can you please provide an example?

csrajmohan:
@OfirArviv We are trying to use llm_as_judge for a table2text generation task. Here 'table' and 'caption' are input and 'text' is output. When creating a new llm_as_judge metric for this task, we are using `rating.single_turn_with_reference` as task for the time-being by mapping question->table, answer->text and reference_answer->reference_text fields. Ideally we want to use our own task here that has table, caption, text and reference text as fields. But llm_as_judge class seems to limit supported task types to `rating.single_turn_with_reference` and `rating.single_turn` only. What is the right way to add custom task definitions with different input/outputs for llm_as_judge?

csrajmohan:
Say I want to use this custom task [prepare/tasks/response_assessment/rating/table2text_single_turn_with_reference.py ](https://github.com/IBM/unitxt/pull/1054/files#diff-ab69e6f0a36e84dd877fe8385815e3e57ed20f9d000ecd14e999232d8ac0c446)while defining a new llm_as_judge metric here [prepare/metrics/llm_as_judge/rating/llama_3_ibm_genai_table2text_template.py](https://github.com/IBM/unitxt/pull/1054/files#diff-45e61ea7d0e54bb6fd84b5d641ba3482aae49bdd2b44a5fad3b4236861d5c320)
It is not possible right now as llm_as_judge class complains about unsupported task.

OfirArviv:
I think there is a confusion between the evaluated task and the judge task.

You have a model you want to evaluate, let's say mistral-7b. This model get a `prompt` with  `table` and produce an `answer`.
We can call the  task here `table_captioning`. This task is not related to the llm-as-judge. This is the task that is done on the evaluated model. The dataset of this table_captioning task also contains a reference_answer for each entry.

The llm as judges gets the inputs the evaluated received (the `question`), it's `answer`, and the `reference_answer` ad produce a rating (which you can choose what it will be in the judge prompt)

csrajmohan:
`The llm as judges gets the inputs the evaluated received (the question), it's answer, and the reference_answer ad produce a rating (which you can choose what it will be in the judge prompt)`

Here I want to pass the table and table caption(two fields) to the judge model, along with answer and reference_answer to produce a rating. Should the judge model always have only one field question as input?

OfirArviv:
> `The llm as judges gets the inputs the evaluated received (the question), it's answer, and the reference_answer ad produce a rating (which you can choose what it will be in the judge prompt)`
> 
> Here I want to pass the table and table caption(two fields) to the judge model, along with answer and reference_answer to Can you please give un an example of an input and output the evaluated model will get and output, and the prompt you want to give the judge in that instance?

yoavkatz:
Hi.  I've updated the LLM as judge example to show how to evaluate more complex tasks (e.g. with document context):

https://github.com/IBM/unitxt/blob/main/examples/standalone_evaluation_llm_as_judge.py

@csrajmohan  - please let me know if this helps.

csrajmohan:
Based on discussions with @OfirArviv and also the example shared by @yoavkatz , we have addressed our task requirement using provided llm_as_judge functionality in Unitxt and created a fresh PR for the same https://github.com/IBM/unitxt/pull/1134

Issue #1059, opened 2024-07-25
A petty issue with MetricsEnsemble's global score


dafnapension:
As seen in the unit test:

https://github.com/IBM/unitxt/blob/a0c39a5267da52e6d37e4ad904c93914b80c56a9/tests/library/test_metrics.py#L1712

In the global score,  as expected, `"score_name": "ensemble_score"`,  `"score": 0.44`, and `"ensemble_score": 0.44`.
However, the CI-s reported in  `"score_ci_high": 0.56`,  and  `"score_ci_low": 0.0`,  are those left from `ensemble_1_recall_macro`  (the last of self.metrics for which CI is computed). 
The association of these values (ci_high = 0.56, and ci_low = 0.0) with "score" is misleading here.

dafnapension:
This inconsistency (of `score_ci_high` / `score_ci_low`  with `score_name`  can be reproduced also with a task of several metrics (each having its own `main_score`), the last of which to be applied (being the first one in task.metrics) gives its name to `score_name`,  but `score_ci_low` and `score_ci_high` are simply the last ones computed in the task:

```python
from unitxt.artifact import fetch_artifact
from unitxt.operators import ApplyMetric, CastFields, Copy, Perturb
from unitxt.standard import StandardRecipe
from unitxt.settings_utils import get_settings
from unitxt.text_utils import print_dict

settings = get_settings()
settings.allow_unverified_code=True

card, _ =fetch_artifact("cards.coedit_error_detection")
print(card.task.metrics)

recipe = StandardRecipe(card=card, loader_limit=50, demos_pool_size=0, num_demos=0, template_card_index=0)
ms = recipe()
copyoperator = Copy(field="target", to_field="prediction")
ms = copyoperator(ms)
castfieldoperator = CastFields(fields={"prediction":"float"}, failure_defaults={})
ms = castfieldoperator(ms)
castfieldoperator = CastFields(fields={"references":"int"}, failure_defaults={}, process_every_value=True)
ms = castfieldoperator(ms)
perturboperator = Perturb(field="prediction", select_from=[0,1], percentage_to_perturb=60)
ms = perturboperator(ms)
applymetricoperator = ApplyMetric(metric_field="metrics", calc_confidence_intervals=True)
ms = applymetricoperator(ms)
trains = list(ms["train"])
print(len(trains))
print_dict(trains[0]["score"]["global"])

print("**** no CI in last metric applied, the metric that is marked by 'score_name'*******")
[...]

PR #1065, opened 2024-07-26
suggested patch for score_ci inconsistency issue

dafnapension:
A suggested patch for issue #1059

### DIFF
--- a/src/unitxt/metrics.py
+++ b/src/unitxt/metrics.py
@@ -231,31 +231,34 @@ class Metric(Artifact):
     def disable_confidence_interval_calculation(self):
         pass
 
-    # update instance["score"]["global"] with the newly computed global score, gs, for the
-    # current metric computed.  gs contains "score" and "score_name" fields that reflect
+    # update instance["score"]["global"] with the newly computed global score, global_score, for the
+    # current metric computed.  global_score contains "score" and "score_name" fields that reflect
     # (the main_score of) the current metric.
     # A simple python-dictionary-update adds new fields to instance["score"]["global"], and also replaces the values
     # of its fields "score" and "score_name", to reflect the current metric, overwriting previous metrics' settings
     # of these fields (if any previous metric exists).
-    # When gs does NOT contain ci score (because CI was not computed for the current metric), but
+    # When global_score does NOT contain ci score (because CI was not computed for the current metric), but
     # one of the previous metrics computed did have, the last of such previous metrics set the values in
     # fields "score_ci_low" and "score_ci_high" in instance["score"]["global"] to reflect its
     # (the previous metric's) CI scores.
-    # Because CI is not computed for the current metric, gs does not contain fields "score_ci_low" and
-    # "score_ci_high" to overwrite the ones existing in instance["score"]["global"], and these might be left in
+    # Because CI is not computed for the current metric, global_score does not contain fields "score_ci_low" and
+    # "score_ci_high" to overwrite the ones existing in instance["score"]["global"], and these might remain in
     # instance["score"]["global"], but their values, that are not associated with the current metric, are,
     # therefore, not consistent with "score_name".
     # In such a case, following the python-dictionary-update, we pop out fields "score_ci_low" and
-    # "score_ci_high" from instance["score"]["global"], so that now all the fields "score.." are consistent with
-    # the current metric: "score_name" and "score". The current metric, named instance["score"]["global"]["score_name"],
-    # only has score value, which repeats in field instance["score"]["global"]["score"], and does not have ci_scores,
+    # "score_ci_high" from instance["score"]["global"], so that now all the fields "score.." in
+    # instance["score"]["global"] are consistent with the current metric: The current metric
+    # is named instance["score"]["global"]["score_name"], its score shows in
+    # field instance["score"]["global"]["score"], and it does not have ci_scores,
     # which is also reflected in the absence of fields "score_ci_low" and "score_ci_high" from instance["score"]["global"].
-    # If CI IS computed for the current metric, gs contains "score_ci_low" and "score_ci_high", and these overwrite
+    # If ci IS computed for the current metric, global_score contains "score_ci_low" and "score_ci_high", and these overwrite
     # the ones existing in instance["score"]["global"] by a simple python-dictionary-update, and no need for any further fixeup.
-    def update_and_adjust_global_score(self, instance: Dict[str, Any], gs: dict):
-        instance["score"]["global"].update(gs)
+    def update_and_adjust_global_score(
+        self, instance: Dict[str, Any], global_score: dict
+    ):
+        instance["score"]["global"].update(global_score)
         for score_ci in ["score_ci_low", "score_ci_high"]:
-            if score_ci in gs:
+            if score_ci in global_score:
                 continue
             if score_ci in instance["score"]["global"]:
                 instance["score"]["global"].pop(score_ci)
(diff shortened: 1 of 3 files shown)


yoavkatz (changes requested):

src/unitxt/metrics.py
```diff
@@ -231,6 +231,15 @@ def set_global_score(instances, global_score: Dict[str, Any]):
     def disable_confidence_interval_calculation(self):
         pass
 
+    def validate_ci_in_global_score(self, gs: dict):
```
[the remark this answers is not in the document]
dafnapension: Thanks, @yoavkatz , not quite clear to me. Please specify for me, as an example, what should be the expected target result in 
https://github.com/IBM/unitxt/blob/a0c39a5267da52e6d37e4ad904c93914b80c56a9/tests/library/test_metrics.py#L1712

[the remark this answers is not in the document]
yoavkatz: I would expect the the ci score to be calculated currently - and If it's not calculated for any reason . it should have nan:

       global_target = {
            "ensemble_0_precision_micro": 0.5,
            "ensemble_0_precision_micro_ci_high": 1.0,
            "ensemble_0_precision_micro_ci_low": 0.0,
            "ensemble_1_recall_macro": 0.33,
            "ensemble_1_recall_macro_ci_high": 0.56,
            "ensemble_1_recall_macro_ci_low": 0.0,
            "ensemble_score": 0.44,
            "score": 0.44,
            "score_ci_high": nan,
            "score_ci_low": nan,
            "score_name": "ensemble_score",
        }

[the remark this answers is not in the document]
yoavkatz: Why wasn't the CI score calculated in the first place?

src/unitxt/metrics.py
```diff
@@ -534,6 +543,7 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
 
         for instance in instances:
             instance["score"]["global"].update(global_score)
+            self.validate_ci_in_global_score(instance["score"]["global"])
```
dafnapension, line 546: Perhaps combine here the `.update(global_score)` with the new validation into one method ?

src/unitxt/metrics.py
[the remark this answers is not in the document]
dafnapension: Because the new `EnsembleMetric` comes without `ci_scores`, unlike most of the InstanceMetrics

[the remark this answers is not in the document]
dafnapension: And I think that to make your example consistent, then you should also add:
```
"ensemble_score_ci_low":  nan,
"ensemble_score_ci_high": nan
```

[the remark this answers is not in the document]
dafnapension: CI is not always calculated. Usually, to not calculate, one uses `metric.disable_ci_calculation`

src/unitxt/metrics.py
yoavkatz: It would be better, but I still want to avoid these special if conditions.   What if we change
image
So the result will be initialized to the scores with 'nan'?

yoavkatz: Or even simpler, create a ci variable with default (nan,nan) and then update the result in any case?

dafnapension: So you agree with my comment here: 
https://github.com/IBM/unitxt/pull/1065#discussion_r1694245144
?

I think that the global score might be overloaded. 
You wrote a special test "agaist" ci_scores that should not show in the global score..

dafnapension: In addition, I think that `nan` should be kept to some error in the calculation (like too few cases, or too many samples yielding nan global score).
If at all, perhaps start with something that indicates `not calculated`.  Perhaps a string value as is, `"not calculated"`, rather than a float.

yoavkatz: I see your point.  I agree we e should not have nan if it was have nan when it's not calculate.  

So how about replacing

        confidence_interval = self.compute_global_confidence_intervals(
            references, predictions, task_data, score_name
        )
        global_score.update(confidence_interval)

with
 
        confidence_interval = self.compute_global_confidence_intervals(
            references, predictions, task_data, score_name
        ) 
        update_confidence_interval(global_score, confidence_interval)  
      
      
 The update_confidence_interval would check if confidence_interval is {}, and if so, it would delete score_ci_low and score_ci_high (and update the global score otherwise with confidence_interval).
 
 
This would not require checking after the fact if the fields exists.

Sorry if I'm focusing on it, but unitxt code is complex as is, and I want to localize complex change.

dafnapension: No problem at all! Certainly nothing to be sorry about!

The issue is that we need to clean something that is already sitting in the global score (in each `instance["score"]["global"]`) . We do not see the whole problem just by looking at the newly computed `global_score` (the one just computed for the current metric, i.e., the one to be **updated into** the  `instance["score"]["global"]`-s).  We can suspect there might be a problem if the current metric does not compute CI, but we are not sure. We need to look back..

What you suggested in the last comment (https://github.com/IBM/unitxt/pull/1065#discussion_r1694255037) : 
`The update_confidence_interval would check if confidence_interval is {}, and if so, it would delete score_ci_low and score_ci_high (and update the global score otherwise with confidence_interval).`
means do look into `instance["score"]["global"]`, and delete if needed.
So I repeat my suggestion: to implement exactly what you say here, in one unified method of update-and-delete-whats-needed (https://github.com/IBM/unitxt/pull/1065#discussion_r1694243720)

yoavkatz: Ok. Got it. We need to delete instance["score"]["global"]  and not  from the global_score, which is just the current one.

LEt's go with what you suggest.

dafnapension: Thanks @yoavkatz , just pushed a unified method. Turns out that with the current `global_score` at hand (in addition to `instance["score"]["global]`), the check (which you rightfully described as complex) is dramatically simpler. So your instincts led us in the right direction, and I am glad that you insisted!
I also wrote a long explanation..

src/unitxt/metrics.py
```diff
@@ -231,6 +231,35 @@ def set_global_score(instances, global_score: Dict[str, Any]):
...
+    # instance["score"]["global"], but their values, that are not associated with the current metric, are,
+    # therefore, not consistent with "score_name".
+    # In such a case, following the python-dictionary-update, we pop out fields "score_ci_low" and
+    # "score_ci_high" from instance["score"]["global"], so that now all the fields "score.." are consistent with
+    # the current metric: "score_name" and "score". The current metric, named instance["score"]["global"]["score_name"],
+    # only has score value, which repeats in field instance["score"]["global"]["score"], and does not have ci_scores,
+    # which is also reflected in the absence of fields "score_ci_low" and "score_ci_high" from instance["score"]["global"].
+    # If CI IS computed for the current metric, gs contains "score_ci_low" and "score_ci_high", and these overwrite
+    # the ones existing in instance["score"]["global"] by a simple python-dictionary-update, and no need for any further fixeup.
+    def update_and_adjust_global_score(self, instance: Dict[str, Any], gs: dict):
```
yoavkatz, line 255: Looks good. Just one point.  Let's use 'global_score' instead of 'gs' , to be consistent with the rest of the code.

yoavkatz (approved):
I approved.  One minor comment about naming.

dafnapension:
Thanks, @yoavkatz , I renamed as you suggested, and merged once all tests passed.

PR #1068, opened 2024-07-28
Added check of type of format and system prompt  to LLM as judge

yoavkatz:


src/unitxt/llm_as_judge.py
```diff
@@ -132,6 +135,16 @@ def prepare(self):
             f"The supported tasks types are: {', '.join(supported_tasks)}."
         )
 
+        if self.format and not isinstance(self.format, Format):
```
OfirArviv, line 138: it can be also a string, no?

elronbandel (changes requested):
This can be done inherently with unitxt built in mechanism by setting:

```python
 format: Optional[Format] = None
 system_prompt: Optional[SystemPrompt] = None
```
Instead of:
```python
 format: Optional[str] = None
 system_prompt: Optional[str] = None
```
This will also fetch them upon instantiation in case they are strings @OfirArviv

elronbandel (approved):

PR #1070, opened 2024-07-29
Update CONTRIBUTING.md - update path name for activate

welisheva22:
Looks like the bin directory may have been changed to Scripts -- the command seems to work when it's written as Scripts instead of bin


elronbandel:
From python [documentation](https://docs.python.org/3/library/venv.html):

> It also creates a bin (or Scripts on Windows) subdirectory containing a copy/symlink of the Python binary/binaries (as appropriate for the platform or arguments used at environment creation time). It also creates an (initially empty) lib/pythonX.Y/site-packages subdirectory (on Windows, this is Lib\site-packages). If an existing directory is specified, it will be re-used.

welisheva22:
Thanks - then might it make sense to include both versions in the documentation, if some environments do use bin? How do you think we should handle this?

elronbandel:
writing both sounds ideal

yoavkatz:
Hi @welisheva22 - Can you update this PR with the above suggested changes?

welisheva22:
Sure - added another commit with the second command option, please review. Thank you.

yoavkatz (approved):

elronbandel (approved):

PR #1071, opened 2024-07-29
saving prediction-references and restore within PipelineMetric (including rag), allows to remove the deep copy from within ApplyMetric, and InstanceFieldOperator

dafnapension:


elronbandel:
What is the purpose? the issue with deep copy?

dafnapension:
Just to discipline metric.  No metric should be allowed to change prediction or references.

What is the problem with rag? Not that it changes prediction and references?

elronbandel:
Could you investigate it? I think that's part of the problem there

dafnapension:
Hi @elronbandel , please write a small piece of code for me that demonstrates that something is wrong in the current status.  Where is the current status? Why did you refer me to the PR of Binyamin?  Is he trying to do something related to main? Does the problem show in main too?

elronbandel (changes requested):
Instead of requiring developers for adopting to this mechanism could we just do it behind the scenes in the PipelineMetric internal code?

PR #1072, opened 2024-07-29
Added prediction type to llm as jusdge to avoid warning

yoavkatz:
Also Clarified the sandalone llm as judge example


src/unitxt/llm_as_judge.py
```diff
@@ -37,6 +37,7 @@ class LLMAsJudge(BulkInstanceMetric):
     inference_model: InferenceEngine
     reduction_map: Optional[Dict[str, List[str]]] = None
     batch_size: int = 32
+    prediction_type = Any  # Because handled with multiple tasks
```
yoavkatz, line 40:
```suggestion
    prediction_type = Any  # Because handles multiple tasks
```

OfirArviv (approved):

PR #1077, opened 2024-07-29
Rag no deep

dafnapension:


prepare/metrics/rag.py
```diff
@@ -349,16 +362,21 @@
...
+        postprocess_steps=[
+            RestorePredictionReferences(name=new_catalog_name),
+        ],
     )
     add_to_catalog(metric, new_catalog_name, overwrite=True)
 
 answer_reward = MetricPipeline(
     main_score="score",
     preprocess_steps=[
+        SavePredictionReferences(name="rag_answer_reward"),
```
yoavkatz, line 379: @dafnapension , @elronbandel - I don't think it's right to expect the user to save and restore values in metrics. Metrics pre-process should only be involved with the logic of the metic.

It has to be done in the generic  pipeline code.

PR #1082, opened 2024-07-30
try to remove deep add save/restore on top of main

dafnapension:


dafnapension:
copying over a comment by @yoavkatz :
[yoavkatz](https://github.com/yoavkatz) [17 hours ago](https://github.com/IBM/unitxt/pull/1077#discussion_r1695898633)
@dafnapension , @elronbandel - I don't think it's right to expect the user to save and restore values in metrics. Metrics pre-process should only be involved with the logic of the metic.

It has to be done in the generic pipeline code.

PR #1087, opened 2024-07-31
removed all central deepcopies, fixed locally only where needed, and verified that all (including rag end-to-end) runs well, in eager and non-eager mode, including catalog preparation (in both eager modes)

dafnapension:


prepare/metrics/normalized_sacrebleu.py
```diff
@@ -116,7 +116,7 @@
         "他方、成績評価の甘い授業が高く評価されたり、人気取りに走る教師が出たりし、成績の安売りや大学教師のレベルダウンという弊害をもたらす恐れがある、などの反省意見もある."
     ],
 ]
-task_data = len(predictions) * [{"target_language": "ja"}]
+task_data = [{"target_language": "ja"}, {"target_language": "ja"}]
```
dafnapension, line 119: before the change, in eager mode, the same 'ja' served both instances, and the second instance was updated automatically once the first instance was. I first (wrongly) interpreted this as duplicated activation of the operator.

dafnapension:
Hi @yoavkatz , is [rag-end-to-end](https://github.com/IBM/unitxt/blob/4fd07dc163afd6faa0498617ccd0f27349b466da/prepare/cards/rag/end_to_end/clapnq.py) the real scenario that you referred at in [this](https://github.com/IBM/unitxt/pull/991#issuecomment-2208928591) comment?
Is there another offending example, that made you add these deep copies, that I can try my fix (in this PR) on?

dafnapension:
Hi @elronbandel , all runs well after I changed per your suggestion.  No one is missing "prediction", that was probably when I was still half way through, so all is well. I generated catalog in both eager modes, and all passed well.
I am very curious about this PR running on heavy data, if you have.  Perhaps on heavy data, the expected gain in performance will be noticeable..

src/unitxt/metrics.py
```diff
@@ -1390,9 +1386,7 @@ def verify(self):
 class MetricPipeline(MultiStreamOperator, Metric):
     main_score: str = None
     preprocess_steps: Optional[List[StreamingOperator]] = field(default_factory=list)
-    postpreprocess_steps: Optional[List[StreamingOperator]] = field(
-        default_factory=list
-    )
+    postprocess_steps: Optional[List[StreamingOperator]] = field(default_factory=list)
```
elronbandel, line 1389:
```suggestion
    postpreprocess_steps: Optional[List[StreamingOperator]] = field(default_factory=list)
```
Non backward compatible should be in a separate pr

elronbandel (changes requested):
@dafnapension this is really good. we need to leave the renaming of the argument "post preprocessing steps" to a different pr . also delete all the lines you commented out.

src/unitxt/operators.py
```diff
@@ -202,13 +201,12 @@ def process(
 
     def get_mapped_value(self, instance, key, mapper, val):
         val_as_str = str(val)  # make sure the value is a string
-        if self.strict and (val_as_str not in mapper):
+        if val_as_str in mapper:
+            return mapper[val_as_str]
```
yoavkatz, line 205: Why is deep copy not needed?  What happens if we map a value to a list and then change the list?

dafnapension: Hi @yoavkatz , you probably mean if this happens inside `metric` (in `ApplyMetric` operator). We noticed that only `PipelineMetric` contains a shameless stepping over `prediction` and `references` (no matter through which step) as part of its `preprocess_steps` or `postprocess_steps`, so we save `prediction` and `references` as step0  of the `preprocess_steps` of `PipelineMetric`, and restore upon leaving. So `prediction` and `references` are saved against any stupidity - copy list and break there, or copy a dictionary and break there.

elronbandel (changes requested):
I agree with @yoavkatz, the main issue is cross instance references. If two instances have pointer to the same mutable variable then we expose the users to a very annoying type of bugs which are very hard to identify and very hard to fix. 

So the rule imho should be: two instances should not have a reference to similar mutable variable. This comes at price of compute but human frustration is more dear. 

Some implications to the rule:
1. Copy() can copy without deep copy as it by definition deals with instance-internal variables. 
2. MapFieldValues() must employ a mechanism that assign independent mutable variables to different instances. 

Another important implication: metrics are different because internal-variable changes can affect other metrics that expect unchanged instances. This means that the at the point of metrics invocation we should make sure every metric get the original untouched instance. 

From this perspective it is not the reposibiliy of the MetricPipeline to restore the instance but the responsibility of whoever invoked that metric. Meaning the unitxt invocation of metrics one after the other.

dafnapension:
So I need to close the PR, or use it to dive into the following problem, that `deepcopy` can not prevent, and we need to think of another solution:
```
>>> mi = 4*[{"a":1}]
>>> from copy import deepcopy
>>> mi = deepcopy(mi)
>>> mi
[{'a': 1}, {'a': 1}, {'a': 1}, {'a': 1}]
>>> mi[0]["b"]=2
>>> mi
[{'a': 1, 'b': 2}, {'a': 1, 'b': 2}, {'a': 1, 'b': 2}, {'a': 1, 'b': 2}]
>>> 
```

elronbandel:
Specifically in the example what happening is that the list is deep copied from the original list and maintaining its original structure where each element on the list are referring to the same dictionary. 

Again if this list is representing instances then the issue is originated in the cross references between instances in the original list. 

Now that we identified this as a problem maybe we should protect our users by ensuring each instance is separated from others in every possible instance creation, e.g in loaders, operator testing function operating on instances and so on.

My suggestion is to change the direction of this PR to enforce the rule of no cross reference between instances. Leaving deep copy where necessary (like MapInstanceValues) and adding deep copy where necessary (test_metric, test_operator, produce, DictLoader) and finally removing deep copy when unnecessary (Copy)

dafnapension:
I think that **any** intervention in the interpretation of the code, that does not comply with the `legal` interpretation of Python should be accompanied by a warning message.
I think that such interventions make a much worse source of agony when debugging. 
I just added one for the `MetricPipeline`

elronbandel:
I agree about performance. We can take few ideas from here: https://stackoverflow.com/questions/24756712/deepcopy-is-extremely-slow

dafnapension:
What is the offending metric that hits through `MapInstanceValues` ?

src/unitxt/operators.py
dafnapension: Hi @yoavkatz , The description of the damage implies `ListFieldValues` operator, not `MapInstanceValues` that your comment sits on its lines.  Isn't it?

dafnapension:
Covered by PR #1119

PR #1090, opened 2024-07-31
Add a metric to calculate the ratio of references included in the prediction

marukaz:
`StringContainment` is a metric that calculates a score based on whether references are contained in the prediction.
It scores 1.0 if any of references are in the prediction.
However, in some cases, the prediction should include all references, so I added another metric that calculates a score by ratio of references included in the prediction.


marukaz:
Sorry, I mistakenly changed a commit message. Close this and reopen another one

PR #530, opened 2024-01-30
Fix parser to allow source name that starts with numeric

marukaz:
close https://github.com/IBM/unitxt/issues/529

To allow source name that starts with numeric, I replaced `[a-zA-Z]` to `\w` in regex.

I read [CONTRIBUTING.md](https://github.com/IBM/unitxt/blob/main/CONTRIBUTING.md) but I don't have the permission to create a branch so I forked this repo.

PR #1091, opened 2024-07-31
Add a metric to calculate the ratio of references included in the prediction

marukaz:
`StringContainment` is a metric that calculates a score based on whether references are contained in the prediction.
It scores 1.0 if any of references are in the prediction.
However, in some cases, the prediction should include all references, so I added another metric that calculates a score by ratio of references included in the prediction.


yoavkatz:
Can you share more details where this is needed? What is the task, and what are fields of the task?

The definition (and invariant)  of references in Unitxt is that if  model returns one of the references , it should get a perfect score.  

In some case, we want the  model to return a list of values (e.g. in multi label classification), and then there is a single reference, which in itself is  a list of values.   If you want to model to return a single value,  it contains a set of values, then you can add to add a reference_fields to the task, with that list, and have the metric access it from the task_data argument.

marukaz:
Yes, our task is a bit unique. We're evaluating the results of semantic search for use in RAG, but unlike typical tasks of this nature, we don't have a reference for the context that should be extracted. Instead, we have a list of entities that should be included in the search results. Therefore, rather than token overlap or ROUGE, we want to evaluate how well the entities are included in the search results.

Also, in typical language generation tasks, a situation where such metrics might be necessary is in summary generation that includes specific keywords. This isn't what I'm currently working on, but when I was a student and doing research on summarization with a newspaper company, I heard that in summarizing news articles or generating headlines, there's often a need to ensure certain keywords are included. I think this metric would be useful for evaluating such tasks as well.

yoavkatz:
> Yes, our task is a bit unique. We're evaluating the results of semantic search for use in RAG, but unlike typical tasks of this nature, we don't have a reference for the context that should be extracted. Instead, we have a list of entities that should be included in the search results. Therefore, rather than token overlap or ROUGE, we want to evaluate how well the entities are included in the search results.
> 
> Also, in typical language generation tasks, a situation where such metrics might be necessary is in summary generation that includes specific keywords. This isn't what I'm currently working on, but when I was a student and doing research on summarization with a newspaper company, I heard that in summarizing news articles or generating headlines, there's often a need to ensure certain keywords are included. I think this metric would be useful for evaluating such tasks as well.

Thanks for the clarification.  Ok. So in this case, I think way to model it is as a this:

 
 card = TaskCard(
    # Load the data from the dictionary.  Data can be  also loaded from HF, CSV files, COS and other sources
    # with different loaders.
    loader=LoadFromDictionary(data=data),
    # Define the QA task input and output and metrics.
    task=Task(
        input_fields={"question": str},
        reference_fields={"entities": str},
        prediction_type=str,
        metrics=["field_string_containment_ratio[field=entities]"],
    ),
    # Create a simple template that formats the input.
    # Add lowercase normalization as a post processor.
    templates=TemplatesDict(
        {
            "simple": InputOutputTemplate(
                instruction="Answer the following question.",
                input_format="{question}",
                output_format="{entities}",
            )
        }
    ),
)

and  

lass StringContainmentRatio(InstanceMetric):
    """Metric that  returns the ratio of  values from a specific field  contained in the prediction.
    """

    reduction_map = {"mean": ["string_containment"]}
    main_score = "string_containment"
    ci_scores = ["string_containment"]
    field : str

    prediction_type = Any  # string representation is compared
    single_reference_per_prediction = False  # multiple references allowed
    score_by_ratio = False

    def compute(
        self, references: List[Any], prediction: Any, task_data: List[Dict]
    ) -> dict:
        result = {
            self.main_score: float(
                any(str(value) in str(prediction) for  value in task_data[self.field])
            )
        }
        contain_results = [
            str(reference) in str(prediction) for reference in references
        ]

         core = sum(contain_results) / len(contain_results)
 s
        result = {self.main_score: score}
        result["score"] = result[self.main_score]
        result["score_name"] = self.main_score
        return result

marukaz:
@yoavkatz Thank you for your suggestion, but it seems

```py
        result = {
            self.main_score: float(
                any(str(value) in str(prediction) for  value in task_data[self.field])
            )
        }
```

is overwritten by

```py
result = {self.main_score: score}
```

Maybe this is not intentional, right?
Do you mean

```py
        contain_results = [
            str(value) in str(prediction) fo rvalue in task_data[self.field])
        ]
```
?

yoavkatz:
> @yoavkatz Thank you for your suggestion, but it seems
> 
> ```python
>         result = {
>             self.main_score: float(
>                 any(str(value) in str(prediction) for  value in task_data[self.field])
>             )
>         }
> ```
> 
> is overwritten by
> 
> ```python
> result = {self.main_score: score}
> ```
> 
> Maybe this is not intentional, right? Do you mean
> 
> ```python
>         contain_results = [
>             str(value) in str(prediction) for value in task_data[self.field])
>         ]
> ```
> 
> ?

Yeah. Sorry.  I did not invest in making sure the code actually works.  I only wanted to show how you can access a reference field through task_data, and that this is the way to do it (not using the "references") field.

marukaz:
@yoavkatz Thanks! I updated code as you suggested

src/unitxt/metrics.py
```diff
@@ -1341,6 +1340,40 @@ def compute(
...
+                        reference_fields={"entities": str},
+                        prediction_type=str,
+                        metrics=["string_containment_ratio[field=entities]"],
+                    )
+    """
+
+    reduction_map = {"mean": ["string_containment"]}
+    main_score = "string_containment"
+    ci_scores = ["string_containment"]
+    field: str = "entities"
```
yoavkatz, line 1360: Thanks. Looks good. I suggest not having a default, so people won't forget to set it by mistake.

yoavkatz: And add a check that is set.

prepare/metrics/string_containment.py
```diff
@@ -31,3 +31,36 @@
...
+outputs = test_metric(
+    metric=metric,
+    predictions=predictions,
+    references=[["dummy"] for _ in references],
+    instance_targets=instance_targets,
+    global_target=global_target,
+    task_data=[{reference_field: w} for w in references],
+)
+
+add_to_catalog(metric, "metrics.string_containment.ratio", overwrite=True)
```
yoavkatz, line 66:
```suggestion
add_to_catalog(metric, "metrics.string_containment_ratio", overwrite=True)
```

Seems it would be easier to find this way in the catalog (not nested).

src/unitxt/metrics.py
marukaz: Yes, I tried it first, but I got

```
❯ python prepare/metrics/string_containment.py
Metric tested successfully!
Artifact metrics.string_containment saved to /home/marukaz/ghq/github.com/marukaz/unitxt/src/unitxt/catalog/metrics/string_containment.json
  File "/home/marukaz/ghq/github.com/marukaz/unitxt/prepare/metrics/string_containment.py", line 40, in <module>
    metric = StringContainmentRatio()
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marukaz/ghq/github.com/marukaz/unitxt/src/unitxt/dataclass.py", line 349, in custom_cls_init
    original_init(self, *args, **kwargs)
  File "/home/marukaz/ghq/github.com/marukaz/unitxt/src/unitxt/dataclass.py", line 349, in custom_cls_init
    original_init(self, *args, **kwargs)
  File "/home/marukaz/ghq/github.com/marukaz/unitxt/src/unitxt/dataclass.py", line 349, in custom_cls_init
    original_init(self, *args, **kwargs)
  [Previous line repeated 5 more times]
  File "/home/marukaz/ghq/github.com/marukaz/unitxt/src/unitxt/dataclass.py", line 504, in __init__
    raise RequiredFieldError(
unitxt.dataclass.RequiredFieldError: Required field 'field' of class StringContainmentRatio not set in StringContainmentRatio
```

then I also tried to set a value when creating the instance like `metric = StringContainmentRatio(field=reference_field)`. In that case, the json became

```
{
    "__type__": "string_containment_ratio",
    "field": "entities"
}
```

so I thought  this became the fixed value.

marukaz: In fact, [BertScore](https://github.com/IBM/unitxt/blob/9c4002caa0e2311eb542823ca2ab43122b3202f1/src/unitxt/metrics.py#L2505) has an attribute called `model_name` that doesn't have a default value. However, the model name is specified in the [prepare](https://github.com/IBM/unitxt/blob/9c4002caa0e2311eb542823ca2ab43122b3202f1/prepare/metrics/rag.py#L6), and looking at the task example in the [documentation](https://github.com/IBM/unitxt/blob/9c4002caa0e2311eb542823ca2ab43122b3202f1/docs/docs/adding_metric.rst), it doesn't seem to be implemented in a way that makes `model_name` mandatory.

marukaz: I may be misunderstanding something as I haven't fully grasped the principles of unitxt. I'd be grateful if you could help me 🙇

marukaz: Should I set default of `field` to `None` and check that it is not `None` in `compute()`?

yoavkatz: Yes.  Setting it to none and checking sound right.  Please use override the verify() method.

def verify(self):
    super().verify()
    If self.field is None:
         .....

yoavkatz: > I may be misunderstanding something as I haven't fully grasped the principles of unitxt. I'd be grateful if you could help me 🙇

Yes.  Unitxt has some principle that are not immediate to understand.  They have their benefits in the ability to reuse and supports a large variety of verification needs, but we continuously try to simplify and better document them.

marukaz: Thank you! I updated code

yoavkatz:
Today, the when the catalog consistency fails one gets this error message.

```
Comparing generated and old catalog...
***** Directories has differences ******
{'file': '/home/runner/work/unitxt/unitxt/src/unitxt/catalog/metrics/string_containment', 'diff': 'old only'}
  File "/home/runner/work/unitxt/unitxt/utils/prepare_all_artifacts.py", line 165, in <module>
    main()
  File "/home/runner/work/unitxt/unitxt/utils/prepare_all_artifacts.py", line 160, in main
    raise RuntimeError("Directories has differences")
RuntimeError: Directories has differences
```

This means (we will update the error message)

```
Catalog consistency check failed.  The catalog json files were regenerated from the prepare/*/*.py files.  

* Some json files (or complete directories) which were committed in the past were not generated:

 home/runner/work/unitxt/unitxt/src/unitxt/catalog/metrics/string_containment
 
This can happen when renaming to catalog names of artifacts in  prepare/*/*.py files , but not deleting the old files. 
 
 To fix recreate and fix this in your local environment:
 
 rm -fr  src/unitxt/catalog
 python utils/prepare_all_artifacts.py
  ````

**Comment by yoavkatz** [comment]
I wanted to fix  but I don't access to your fork.  In general it's best to work in a branch , so we can more easily access and change.  Also note we need to commit changes to be signed off:

https://github.com/IBM/unitxt/blob/main/CONTRIBUTING.md#legal

File: src/unitxt/metrics.py [hunk 3]
```diff
@@ -1341,6 +1340,47 @@ def compute(
...
+        result = {self.main_score: score}
+        result["score"] = result[self.main_score]
+        result["score_name"] = self.main_score
+        return result
+
+    def verify(self):
+        super().verify()
+        if self.field is None:
+            raise ValueError(
+                "StringContainmentRatio metric requires the field attribute to be set."
```
yoavkatz, line 1380:
```suggestion
                "StringContainmentRatio metric requires the 'field' attribute to be set."
```

src/unitxt/metrics.py
```diff
@@ -1341,6 +1340,47 @@ def compute(
...
+    main_score = "string_containment"
+    ci_scores = ["string_containment"]
+    field: str = None
+
+    prediction_type = Any  # string representation is compared
+
+    def compute(
+        self, references: List[Any], prediction: Any, task_data: List[Dict]
+    ) -> dict:
+        contain_results = [
```
yoavkatz, line 1367: if self.field not in  task_data:
    raise ValueError(f"'{field}' field required by {__class__.__name__} is not in passed in task_data: {task_data}")

yoavkatz (approved):
I approved.  Added one minor request for an error message.

marukaz:
@yoavkatz 

> In general it's best to work in a branch , so we can more easily access and change

Yes, I wanted to do that too, and when I [first created a pull request](https://github.com/IBM/unitxt/pull/530#issue-2107340017) for this repository, I actually tried to create a branch. However, I didn't have write permissions for this repository, so I couldn't create one. There was nothing specifically written about this in CONTRIBUTING.md, and since I had been contributing via forks until today without anyone saying anything, I thought the policy was not to give write permissions to external contributors. How can I obtain the necessary permissions?

PR #1103, opened 2024-08-01
Simplify HF vs Direct APIs

yoavkatz:
When using HF APIs only the HF version code and catalog will be used


yoavkatz:
HF API will be deprecated in the future.

PR #1111, opened 2024-08-05
Add logprobs functionality

arielge:


src/unitxt/processors.py
```diff
@@ -277,3 +279,57 @@ def process_value(self, text: Any) -> Any:
 
         except:
             return 0
+
+
+class InferDictsToBinaryLogprobs(FieldOperator):
+    binary_class_names: list[str]
```
yoavkatz, line 285: This is a bit subtle code ,that is  worth documentation.

If I understand right, the approach taken is to get the model to generate a free-form  textual answer (but also ask it to return the top tokens log prob per position) .  Then we find the first position which has at least one of binary choices in the top tokens list.  Then we estimate the probability from the log_probs.  

The benefit of this approach is that it requires only 1 inference per call and fits the existing mechanism.
An alternative run inference twice, once with a forced "Yes" completion and once with a "No" completion. 

"Is this the above anaser faithful to the context? Yes" <-- take the log prob of the completion 
"Is this the above anaser faithful to the context? No" <-- take the log prob of the completion

Then you can be sure you get both option.

src/unitxt/processors.py
```diff
@@ -277,3 +279,57 @@ def process_value(self, text: Any) -> Any:
...
+
+
+class InferDictsToBinaryLogProbsLastToken(InferDictsToBinaryLogprobs):
+    def process_value(self, obj: Any) -> Any:
+        try:
+            n_last_tokens = min([3, len(obj)])
+        except:
+            return []
+
+        for i in range(-1, -(n_last_tokens + 1), -1):
```
yoavkatz, line 326: rather than duplicate the code between the classes, you can have a  only method to return the range you traverse .

src/unitxt/processors.py
yoavkatz: I'm not sure what I prefer.  If your approach works in most cases, maybe it's simpler.  However, the model may not return "yes" or "no" anywhere in the answer.

src/unitxt/processors.py
```diff
@@ -277,3 +279,57 @@ def process_value(self, text: Any) -> Any:
...
+            except:
+                pass
+        return []
+
+    def get_pos_neg_probs(self, pred_dict):
+        token_logprobs = pred_dict["top_tokens"]
+        pos_probs = [
+            np.exp(d["logprob"])
+            for d in token_logprobs
+            if self.binary_class_names[1].lower() in d["text"].lower()
```
yoavkatz, line 309: What happens is the token text  is "noone" - it will include "no" in the answer.

src/unitxt/metrics.py
```diff
@@ -567,11 +567,12 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
 
         result = self._compute(references, predictions, task_data)
         global_score.update(self._add_score_prefixes_to_score_dict(result))
-        score_name = global_score["score_name"]
-        confidence_interval = self.compute_global_confidence_intervals(
-            references, predictions, task_data, score_name
-        )
-        global_score.update(confidence_interval)
+        score_names = self.ci_scores if self.ci_scores else [global_score["score_name"]]
```
yoavkatz, line 570: This is an unrelated changes of allowing ci_scores on multiple scores in global metrics.

src/unitxt/inference.py
```diff
@@ -466,19 +469,48 @@ def verify(self):
...
             )
             for instance in dataset
         ]
+
+    def _infer_log_probs(self, dataset):
+        model, params = self._load_model_and_params()
+
+        params = {
+            **params,
+            "return_options": {
```
yoavkatz, line 500: I din't think this needs to be hard coded.  It can be set in the params externally.

src/unitxt/processors.py
arielge: Your understanding is correct, here we are adding support for this implementation (which seems to work reasonably well) and if someone wants they could add the alternative one as well.

src/unitxt/processors.py
arielge: Yes, this can happen, although not super likely given a prompt that asks for yes or no.
But also, there are several different alternatives of how to say yes and no depending on the model as well as the tokenizer, and often there are special tokens somewhere in there because it is the first word. Specifically tokens that start with space may appear in various ways for different tokenizers and inference APIs. Probably a more complicated regex would be needed to cover all of them.

yoavkatz: I also saw tokens with spaces or period.  That's why a minimal regex is needed.

src/unitxt/inference.py
arielge: It can, although one would need to specify this specific combination of parameters for this to work. Hard coding removes this burden from the user.

src/unitxt/processors.py
yoavkatz, line 285: If the code supports only two classes, let's call it pos_class_name, and neg_class_name.  
Otherwise, you need to add validation (e.g. the list has only two elements).

Another validation is to check that the two texts don't overlap ("Ok","Not Ok")

src/unitxt/processors.py
```diff
@@ -277,3 +279,57 @@ def process_value(self, text: Any) -> Any:
...
+
+
+class InferDictsToBinaryLogprobs(FieldOperator):
+    binary_class_names: list[str]
+
+    def process_value(self, obj: Any) -> Any:
+        try:
+            n_first_tokens = np.min([3, len(obj)])
+        except:
+            return []
```
yoavkatz, line 291: How could this code fail?

src/unitxt/processors.py
```diff
@@ -277,3 +279,57 @@ def process_value(self, text: Any) -> Any:
...
+        for i in range(n_first_tokens):
+            try:
+                pos_probs, neg_probs = self.get_pos_neg_probs(pred_dict=obj[i])
+                if pos_probs or neg_probs:
+                    sum_probs = sum(pos_probs) + sum(neg_probs)
+                    if sum_probs > 0.0001:
+                        return sum(pos_probs) / sum_probs
+            except:
+                pass
+        return []
```
yoavkatz, line 302: We should decide what this post processor return.  I think it's the probably of the positive class.
So if it's not able to find anything it should return np.nan.

prepare/processors/processors.py
```diff
@@ -369,3 +371,35 @@
...
+add_to_catalog(
+    SequentialOperator(
+        steps=[
+            InferDictsToBinaryLogProbsLastToken(
+                binary_class_names=(
+                    "No",
+                    "Yes",
+                ),
+                field="prediction",
+                process_every_value=False,
```
elronbandel, line 400: Notice that we have a field called "options" assigned by many templates (such as the classification templates and multiple choice templates) with the options for completion based on the verbalization of the template. This is a safer way to do it because sometimes when you switch the template it's now "True" "False" and not "Yes" "No" and it's hard to pass it on. That's why so far all the verbalization reposibilty was in the template and externalizing verbalization to anywhere else in the pipeline can lead to issues. If you need help to integrate with the current mechanism of the "options" field you can either contact me or @perlitz who did an extensive use of it in the past specifically in the context of logprobs.

elronbandel (changes requested):
This PR has a better existing solution (detailed in the comments) and should be handled differently IMHO based on the similar solutions for logprobs we did in the past.

src/unitxt/processors.py
arielge: Not sure, but it can. @lilacheden do you remember?

prepare/processors/processors.py
arielge: @elronbandel Thanks, but I am not sure I follow how this post-processor is related to the template fields. We want this to be used on the inference engine predictions, across many templates for which this pattern would be applicable.
BTW, there are some other processors like this that have specific strings they expect, for example YesToOneElseZero, StanceToProCon

src/unitxt/processors.py
arielge: so I guess something more like `^(\W|\\n)?[yY]es(\W|\\n)?$`? not sure

src/unitxt/processors.py
lilacheden: It failed when obj was None due to some runtime issue. Maybe it deserves a separate check.

yoavkatz: Ok. Let's remove this and fix the problem when it occurs. References and predictions should not be none.

src/unitxt/inference.py
yoavkatz: Yes. But sometimes people want to include input tokens in the results and sometime not.  Also the number of top predictions may change .  I suggest that the code will fail with a clear error message, if it does not get what it expects.

arielge: They do not really have the option - you must set either `input_tokens=True` or `input_text=True`, otherwise the client returns an error. And if you set `input_text=True,` the input is included in the "generated_text" field, which can cause confusion and inconsistent behavior. So given that this is the infer_with_logprobs method, and this is the only real way to get logprobs, I still think it makes more sense to pass these parameters than expect the user to figure out for themselves that it is the way to do this. The only real degree of freedom here is the top_n tokens, where the maximum one can request is 5, but if you really want you can ask to get less.

yoavkatz: I understand. Some points:
1. I would think it would be possible to possible to return only the generated token (which is what you need)

image

If not, it's a bug.

If we select to  return the input_tokens, then it will include the "Yes"/"No" tokens of the prompt.  Is that why you had the search for end of the token list?  
 
1 . we need the behavior to be consistent across all inference engines (OpenAI ).  So if one inference engine sets the default params, the others should too.  Note that people may want to use the log probabilities in different ways than yours.

2. If we s set parameters We need to check if the user set these parameters already, an if so raise an error / warning.

arielge: 1. Indeed looks like a bug. I saw the API documentation but in practice the behavior is different. 
image
2. Certainly, we can raise an error if the user tries to pass the wrong parameters. But given that this specific API behaves in a very unpredictable and undocumented manner currently (unlike say OpenAI), I really think we should have a default behavior in this method.

arielge: Regarding searching the end of the token list - the processor that searches from the end is for a situation where the answer is expected to come at the end (say, if the model is asked to generate an explanation before answering).

prepare/processors/processors.py
yoavkatz: I agree , it need not use a task "options" field.
A processor is tied to a template and not necessary to a task.

elronbandel: The "options" field (which is not a task field but an internal unitxt field)  was our way to use it internally for getting log probs. It takes into account template nuances such as whether the model is expected to output the choice numeral (a. 1. etc) in multiple choice, and if we have target prefix. All I suggest is that you fully understand the current mechanism as it was developed very carefully.

src/unitxt/processors.py
arielge: done

src/unitxt/processors.py
arielge: done

src/unitxt/processors.py
arielge: done

src/unitxt/processors.py
arielge: done

src/unitxt/processors.py
arielge: done

src/unitxt/processors.py
```diff
@@ -277,3 +279,59 @@ def process_value(self, text: Any) -> Any:
...
         except:
             return 0
+
+
+class InferDictsToBinaryLogprobs(FieldOperator):
+    neg_class_name: str
+    pos_class_name: str
+
+    take_logprobs_from_end: bool = False
+    num_logprobs_to_take: int = 3
```
yoavkatz, line 289: Thanks for making the changes.  If I understand right  this attribute is the index of the first token to search for the classes in.  If so the name should reflect it,. Also why start from 3, and not from the first token?

arielge: This is not the index of the first token, but the total number of tokens to look at. So if the value is 3 we look at index 0,1,2 (in regular case) or -1,-2,-3 (if we look at tokens at the end rather than the beginning). This is because in the beginning/end there are often special characters, newlines etc. Once we get to a token where the class names are candidates we stop and return the values at this index.

src/unitxt/inference.py
arielge: done

src/unitxt/metrics.py
yoavkatz: IS this right?

prepare/processors/processors.py
yoavkatz: Yes. I was not too familiar with the option mechanism (we should document it).

Looking at the code. It's not used in  unitxt log prob inference but in the internal package inference.
"options" are  added to task/data  in the MultipleChoiceTemplate (if they are not already in the task).

In any case, it's a different scenario in the LLM as judge, where the user defines a template

"Is the following sentence correct? Answer 'with one word (Yes' or 'No')

'The sky is blue'
"

The model returns

"Yes it is."

And the post processor wants to return the probability of "Yes" vs "No" in the relevant token, (the first) so it converts the text into a float score.

Perhaps if force the model to use a MultipleChoiceTemplate, it would have the options fields - but do we want this for LLM as Judge? It seems cumbersome.

src/unitxt/processors.py
yoavkatz: Ok. Now I understand. Thanks.

yoavkatz:
This PR has 4 changes:
1. Added log inference to WML inference engine
2. Add ability to load use space_id and not project_id credentials to WML.
3. Add ability to calculate confidence interval for multiple scores
4. A post processor to convert log probs dictionary to probabilities of specific class

@elronbandel - I think your concerns are only around (4).

@pawelknes - can you please review (1) + (2)?  @elronbandel - do you  review (3)?

Regarding (4) - I think @elronbandel thinks it is better addressed in an inference engine, but right now it's very localized changes.


@arielge - can you say how you are going to use the the new post processor?

src/unitxt/inference.py
```diff
@@ -449,7 +457,10 @@ def _initialize_wml_client(self):
             self.credentials = self._read_wml_credentials_from_env()
 
         client = APIClient(credentials=self.credentials)
-        client.set.default_project(self.credentials["project_id"])
+        if "space_id" in self.credentials:
```
pawelknes, line 460: I see that there may be an issue when a user passes credentials directly to the engine as an instance of `ibm_watsonx_ai.credentials.Credentials` (which is allowed). This class does not have either `space_id` or `project_id` as its attributes. 
So one way could be to require `credentials` to be always a dictionary. The other solution could be to sill allows passing an instance of `Credentials` but also demand providing any of `project_id` or `space_id` directly to the engine

As for inference with logprobs, it looks good to me

lilacheden: @pawelknes - The project_id/space_id in the credentials are set by the PROJECT_ID/SPACE_ID env variables, if they exist. We can add a check to set them according to WML_PROJECT_ID/WML_SPACE_ID env variables if the credentials object doesn't contain either project_id or space_id.

pawelknes: Alright, there is also another PR with minor fixes to the engine so that'll resolve any related issue anyway

elronbandel:
Regarding (4) The desired design here is described in this issue: https://github.com/IBM/unitxt/issues/1128

The benefit of implementing it this way is that the algorithm @arielge is suggesting will be compatible with many tasks such as multiple choice tasks and classification tasks. And will work out of the box with all the related templates and tasks without requiring any modifications.

arielge:
Thanks @yoavkatz @elronbandel. The intention was to use the post processor as part of a flow that includes an inference engine. So for example you can have a judge metric that is initialized with a specific template and task, processes the user inputs with the given template and sends to the inference engine. Then the inference engine outputs are fed to the post-processor to get the desired score.
@elronbandel I understand your desire to integrate this into the inference engine. This is also a possibility, although I assume different users may want slightly different things (infer for every class is one option, infer once and look at the top probabilities is another; and maybe binary and multi-class will not be treated the same). Currently what we are aiming for is flexibility to mix and match inference engines, templates and post-processors, without necessarily imposing our approach on the inference engine logic or requiring extra fields (as we are dealing specifically with the yes/no answer use case and not a general multi-class classification solution).

arielge:
superseded by #1243 and #1205

PR #1117, opened 2024-08-07
postpreprocess_steps -> postprocess_steps

dafnapension:


elronbandel (changes requested):
We should maintain support for the old name and mark it with DeprecatedField (you created If you remember)

src/unitxt/metrics.py
```diff
@@ -1412,6 +1416,13 @@ def verify(self):
 
     def prepare(self):
         super().prepare()
+        if (
+            hasattr(self, "postpreprocess_steps")
+            and self.postpreprocess_steps is not None
+            and isinstance(self.postpreprocess_steps, list)
+            and len(self.postpreprocess_steps) > 0
+        ):
+            self.postprocess_steps = self.postpreprocess_steps
```
yoavkatz, line 1425: Need to check that self.postproces_step is not set (throw error if both are set).

dafnapension: Done.

elronbandel (approved):

Issue #987, opened 2024-07-03
Slow performance due copying of instances


yoavkatz:
Current dict_utils performs overwrite of dictionaries:

 ```
if query.strip() == "":
        # change the whole input dic, as dic indeed matches ""
        if isinstance(dic, dict):
            if value is None or not isinstance(value, dict):
                raise ValueError(
                    f"Through an empty query, trying to set a whole new value, {value}, to the whole of dic, {dic}, but value is not a dict"
                )
            dic.clear()
            dic.update(value)
            return

        if isinstance(dic, list):
            if value is None or not isinstance(value, list):
                raise ValueError(
                    f"Through an empty query, trying to set a whole new value, {value}, to the whole of dic, {dic}, but value is not a list"
                )
            dic.clear()
            dic.extend(value)
            return
```
            
 This caused problems because operators made changes to data of their source and it cause conflict with other operators. To workaround this, we now copy instances :
 
 
 
 ```
class InstanceFieldOperator(InstanceOperator):
 
 def process(
        self, instance: Dict[str, Any], stream_name: Optional[str] = None
    ) -> Dict[str, Any]:
        # Need to deep copy instance, because when assigning two dictionary fields,
        # dict_set() the target field dictionary fields.
        # This means that if this target field was assigned to another field before,
        # the field is updated as well.
        instance = deepcopy(instance)
    
    
```
However this cause performance issue with large datasets (6x runtime increase)  We should change the code such that new instances are returned with shallow copy - and only relevant atomic are copied.

elronbandel:
To me operators should modify dictionaries in place, the reason is that that's how most people work with dictionaries. In cases where this creates confusion we should avoid it. Also I think we should have a separate issue for creating performance benchmark to unitxt we run on different pipelines, so we can keep track over performance across PRs.

yoavkatz:
The problem is when operators modify nested dictionary

instance:
     predictions : { "a" : 3, "b" :4}
     references : [{ "a" : 5, "b" :6}]
      
 Then we have an operator Copy(field="predictions", to_field="orig_predictions"}
 
 instance:    
         predictions : { "a" : 3, "b" :4}
         orig_predictions = { "a" : 3, "b" :4} (same object as above)
          references : [{ "a" : 5, "b" :6}]
     
  Now we do Copy(field = "references/0/a" to, "predictions/a")
  
  instance:
         predictions : { "a" : 5, "b" :4}
[...]

Issue #1069, opened 2024-07-28
Add warning when metrics scores overwrite each other.


yoavkatz:
Right now , metrics score overwrite each other (Not referring to the 'score' and 'score_name' , but two metrics can have the same 'f1' score.   

Users find it confusing.  We should at first step make this into a warning.

"Metric 'token_overlap' has the same score as another metric in the metrics list ['token_overlap', 'macro_f1'] .  To avoid overwriting the value , add a score_prefix to the metric (e.g. score_prefix="token_overlap_") .   For more information, see ... "

dafnapension:
Resolved with PR #1124

PR #1119, opened 2024-08-08
Rag no deep 2

dafnapension:
This goes along with @elronbandel 's comment:  (https://github.com/IBM/unitxt/pull/1087#pullrequestreview-2222076506)
```
Another important implication: metrics are different because internal-variable changes can affect other metrics that expect unchanged instances. This means that the at the point of metrics invocation we should make sure every metric get the original untouched instance.

From this perspective it is not the reposibiliy of the MetricPipeline to restore the instance but the responsibility of whoever invoked that metric. Meaning the unitxt invocation of metrics one after the other.
```

In other words: unitxt intervenes in python's interpretation, changing 'natural' meaning, only where unitxt is responsible: the launching of metric after metric, assuring for each metric that it is treated as an 'only child' and can change the instance as it sees fit; it is aware and it bears the consequences.. and all -- only for itself, as unitxt encapsulates it from the other metrics.

Other surprising deepcopies, like in `InstanceFieldOperator`, for each and every invocation thereof, are removed. And so is the deepcopy in `ListFieldValues` that is surprising, not part of the definition, and not explicitly requested.    

All in all, performance is expected to improve, while 'observing the law'..
(in reference to issue https://github.com/IBM/unitxt/issues/987 )

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -1787,12 +1787,15 @@ class ApplyMetric(StreamOperator, ArtifactFetcherMixin):
         from .metrics import Metric
 
         # Number of instances in input stream is assumed to be small. This is why
-        # each metric consumes all of them and lays in its main memory, and even generates some 1000
-        # copies thereof for the sake of CI.
-        # So we start with deep copying here, to make a 'frozen' status of the stream, a status to be
-        # applied to each of the metrics listed in metric_field, so that the evaluation of one does not
-        # affect the evaluation of another (typically, affecting via change of
-        # instance as part of preprocess_steps of MetricPipeline).
+        # each metric consumes all of them and lays them in its main memory, and even generates
+        # some 1000 copies thereof for the sake of CI.
+        # So we start with deep copying here, to make a 'frozen' status of the stream, having
+        # passed the preprocess_steps of the task, and inference, and now getting to be evaluated,
+        # a frozen status to be fed into each of the metrics listed in metric_field,
+        # so that the evaluation of one does not affect the evaluation of another
+        # (typically, affecting via change of instance as part of
+        # preprocess_steps of MetricPipeline, as illustrated in docs/adding_metrics/Using Metric Pipelines).
+
         instances_upon_entrance_to_metrics_evaluations = []
         for instance in stream:
             instances_upon_entrance_to_metrics_evaluations.append(deepcopy(instance))
@@ -1825,9 +1828,8 @@ class ApplyMetric(StreamOperator, ArtifactFetcherMixin):
             multi_stream = MultiStream(
                 {
                     "tmp": ListStream(
-                        instances_list=deepcopy(
-                            instances_upon_entrance_to_metrics_evaluations
-                        )
+                        instances_list=instances_upon_entrance_to_metrics_evaluations,
+                        copying=True,  # ensures deep copy when iterating over instances
                     )
                 }
             )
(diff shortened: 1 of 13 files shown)


src/unitxt/settings_utils.py
```diff
@@ -126,7 +126,7 @@ def __getattr__(self, key):
     settings.artifactories = None
     settings.default_recipe = "standard_recipe"
     settings.default_verbosity = "info"
-    settings.use_eager_execution = False
+    settings.use_eager_execution = True
```
elronbandel, line 129:
```suggestion
    settings.use_eager_execution = False
```

src/unitxt/operators.py
```diff
@@ -1822,17 +1823,24 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
 
             if not self.calc_confidence_intervals:
                 metric.disable_confidence_interval_calculation()
-
+            multi_stream = MultiStream(
+                {
+                    "tmp": ListStream(
```
elronbandel, line 1828:
```suggestion
                    "tmp": ListStream(
                               copying=True,
```

src/unitxt/operators.py
```diff
@@ -1787,7 +1787,18 @@ class ApplyMetric(StreamOperator, ArtifactFetcherMixin):
...
+        # copies thereof for the sake of CI.
+        # So we start with deep copying here, to make a 'frozen' status of the stream, a status to be
+        # applied to each of the metrics listed in metric_field, so that the evaluation of one does not
+        # affect the evaluation of another (typically, affecting via change of
+        # instance as part of preprocess_steps of MetricPipeline).
+        instances_upon_entrance_to_metrics_evaluations = []
+        for instance in stream:
+            instances_upon_entrance_to_metrics_evaluations.append(deepcopy(instance))
+
+        first_instance = instances_upon_entrance_to_metrics_evaluations[0]
```
elronbandel, line 1801:
```suggestion
        stream.copying = True
        instances_upon_entrance_to_metrics_evaluations = list(stream)
```

elronbandel: Maybe we should implement:
```python
list(stream.deepcopy())
```

src/unitxt/operators.py
```diff
@@ -1822,17 +1823,24 @@ def process(self, stream: Stream, stream_name: Optional[str] = None) -> Generato
...
+                }
+            )
             multi_stream = metric(multi_stream)
-            multi_stream = CopyFields(
-                field_to_field={f"{k}_orig": k for k in keys_to_restore}
-            )(multi_stream)
+            for evaluated_instance, freezed_instance in zip(
+                multi_stream["tmp"], instances_upon_entrance_to_metrics_evaluations
+            ):
+                freezed_instance["score"] = deepcopy(evaluated_instance["score"])
```
elronbandel, line 1839: Maybe add warning when score with specific name is being overridden?
https://github.com/IBM/unitxt/issues/1069

elronbandel: UnitxtWarning("score x from metric x_metric override score y from metric y_metric, consider adding score prefix to one of the metrics.",  link_to_prefix_docs, )

src/unitxt/stream_operators.py
```diff
@@ -155,5 +156,5 @@ class DuplicateSplit(MultiStreamOperator):
     def process(self, multi_stream: MultiStream) -> MultiStream:
         assert self.split in multi_stream
         generators = multi_stream
-        generators[self.to_split] = generators[self.split]
+        generators[self.to_split] = copy.deepcopy(generators[self.split])
```
elronbandel, line 159:
```suggestion
        stream_copy = copy.deepcopy(multi_stream[self.split])
        stream_copy.copying = True
        multi_stream[self.to_split] = stream_copy
        return multi_stream
```

elronbandel: Explanation: when copying stream we need to make sure the instances are independent from the origin stream instances

dafnapension:
May be re-considered when the excessive deepcopies will start hindering performance, or confuse users by their unexpected interpretation of copy as deepcopy.

dafnapension:
Try to profile the performance difference

dafnapension:
Going in loops when trying to rebase on main. Will generate a whole new branch from main

PR #1123, opened 2024-08-08
Refactor RenameFields to Rename with appropirate deprecation process

elronbandel:
Signed-off-by: elronbandel <elron.bandel@ibm.com>


dafnapension:
I think that the `field_to_field` option is the only one that allows a programmatic use, e.g.:
https://github.com/IBM/unitxt/blob/7a337453f27db5dda64cff8cbe3b69b4e2e41ca7/src/unitxt/operators.py#L1827

PR #1126, opened 2024-08-08
Update summarization task and templates to support multiple references summaries - non backward compatible change

yoavkatz:
Will impact cards and templates defined outside unitxt on these task.

Changes task field from summary (str) to summaries (List[str])

### DIFF
--- a/prepare/cards/xsum.py
+++ b/prepare/cards/xsum.py
@@ -9,7 +9,7 @@ from unitxt.test_utils.card import test_card
 card = TaskCard(
     loader=LoadHF(path="EdinburghNLP/xsum"),
     task="tasks.summarization.abstractive",
-    preprocess_steps=[Wrap(field="summaries", inside="list", to_field="summaries")],
+    preprocess_steps=[Wrap(field="summaries", inside="list")],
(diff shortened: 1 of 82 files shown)


prepare/cards/xsum.py
```diff
@@ -3,11 +3,13 @@
     TaskCard,
 )
 from unitxt.catalog import add_to_catalog
+from unitxt.collections_operators import Wrap
 from unitxt.test_utils.card import test_card
 
 card = TaskCard(
     loader=LoadHF(path="EdinburghNLP/xsum"),
     task="tasks.summarization.abstractive",
+    preprocess_steps=[Wrap(field="summaries", inside="list", to_field="summaries")],
```
elronbandel, line 12:
```suggestion
    preprocess_steps=[Wrap(field="summaries", inside="list")],
```
equivalent

elronbandel (commented):
Have we got any datasets that are multi reference?

yoavkatz:
> Have we got any datasets that are multi reference?

Not here, but there are internal datasets 
with multiple reference. We want to be have one common task definition.

elronbandel (approved):

PR #1134, opened 2024-08-12
Add Scigen table2text task with llm_as_judge metric

csrajmohan:
Add Scigen dataset for table2text task with llm_as_judge metric using llama3 as judge model
This includes,
* addition of scigen task card
* addition of a llm_as_judge metric and prompt template for table2text 
* addition of a structured data operator (name: ConstructTableFromRowsCols, Purpose: Builds a single table in standard format by mapping column and row fields into one structure)


prepare/cards/scigen.py
```diff
@@ -0,0 +1,59 @@
...
+        ConstructTableFromRowsCols(
+            fields=["table_column_names", "table_content_values"],
+            to_field="table",
+        ),
+        SerializeTableAsIndexedRowMajor(field_to_field=[["table", "input"]]),
+        Set({"type_of_input": "table"}),
+        Set({"type_of_output": "text"}),
+        RenameFields(field_to_field={"text": "output", "table_caption": "caption"}),
+    ],
+    task=Task(
```
elronbandel, line 28: @csrajmohan can you change the task here to this one:
https://github.com/IBM/unitxt/blob/dcbdb31348066ffccd2de4482902c105744ea72e/prepare/cards/numeric_nlg.py#L15-L28

This will allow users to explore in the interactive UI (works only with catalog tasks) and also I think a better way to define things

elronbandel (approved):
Thanks @csrajmohan ! can we change the default metric to be the default of the task? Then at any point you can specify it in the recipe `card=...metrics=[metrics.llm_as_judge.rating.llama_3_70b_instruct_ibm_genai_template_table2text_single_turn_with_reference]` Im concerned that closed api as a default metric will make it hard for users to use this task. what do you think @csrajmohan ?

csrajmohan:
Task 'tasks.generation.from_pair' has some automated metrics bleu, rouge etc. by default but for Scigen dataset they were not much useful so we needed llm as judge metric. We are using Llama3 via closed api as a judge metric for now. To make the card usable without such restriction, we may have to use some huggingface model like mistral-7b as judge later though it may not be as performant for our task.

elronbandel:
It's your decision. Now we just need to get it to pass the tests. I don't know why it fails, it seems like this is similar: https://github.com/IBM/unitxt/pull/1168 and there it passes the tests. Do you have any idea @csrajmohan what is the difference?

csrajmohan:
No idea @elronbandel. The error `genai.exceptions.ApiResponseException: Failed to handle request to https://bam-api.res.ibm.com/v2/text/generation/limits?version=2023-11-22.
{
  "error": "Unauthorized",
  "extensions": {
    "code": "AUTH_ERROR",
    "state": null,
    "reason": "TOKEN_INVALID"
  },
  "message": "Invalid or missing JWT token",
  "status_code": 401
}`

elronbandel:
@pawelknes Do you know why it might happened here? (and not on other PRs like https://github.com/IBM/unitxt/pull/1168)

pawelknes:
@elronbandel @csrajmohan It worked for me locally after pulling the branch. Maybe the `GENAI_KEY` we use in tests is either no longer valid, or have some limitations for particular models/tasks (I don't know exact details of levels of authorization with genai - if there are any). But on the other hand, other artifacts using genai were prepared successfully. I'll re-run the failed job to see if that was just some temporary issue

csrajmohan:
@pawelknes Any luck with resolving the token issue?

pawelknes:
> @pawelknes Any luck with resolving the token issue?

No clue to be honest. My only idea is that it is related somehow to the api key we use in tests as suggested in the error message (and the fact that I was able to run it locally with my own credentials). It'd be good to confirm that the used key is still valid and we have access to all the resources we need. @elronbandel what do you think?

Issue #1136, opened 2024-08-12
fixing the typo:   "def text_context_correctness(self):" -> "def test_context_correctness(self):"  blows the test


dafnapension:

matanor:
hi @dafnapension , what do you mean by "blows the test"?
is this issue only about fixing the typo?

dafnapension:
ahh... :-) once you change to test_ , the tests go into this method too (which they did not thus far, because of some settings that tells them to look for  test_ ..), and do not manage to complete it successfully, for some error that I do not understand..

matanor:
oh i see.. created a fix ..
thanks for identifying this issue! 

indeed with the test i realize now some change were introduced to the metric names.. (we may need to adjust out user code for that.. )

PR #1142, opened 2024-08-13
Add a test for sentence-bert based answer correctness metrics

matanor:


dafnapension:
Hi @matanor , I thought that `tetst_metrics.py` should be fixed. Is this PR in reference to issue https://github.com/IBM/unitxt/issues/1136 that I assigned to you (because your name shows in gitlens..)?

matanor:
> Hi @matanor , I thought that `tetst_metrics.py` should be fixed. Is this PR in reference to issue #1136 that I assigned to you (because your name shows in gitlens.. as the one who moved the tests from prepare to test_metrics)?

hi @dafnapension ! you mean we should first fix the above issue, before considering this PR ? (though the issue is on `context_correctness`, here its about new tests for `answer_correctness`)

dafnapension:
I must admit that I am less familiar with the exact inference power of the different models, I participate in unitxt more on the level of "pattern recognition" of code issues: if I see a problem (like [this issue](https://github.com/IBM/unitxt/issues/1136)), or asked to help with some specific pieces of code.
I only have my own Win10 machine (running on Dell), no access to any GPU.
So in these terms (of סדר וניקיון) I bumped into this [issue](https://github.com/IBM/unitxt/issues/1136), and looked in gitlens to see who can understand what is going on, and found your name in gitlens.
Now you explained in the PR (that made your name show in gitlens) that you see value in moving tests from prepare to `test_metrics`.
So now, in terms of סדר וניקיון, even before the correctness of the scores..  I am puzzled about two issues:
(1) why heavy tests again in prepare, and not in test_metrics,
(2) what about this "def text_context_correctness(self):"  (note the  text_...   which prevents this method from being tested at all) that I had the impression that is related to your current PR.

dafnapension:
Hi @matanor , this is how I manage to run your `prepare/metrics/rag_answer_correctness.py`  on my little Dell:

image

dafnapension (approved):

matanor:
@dafnapension same question as [here](https://github.com/IBM/unitxt/pull/1143#issuecomment-2286475422) .. is it ok to merge?
the error in `Test Catalog Preparation` seemd unrelated?

dafnapension:
Yes yes, from my perspective. I suspect, though that the 'merge' button is not applicable without passing all automatic tests.

PR #1143, opened 2024-08-13
Fix typo in method name, and the contained test

matanor:
- text_context_correctness -> test_context_correctness
- Fix the context_correctness metric names in the test, so it passes

Closes https://github.com/IBM/unitxt/issues/1136

### DIFF
--- a/tests/library/test_metrics.py
+++ b/tests/library/test_metrics.py
@@ -1770,7 +1770,7 @@ class TestConfidenceIntervals(UnitxtTestCase):
             global_target=global_target,
         )
 
-    def text_context_correctness(self):
+    def test_context_correctness(self):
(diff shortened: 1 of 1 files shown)


dafnapension (approved):

matanor:
@dafnapension is it ok to merge?
the failure in `Test Catalog Preparation` is unrelated, i think ..

elronbandel:
@matanor I fixed this in the main

dafnapension:
Hi @matanor , I merged this one into main, before something else breaks..
About the other two: please see if they take much time to run. If too much - then perhaps better to test in prepare, preferably, in the same file where the tested metric is added to the catalog. Better, I think, than in a totally separated file.

PR #1146, opened 2024-08-13
add a test for RAG faithfulness using sentence embebbers

matanor:


dafnapension:
Hi @matanor , why is this sitting in `prepare` and not in `test_metrics` ? You add no metric to the catalog, just test, isnt it?

dafnapension (approved):

PR #1147, opened 2024-08-13
Update operators.py --- copy edits (grammar, consistency, clarity)

welisheva22:
In addition to fixing typos for the word "specialized" and changed "dedicated for" to "dedicated to" --  I also adjusted something that seemed to render differently than planned --

The auto docs said "additional arguments are field names passed to the function" but it showed up on the web page as "(additional areguments are field names passed to the) - name of function" -- I changed it to "field names passed in" -- I hope this renders better.

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -14,9 +14,9 @@ To enhance the functionality of Unitxt, users are encouraged to develop custom o
 This can be achieved by inheriting from any of the existing operators listed below or from one of the fundamental :class:`base operators<unitxt.operator>`.
 The primary task in any operator development is to implement the `process` function, which defines the unique manipulations the operator will perform.
 
-General or Specelized Operators
+General or Specialized Operators
 --------------------------------
-Some operators are specielized in specific data or specific operations such as:
+Some operators are specialized in specific data or specific operations such as:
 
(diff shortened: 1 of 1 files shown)


src/unitxt/operators.py
```diff
@@ -777,7 +777,7 @@ class Apply(InstanceOperator):
     Args:
         function (str): name of function.
         to_field (str): the field to store the result
-        additional arguments are field names passed to the function
+        additional arguments are field names passed in
```
yoavkatz, line 780:
```suggestion
        any additional arguments are field names whose values will be passed directly to the specified function
```

welisheva22: Hi- thanks for that -- I think it makes sense but I think the code has an issue with using the word "function" at the end -- which is why at first the word function got out of the parentheses -- we can switch the word order to share the same meaning and not end the sentence with the word in. Let me see... I'll propose another change incorporating your words. I can understand if it didn't seem clear enough without them. Thanks!

welisheva22: p.s.: you can see the issue when you look at the rendered web page -- as opposed to looking at it in the python file

welisheva22:
Closed this and created another pull request for all of the changes, since this one seems to have a commit not signed off properly?

PR #1151, opened 2024-08-15
Dedicate nltk a mixin and download all versions of punkt

elronbandel:


yoavkatz (approved):

src/unitxt/metrics.py
```diff
@@ -2108,7 +2108,17 @@ class F1MacroMultiLabel(F1MultiLabel):
     average = None
 
 
-class Rouge(InstanceMetric):
+class NLTKMixin(Artifact):
+    def prepare(self):
+        super().prepare()
+        import nltk
```
yoavkatz, line 2114: make add as a _requirement_list?

elronbandel: There is not way to merge the requirements list of both classes

PR #1169, opened 2024-08-19
Update glossary.rst -- boldfaced and pre-capitalized terms

welisheva22:
Previously some terms were capitalized and some where not, now they are all boldfaced and pre-capped when the term is mentioned at the beginning of the definition.


welisheva22:
Closing, redoing this with proper signoff comment.

PR #1170, opened 2024-08-19
Update glossary.rst -- Boldface and Pre-Capitalize Terms

welisheva22:
Previously only some terms were boldfaced and there did not seem to be uniform capitalization. Now, all terms are boldfaced and pre-capped at the beginning of the definition.

Note: I did not change the wording -- some of the definitions did not use all exact words from the term, but I think that's OK. They all used at least one of the words from the term in the beginning of the definition, and it seems valuable to keep all of the words in the terms in case people are looking for different things.


welisheva22:
(Well, the wording was slightly tweaked in a previous pull request, so that was already all set.) :)

elronbandel (approved):

PR #1171, opened 2024-08-20
Add more LLMs as judges

pvn25:
Add judges for metrics to assess conversation quality on (1) Grounded: Every substantial claim in the response of the model is derivable from the content of the document. (2) IDK: Does the model response say I don't know?


elronbandel:
It seems like your example is crashing the tests for some reason so, just exclude it from the examples tests in the `tests/test_examples.py`

pvn25:
ok, I have fixed this. I am not sure how to resolve the DCO check now?

eladven (commented):
Regarding the DCO. You can simply click `details`  near the icon to open the check, and then click `Set DCO to pass`. Generally, it checks that all your commits has a sign off (the commit message contains something like "Signed-off-by: Elad Venezian <eladv@il.ibm.com>").

Regarding the PR itself: Looks great. I think you should also add these examples to our  examples table at examples.rst including explanation about what these examples shows which is different then current examples.

pvn25:
Thanks @eladven I just updated the examples.rst. Pls lmk if it is sufficient.

eladven (approved):

PR #1183, opened 2024-08-27
Support enumeration answer format in multiple choice template

antonpibm:
The current Multiple Choice Template supports 2 answer formats:
1. The full answer, which was converted to the index
2. The answer index, which must be passed as an int

This PR adds the ability to also provide the enumeration representation the answer, e.g. for 

> Which name is the longest?, Choices: A. Marry, B. Bob, C. Sam, D. Alfred.

Users can specify "D" as the correct answer

The default behavior is kept


elronbandel:
This can be solved with MapInstanceValues mapping `A` - > 0 and so on, in the preprocessing steps.

PR #1191, opened 2024-09-02
Add support for Literal, TypedDict and NewType for unitxt type checking

elronbandel:


src/unitxt/types.py
```diff
@@ -0,0 +1,34 @@
...
+    image: Any
+
+
+class Audio(TypedDict):
+    audio: Any
+
+
+class Table(TypedDict):
+    header: List[str]
+    rows: List[List[Any]]
```
eladven, line 26: I think that the basic (and common table is a text table, where rows: List[List[Text]]

src/unitxt/utils.py
```diff
@@ -87,6 +88,23 @@ def is_module_available(module_name):
         return False
 
 
+def remove_numeric_and_strings(input_str):
```
eladven, line 91:
```suggestion
def remove_numerics_and_quoted_texts(input_str):
```

eladven (approved):

PR #1192, opened 2024-09-02
Add rag dataset + openai format dialog operator

OfirArviv:


prepare/cards/chat_rag_bench.py
```diff
@@ -0,0 +1,64 @@
...
+    ]
+
+    test_card(
+        card_for_test,
+        strict=True,
+        demos_taken_from="test",
+    )
+    add_to_catalog(
+        card,
+        f"cards.rag.response_generation.chat_rag_bench.train.user_assistant_format.{subset}",
```
elronbandel, line 62:
```suggestion
        f"cards.rag.response_generation.chat_rag_bench.{subset}.user_assistant_format",
```
I think should be ordered by importance

OfirArviv:
> Unless there is another version I dont think that the `.train.` in the name is necessary. Unless there is a reason?

in normal cases we will want to run the normal split and just run the benchmark. It should be all test set. Now we have a smaller test set so we can keep some to train

prepare/cards/chat_rag_bench.py
OfirArviv: in your way it many directories, one for each subset. In my way there is no directory per subset, and all the subsets are in the same directory. Its easier to read.

elronbandel:
I see, then at least save the original by its name and then add the '.train' as additional option, what do you think?

OfirArviv:
> I see, then at least save the original by its name and then add the '.train' as additional option, what do you think?

good idea. Its free for us and we add another thing to the catalog :)

elronbandel (approved):

Issue #1106, opened 2024-08-04
More meaningful error message when catalog consistency fails


yoavkatz:
Today, the when the catalog consistency fails one gets this error message.
```
Comparing generated and old catalog...
***** Directories has differences ******
{'file': '/home/runner/work/unitxt/unitxt/src/unitxt/catalog/metrics/string_containment', 'diff': 'old only'}
  File "/home/runner/work/unitxt/unitxt/utils/prepare_all_artifacts.py", line 165, in <module>
    main()
  File "/home/runner/work/unitxt/unitxt/utils/prepare_all_artifacts.py", line 160, in main
    raise RuntimeError("Directories has differences")
RuntimeError: Directories has differences
```

It should return a more meaningful error such as.

```
Catalog consistency check failed.  The catalog json files were regenerated from the prepare/*/*.py files.  

* Some json files (or complete directories) which were committed in the past were not regenerated:

 home/runner/work/unitxt/unitxt/src/unitxt/catalog/metrics/string_containment
 
This can happen when renaming to catalog names of artifacts in  prepare/*/*.py files , but not deleting the old files.

* Some files which were generated were not committed.

 home/runner/work/unitxt/unitxt/src/unitxt/catalog/metrics/new_string_containment.json
 
 This can happen when not committing json files together with prepare/*/*.py.
 
 
 To fix recreate and fix this in your local environment:

 rm -fr  src/unitxt/catalog
 python utils/prepare_all_artifacts.py
 
This will cause all artifacts to be regenerated and you can use git status to see the changes and then commit them.

  ````

dafnapension:
Resolved with PR #1201

PR #1201, opened 2024-09-04
A suggested PR for issue 1106: More meaningful error message when catalog consistency fails

dafnapension:
Error messages are now informed and detailed, per issue https://github.com/IBM/unitxt/issues/1106, and on the way, catalog consistency check is made faster.


utils/prepare_all_artifacts.py
```diff
@@ -28,137 +28,158 @@ def import_module_from_file(file_path):
...
+# flake8: noqa: C901
+def main():
+    catalog_dir = constants.catalog_dir
+    catalog_back_dir = catalog_dir + "_back"
 
-def prepare_all_catalog_artifacts(catalog_dir):
     os.environ["UNITXT_USE_ONLY_LOCAL_CATALOGS"] = "True"
     os.environ["UNITXT_TEST_CARD_DISABLE"] = "True"
     os.environ["UNITXT_TEST_METRIC_DISABLE"] = "True"
     os.environ["UNITXT_SKIP_ARTIFACTS_PREPARE_AND_VERIFY"] = "True"
```
dafnapension, line 39: Is `os.environ["UNITXT_SKIP_ARTIFACTS_PREPARE_AND_VERIFY"]` used anywhere in unitxt?

elronbandel (commented):
Initially the idea of doing it in rounds was that there are dependencies between catalog artifacts and that way we can make sure each artifact has its dependencies. if not in the first round then in the second... this does not take a lot of time since they fail immediately if they dont have their dependencies.

dafnapension:
To be fanatically precise..  each prepare_file should complete its run successfully (excluding tests). For that success, among others, the prepare_file should find all its dependencies in the catalog as is (the branch's catalog), not the catalog to be generated.  
The second condition for success is identical sets: the branch's catalog, and the generated catalog.
Together, both conditions, if met, imply the same. But in case of error, the report should be different and more accurate, in the spirit of the Issue referred.

PR #1203, opened 2024-09-06
Add TogetherAiInferenceEngine

yifanmai:
Add `TogetherAiInferenceEngine`, which uses the [Together AI API](https://docs.together.ai/) for inference.

### DIFF
--- a/src/unitxt/inference.py
+++ b/src/unitxt/inference.py
@@ -347,6 +347,96 @@ class OpenAiInferenceEngine(
         return outputs
 
 
+class TogetherAiInferenceEngineParamsMixin(Artifact):
+    max_tokens: Optional[int] = None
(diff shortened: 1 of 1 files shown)


elronbandel (approved):

src/unitxt/inference.py
```diff
@@ -347,6 +347,96 @@ def _infer_log_probs(self, dataset):
...
+):
+    label: str = "together"
+    model_name: str
+    _requirements_list = {
+        "together": "Install together package using 'pip install --upgrade together"
+    }
+    data_classification_policy = ["public"]
+    parameters: Optional[TogetherAiInferenceEngineParamsMixin] = None
+
+    def prepare(self):
```
elronbandel, line 376:
```suggestion
    def prepare_engine(self):
```

PR #1205, opened 2024-09-08
Integrate metrics team LLMaJ with current unitxt implemantation

lilacheden:
Integrating the metrics-llmaj pipeline with the current unitxt llmaj. 
This required some changes outside the scope of our new catalog:

1. moving code from fm-eval to unitxt, including changes from @arielge's  old pull request with the log probs inference/processors (https://github.com/IBM/unitxt/pull/1111), our tasks and templates and some more _infer_log_probs support. 

2. Changing the LLMAsJudge class to allow different processing of the input. Specifically to let the template get the different dataset fields (answer, contexts etc) rather than the full template+response of the previous model.
Ideally we'd have a LLMAsJudge parent class and deriving classes, but to make it backward compatible I kept the LLMAsJudge api as before and worked around it for now by adding a class `LLMAsJudgeTaskFormatter` as an attribute. Hope we can agree on a better design together.


src/unitxt/inference.py
```diff
@@ -250,6 +253,47 @@ def _infer(self, dataset):
...
         ]
 
+    def _infer_log_probs(self, dataset):
+        from genai.schema import TextGenerationParameters
+
+        logprobs_return_options = {
+            "generated_tokens": True,
+            "input_text": False,
+            "input_tokens": False,
+            "token_logprobs": True,
```
OfirArviv, line 263: @elronbandel  we should probably have this argument configurable in the abstract class somehow

src/unitxt/inference.py
```diff
@@ -579,10 +626,14 @@ def process_data_before_dump(self, data):
...
     @staticmethod
     def _read_wml_credentials_from_env() -> (
-        Dict[Literal["url", "apikey", "project_id"], str]
+        Dict[Literal["url", "apikey", "project_id", "space_id"], str]
     ):
         credentials = {}
-        for env_var_name in ["WML_URL", "WML_PROJECT_ID", "WML_APIKEY"]:
+        project_or_deployment_var_name = (
+            "WML_SPACE_ID" if "WML_SPACE_ID" in os.environ else "WML_PROJECT_ID"
+        )
```
OfirArviv, line 634: what is the difference between the different env vars?

src/unitxt/api.py
```diff
@@ -138,11 +139,18 @@ def infer(
     engine,
     dataset_query: Optional[str] = None,
     return_data=False,
+    infer_log_probs=False,
```
OfirArviv, line 142: return_log_probs

src/unitxt/api.py
```diff
@@ -138,11 +139,18 @@ def infer(
...
     **kwargs,
 ):
     dataset = produce(instance_or_instances, dataset_query, **kwargs)
     engine, _ = fetch_artifact(engine)
-    raw_predictions = engine.infer(dataset)
+    if infer_log_probs:
+        raw_predictions = engine.infer_log_probs(dataset)
```
OfirArviv, line 148: no one guarantee you the engine implemented the interface. You need to check it

src/unitxt/inference.py
lilacheden: The wml credentials can be associated with a project (id set in env var WML_PROJECT_ID) or a space (id set in env var WML_SPACE_ID), we were asked to support both options.

src/unitxt/api.py
lilacheden: done

src/unitxt/api.py
lilacheden: done

src/unitxt/inference.py
```diff
@@ -250,6 +253,47 @@ def _infer(self, dataset):
             )
         ]
 
+    def _infer_log_probs(self, dataset):
```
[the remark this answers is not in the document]
lilacheden: I can suggest to change both _infer and _infer_log_probs output to contain both predictions and metadata, as it is required by users. so each dict will contain:
```
final_results = {
                "prediction": predict_result, # text for infer, logprobs dict for infer_log_probs
                "input_tokens": result.input_token_count,
                "output_tokens": result.generated_token_count,
                "model_id": self.model_name,
            }
```

src/unitxt/llm_as_judge.py
```diff
@@ -24,27 +32,105 @@ class LLMAsJudge(BulkInstanceMetric):
...
+                f"Provided format argument to 'LLMAsJudge' metric is not of type Format, but {type(self.format)}"
+            )
+
+        if self.system_prompt and not isinstance(self.system_prompt, SystemPrompt):
+            raise ValueError(
+                f"Provided system_prompt argument to 'LLMAsJudge' metric is not of type SystemPrompt, but {type(self.system_prompt)}"
+            )
+
+        if isinstance(self.inference_model, OpenAiInferenceEngine):
+            if self.format and type(self.format) is not SystemFormat:
```
OfirArviv, line 68: why the second condition? The format should just be None, no?

OfirArviv: or did you mean EmptyFormat?

src/unitxt/llm_as_judge.py
```diff
@@ -24,27 +32,105 @@ class LLMAsJudge(BulkInstanceMetric):
...
     template: Template
     system_prompt: SystemPrompt = Field(default_factory=EmptySystemPrompt)
     format: Format = Field(default_factory=SystemFormat)
-    strip_system_prompt_and_format_from_inputs: bool = True
     inference_model: InferenceEngine
     reduction_map: Optional[Dict[str, List[str]]] = None
     batch_size: int = 32
     prediction_type = Any  # Because handled with multiple tasks
+    infer_log_probs = False
+    include_meta_data: bool = False
```
eladven, line 50: Why does the base class should be aware of these implementation parameters

src/unitxt/llm_as_judge.py
```diff
@@ -24,27 +32,105 @@ class LLMAsJudge(BulkInstanceMetric):
...
+            system_prompt=self.system_prompt,
+            format=self.format,
+            return_data=True,
+            return_log_probs=self.infer_log_probs,
+            return_meta_data=self.include_meta_data,
+        )
+
+        return self.get_results_from_outputs(outputs)
+
+    def prepare_instances(self, references, predictions, task_data):
```
OfirArviv, line 107: should the class be an Abstract class?

src/unitxt/llm_as_judge.py
```diff
@@ -24,27 +32,105 @@ class LLMAsJudge(BulkInstanceMetric):
...
         batch_size (int): The size of the bulk.
     """
 
     main_score: str = "llm_as_judge"
-    task: Literal[
-        "rating.single_turn",
-        "rating.single_turn_with_reference",
-        "pairwise_comparative_rating.single_turn",
-    ]
+    task: str
```
OfirArviv, line 41: Shouldn't the be a TASK type as well?

src/unitxt/llm_as_judge.py
```diff
@@ -209,5 +242,73 @@ def compute(
...
+    mapping: Dict[str, str] = {}
+    prediction_field: Optional[str] = None
+    include_meta_data: bool = True
+
+    # Allow for input which is a dictionary of all input fields. In this case, all input fields are
+    # treated as the task data, and the predictions and references are taken directly from there
+    # by the judge's template
+    def preprocess_instance(self, instance):
+        if "task_data" not in instance:
+            instance["task_data"] = instance.copy()
```
OfirArviv, line 265: Add a comment on why you are doing this so future readers will understand

src/unitxt/llm_as_judge.py
```diff
@@ -209,5 +242,73 @@ def compute(
...
+        return results
+
+    def prepare_instances(self, references, predictions, task_data):
+        input_instances = self._get_input_instances(task_data)
+        return self._get_instance_for_judge_model(
+            input_instances, predictions, references
+        )
+
+
+class TaskBasedLLMasJudge(LLMAsJudgeBase):
```
OfirArviv, line 254: why not have this class as the base class?

OfirArviv (changes requested):

src/unitxt/llm_as_judge.py
OfirArviv: and the original llm as judge should be implemented using this class as base

src/unitxt/llm_as_judge.py
lilacheden: removed these from base class

src/unitxt/llm_as_judge.py
lilacheden: not sure, this verification part is taken as is from LLMasJudge in main

OfirArviv: I think it is an error then

src/unitxt/llm_as_judge.py
lilacheden: @OfirArviv - let's discuss offline

src/unitxt/inference.py
```diff
@@ -47,6 +47,9 @@ class TextGenerationInferenceOutput:
 class InferenceEngine(abc.ABC, Artifact):
     """Abstract base class for inference."""
 
+    label: str
```
OfirArviv, line 50: I'm not sure we want to face having a model_name. Having an optional label is fine, but make it optional. What is the need for it?

src/unitxt/inference.py
```diff
@@ -80,16 +83,29 @@ def infer(
...
+                f"Inference engine {self.__class__.__name__} does not support return_meta_data,"
+                f"Please set return_meta_data=False"
+            )
 
         [self.verify_instance(instance) for instance in dataset]
         if settings.mock_inference_mode:
             return [instance["source"] for instance in dataset]
         return self._infer(dataset, return_meta_data)
 
+    def get_engine_id(self):
```
OfirArviv, line 97: what is the need for this?

src/unitxt/inference.py
```diff
@@ -137,10 +153,11 @@ def infer_log_probs(
...
         return_meta_data is only supported for some InferenceEngines.
         """
-        assert return_meta_data is False or hasattr(self, "get_return_object"), (
-            f"Inference engin {self.__class__.__name__} does not support return_meta_data,"
-            f"Please set return_meta_data=False"
-        )
+        if return_meta_data and not hasattr(self, "get_return_object"):
+            raise NotImplementedError(
+                f"Inference engine {self.__class__.__name__} does not support return_meta_data,"
+                f"Please set return_meta_data=False"
```
OfirArviv, line 159: please also add the function name

src/unitxt/inference.py
```diff
@@ -80,16 +83,29 @@ def infer(
         If return_meta_data - returns a list of TextGenerationInferenceOutput, else returns a list of the string
         predictions.
         """
-        assert return_meta_data is False or hasattr(self, "get_return_object"), (
-            f"Inference engin {self.__class__.__name__} does not support return_meta_data,"
-            f"Please set return_meta_data=False"
-        )
+        if return_meta_data and not hasattr(self, "get_return_object"):
+            raise NotImplementedError(
```
OfirArviv, line 87: please also add the function name

src/unitxt/api.py
```diff
@@ -142,16 +144,42 @@ def infer(
     engine,
     dataset_query: Optional[str] = None,
     return_data=False,
+    return_log_probs=False,
```
OfirArviv, line 147: add typing please to the args

src/unitxt/llm_as_judge.py
```diff
@@ -14,38 +14,119 @@
...
+                raise ValueError(
+                    "Error in 'LLMAsJudge' metric. Inference model 'OpenAiInferenceEngine' does "
+                    "not support system prompt. Please remove the system_prompt definition from the recipe"
+                    " (Current implementation of Unitxt does not support this."
+                    " Support will be added in future updates)."
+                )
+
+    def get_full_task_name(self):
+        raise NotImplementedError
+        # f"tasks.response_assessment.{self.task}",
```
OfirArviv, line 82: remove

src/unitxt/llm_as_judge.py
```diff
@@ -14,38 +14,119 @@
...
+        task_data: List[Dict],
+    ) -> List[Dict[str, Any]]:
+        instances = self.prepare_instances(references, predictions, task_data)
+        outputs = self.infer_instances(instances)
+        return self.get_results_from_outputs(outputs)
+
+    def infer_instances(self, instances):
+        raise NotImplementedError
+
+    def prepare_instances(self, references, predictions, task_data):
```
OfirArviv, line 97: add a docstirng explaining the purpose of this method

src/unitxt/llm_as_judge.py
```diff
@@ -14,38 +14,119 @@
...
+
+    def compute(
+        self,
+        references: List[List[Any]],
+        predictions: List[Any],
+        task_data: List[Dict],
+    ) -> List[Dict[str, Any]]:
+        instances = self.prepare_instances(references, predictions, task_data)
+        outputs = self.infer_instances(instances)
+        return self.get_results_from_outputs(outputs)
```
OfirArviv, line 92: This method is not declared in this class as far as I can see. Add it with explanation of what it should do

OfirArviv (changes requested):

src/unitxt/llm_as_judge.py
OfirArviv: @elronbandel

src/unitxt/llm_as_judge.py
```diff
@@ -209,5 +267,141 @@ def compute(
...
+        post-processing must support the logprobs output.
+        judge_to_generator_fields_mapping (Dict[str, str]): optional mapping between the names of the fields in the generator task and the
+        judge task. For example, if the generator task uses "reference_answers" and the judge task  expect "ground_truth",
+        include  {"ground_truth": "reference_answers"} in this dictionary.
+        prediction_field: if indicated, and prediction exist, copy prediction to this field name in task_data.
+        include_meta_data (bool): whether to include the inference per-instance metadata in the returned results.
+
+    """
+
+    infer_log_probs: bool = True
```
OfirArviv, line 309: this probably should be false. The common way to use it will be with plain text

prepare/metrics/llm_as_judge/binary_judge.py
```diff
@@ -0,0 +1,46 @@
...
+    return None if metric_type == "context_relevance" else "answer"
+
+
+for metric_type, template_dict in metric_type_to_template_dict.items():
+    for template_short_name, template_name in template_dict.items():
+        task_name = f"tasks.rag_eval.{metric_type}.binary"
+
+        for use_logprobs in [True, False]:
+            logprobs_label = "_logprobs" if use_logprobs else ""
+            metric_label = f"{metric_type}_{template_short_name}{logprobs_label}"
```
OfirArviv, line 31: also add the model and the "wml"

prepare/engines/classification/classification_engines.py
```diff
@@ -0,0 +1,56 @@
...
+)
+from unitxt.random_utils import get_seed
+
+
+def get_inference_engine(model_name, framework_name):
+    if framework_name == "ibm_wml":
+        return WMLInferenceEngine(
+            model_name=model_name,
+            max_new_tokens=5,
+            random_seed=get_seed(),
```
OfirArviv, line 15: this will not do what I think you think it will do. The see will be fixed to the one that was set when the json was created.

prepare/engines/classification/classification_engines.py
```diff
@@ -0,0 +1,56 @@
...
+for judge_model_name, infer_frameworks in model_names_to_infer_framework.items():
+    for infer_framework in infer_frameworks:
+        template_format = (
+            "formats.llama3_instruct"
+            if "llama" in judge_model_name
+            else "formats.empty"
+        )
+
+        inference_engine = get_inference_engine(judge_model_name, infer_framework)
+        inference_engine_label = inference_engine.get_engine_id()
```
OfirArviv, line 50: If that what you wanted to use the "get_engine_id" the its no good. The engine name in the catalog should be descriptive as possible. This include the actual engine ("wml", "bam", "hf", etc.) and any specific args. In this case max_new_tokens=5, is not a standard setting, and maybe also the greedy args

prepare/templates/rag_eval.py
```diff
@@ -0,0 +1,299 @@
+from unitxt import add_to_catalog
+from unitxt.templates import InputOutputTemplateWithCustomTarget, TemplatesList
+
+
+def add_rag_templates(
+    templates_dict, task, out_field="textual_label", reference_field="number_val"
+):
```
OfirArviv, line 7: maybe we should split this file? It's unreadable imo and should be used by other users as well ideally @elronbandel wdyt?

prepare/tasks/rag_eval.py
```diff
@@ -0,0 +1,107 @@
+from typing import List, Union
```
OfirArviv, line 1: again, unreadable imo. And the catalog should be universal. @elronbandel wdyt?

src/unitxt/llm_as_judge.py
lilacheden: There is no EmptyFormat, not sure what to change it to

src/unitxt/llm_as_judge.py
lilacheden: it's a type for the original llm as judge, for the new one the task can vary

src/unitxt/llm_as_judge.py
lilacheden: ok

prepare/engines/classification/classification_engines.py
lilacheden: it sets the seed to 42, which is fine by me - I just want it to be deterministic

prepare/engines/classification/classification_engines.py
lilacheden: This is why these engines are included under the "classification" folder in the catalog - they are engines with specific params designed for classification. Do you want me to add "classification_" to the label of these engines (e.g. "classification_wml") in the catalog?

OfirArviv: wml should be part of the engine id.
About the max_new_tokens, if its under classification, it ok

prepare/engines/classification/classification_engines.py
OfirArviv: so set it manually. The json are being prepared each pr as part of the test. In it ever change it will cause problems

prepare/tasks/rag_eval.py
OfirArviv: did you open an issue for it?

prepare/templates/rag_eval.py
OfirArviv: did you open an issue for it?

src/unitxt/llm_as_judge.py
OfirArviv: did you open an issue for it?

OfirArviv (approved):

src/unitxt/llm_as_judge.py
lilacheden: yes

PR #1209, opened 2024-09-09
add unified decorator for warnings and unit tests

MikolajCharchut:
Added decorator with warning and message for classes. It raises specific warning when object of a class is initialised. Additionally I created tests for that decorator.


elronbandel (changes requested):
this is great thanks! can you blend in the logic from those:

```python
def decorator(obj):
        alt_text = f" Use {alternative} instead." if alternative is not None else ""
        alt_text += msg if msg is not None else ""
        if callable(obj):
            func = obj
        elif hasattr(obj, "__init__"):
            func = obj.__init__
        else:
            raise ValueError("Unsupported object type for deprecation.")
        return depraction_wrapper(func, version, alt_text)
    return decorator
def init_warning(msg=""):
    # Decorator that raises warning when class is initialized
    def decorator(initiated_class):
        UnitxtWarning(msg)
         return initiated_class

     return decorator
 ```
 So we have only one warnings decorator for functions and classes that uses UnitxtWarning and can work for all the cases above. Also have one test that check warning is not raised by class declaration only by class instantiation.

File: tests/library/test_unified_warnings_decorator.py [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+            def __init__(self, name):
+                self.name = name
+
+        with self.assertWarns(UserWarning) as warning_context:
+            obj = TestClass("Initialized_object")
+
+        self.assertEqual(str(warning_context.warning), "Class object initialized!")
+        self.assertEqual(obj.name, "Initialized_object")
+
+    def test_warning_called_on_instance_creation(self):
```
elronbandel, line 19: no sure what is the difference between the tests?

MikolajCharchut:
@elronbandel I blended the logic to make decorator compatible with both classes and functions, but should I leave the `warning_type` argument or delete it and just use `UnitxtWarning(msg)` in both cases?

PR #1218, opened 2024-09-17
Changing v11 instruction to test LLMAJ grounded ensemble

KMirshahi:


elronbandel:
@KMirshahi what is the status of this? To me it looks like an important separation between instruction and the few shot examples. But not sure who's using it and how. @OfirArviv @pvn25 do you have a comment to add here?

PR #1222, opened 2024-09-18
Add random templates for bluebench

perlitz:


elronbandel (changes requested):
I think we should call them `*.blue_bench` the current name is too long and might change if you add more templates to the lists. I think blue_bench will capture their purpose and make it clear what this list is used for.

perlitz:
> I think we should call them `*.blue_bench` the current name is too long and might change if you add more templates to the lists. I think blue_bench will capture their purpose and make it clear what this list is used for.

Sure, I fixed it

elronbandel (approved):

PR #1227, opened 2024-09-20
Add wml based llmaj metric

perlitz:


prepare/metrics/llm_as_judge/pairwise_rating/llama_3_ibm_genai_arena_hard_template.py
```diff
@@ -1,6 +1,8 @@
 from unitxt import add_to_catalog
```
OfirArviv, line 1: change file name so its not only ibm_genai

prepare/metrics/llm_as_judge/pairwise_rating/llama_3_ibm_genai_arena_hard_template.py
```diff
@@ -10,27 +12,49 @@
...
+            ):
+                continue  # currently not supported
+
+            # if inference engine is generic, these configurations will be defined when it is saved to the catalog
+            if inference_engine_name != "generic_engine":
+                inference_model = inference_engine(
+                    model_name=model_id, max_new_tokens=2048, random_seed=42
+                )
+            else:
+                inference_model = inference_engine()
```
OfirArviv, line 40: doesn't it need a default?

examples/run_generic_inference_engine.py
```diff
@@ -2,19 +2,34 @@
 from unitxt.inference import GenericInferenceEngine
 
 if __name__ == "__main__":
```
OfirArviv, line 4: if its an example you need to explain in text what you are doing here

examples/run_generic_inference_engine.py
```diff
@@ -2,19 +2,34 @@
...
+        {
+            "question": "If a ate an apple in the morning, and one in the evening, what is the number of apples I have eaten?, answer only with numerals",
             "answers": ["2"],
         },
     ]
     dataset = produce(instances, recipe)
 
-    predictions = generic_engine.infer(dataset)
-
+    # now, trying without a default, make sure you have something like
```
OfirArviv, line 21: in line 24 it with default

OfirArviv (changes requested):

prepare/metrics/llm_as_judge/pairwise_rating/llama_3_ibm_genai_arena_hard_template.py
eladven: I agree with Ofir. I think that we should add a default (I think that our main use is IBM genai), which will enable us to use mainly the generic metric. The internal users will use it the same way they used the metric today, and the external users will change the actual engine with the env var

prepare/metrics/llm_as_judge/pairwise_rating/llama_3_ibm_genai_arena_hard_template.py
perlitz: done

prepare/metrics/llm_as_judge/pairwise_rating/llama_3_ibm_genai_arena_hard_template.py
perlitz: added

examples/run_generic_inference_engine.py
perlitz: yes

examples/run_generic_inference_engine.py
perlitz: added

src/unitxt/inference.py
```diff
@@ -212,8 +212,8 @@ class GenericInferenceEngine(InferenceEngine):
     default: str = None
 
     def prepare_engine(self):
-        if "inference_engine" in os.environ:
-            engine_reference = os.environ["inference_engine"]
+        if "UNITXT_INFERENCE_ENGINE" in os.environ:
+            engine_reference = os.environ["UNITXT_INFERENCE_ENGINE"]
         else:
```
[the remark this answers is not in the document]
perlitz: fixed

examples/run_generic_inference_engine.py
```diff
@@ -1,20 +1,52 @@
...
+    # - template: Selects the specific template within the card (from templates.qa.open.simple).
+    # - demos_pool_size and num_demos: Control the number of demonstration examples used (set to 0 here).
     recipe = "card=cards.almost_evil,template=templates.qa.open.simple,demos_pool_size=0,num_demos=0"
+
+    # Create a list of instances (data points) for inference.
+    # Each instance has a "question" and its corresponding "answers".
     instances = [
-        {"question": "How many days there are in a week", "answers": ["7"]},
         {
-            "question": "If a ate an apple in the morning, and one in the evening, how many apples did I eat?",
```
OfirArviv, line 12: there are 2 questions here

Issue #1215, opened 2024-09-12
WMLInferenceEngine concurrency_limit


allyssonf:
`WMLInferenceEngine` under the hood uses `ibm_watsonx_ai` _generate_text_ function that implements concurrent requests...
What is happening is that `concurrency_limit` parameter is enabled and set to 10 leading to:

```json
{
    "errors": [
        {
            "code": "rate_limit_reached_requests",
            "message": "Rate limit of 8 requests per 1s was reached for instance id <INSTANCE_ID> (user <USER_ID>, plan v2-standard)",
            "more_info": "https://cloud.ibm.com/apidocs/watsonx-ai"
        }
    ],
    "trace": "1a29827ea1b6c65ae7f7376487c3c7fe",
    "status_code": 429
}
```

The current implementation of UNITXT does not allow us to change this value so it would be great if the user could set its value in `WMLInferenceEngine` implementation.

elronbandel:
Would you like to set it in the WMLInferenceEngine constructor?

allyssonf:
> Would you like to set it in the WMLInferenceEngine constructor?

@elronbandel , here is the code for `_infer` implementation from `unitxt/src/unitxt/inference.py:615`

```python
    def _infer(self, dataset):
        from ibm_watsonx_ai.foundation_models import ModelInference

        model = ModelInference(
            model_id=self.model_name,
            deployment_id=self.deployment_id,
            api_client=self._client,
        )

        return model.generate_text(
            prompt=dataset["source"],
            params=self.to_dict([WMLInferenceEngineParamsMixin], keep_empty=False),
            concurrency_limit=8 # added this line to avoid 429
        )
```

I would say that adding the parameter to `infer` from WMLInferenceEngine would be an option but thinking twice your suggestion seems better...

Thanks!

PR #1231, opened 2024-09-24
Add concurency_limit parameter to WMLInferenceEngine

elronbandel:
Closes: https://github.com/IBM/unitxt/issues/1215

### DIFF
--- a/src/unitxt/inference.py
+++ b/src/unitxt/inference.py
@@ -559,6 +559,7 @@ class WMLInferenceEngine(
         parameters (WMLInferenceEngineParams, optional): Instance of WMLInferenceEngineParams
             which defines inference parameters and their values. Deprecated attribute, please
             pass respective parameters directly to the WMLInferenceEngine class instead.
+        concurrency_limit (int): number of requests that will be sent in parallel, max is 10.
 
     Examples:
         from .api import load_dataset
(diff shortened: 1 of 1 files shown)


allyssonf:
@elronbandel , looks good to me. But is it possible to update the documentation adding this new parameter?

```python
class WMLInferenceEngine(
    InferenceEngine, WMLInferenceEngineParamsMixin, PackageRequirementsMixin
):
    """Runs inference using ibm-watsonx-ai.

    Attributes:
        credentials (Dict[str, str], optional): By default, it is created by a class
            instance which tries to retrieve proper environment variables
            ("WML_URL", "WML_PROJECT_ID", "WML_APIKEY"). However, a dictionary with
            the following keys: "url", "apikey", "project_id" can be directly provided
            instead.
        model_name (str, optional): ID of a model to be used for inference. Mutually
            exclusive with 'deployment_id'.
        deployment_id (str, optional): Deployment ID of a tuned model to be used for
            inference. Mutually exclusive with 'model_name'.
        parameters (WMLInferenceEngineParams, optional): Instance of WMLInferenceEngineParams
            which defines inference parameters and their values. Deprecated attribute, please
            pass respective parameters directly to the WMLInferenceEngine class instead.
       concurrency_limit (int): number of requests that will be sent in parallel, max is 10. 
```

pawelknes (approved):
lgtm

elronbandel:
Done @allyssonf.

PR #1233, opened 2024-09-25
Added calls to prepare() to ensure requirement checks are done early

yoavkatz:


yoavkatz:
This PR tries to deal with the fact that dependency checks were not performed early enough , and when running examples with required installed packages returned  "module was not found" errors.

However, there seems to be a problem with installing ibm_watsonx_ai  required for building inference engines.

elronbandel (approved):

yoavkatz:
@elronbandel - the problem is that watsonx_ai requires python 3.10.  What do you suggest we do?

PR #1236, opened 2024-09-25
adding vllm remote inference engine

assaftibm:


src/unitxt/inference.py
```diff
@@ -317,18 +317,26 @@ class OpenAiInferenceEngine(
...
-        api_key_env_var_name = "OPENAI_API_KEY"
-        api_key = os.environ.get(api_key_env_var_name)
+    @classmethod
+    def get_api_param(cls, inference_engine: str, api_param_env_var_name: str):
+        api_key = os.environ.get(api_param_env_var_name)
         assert api_key is not None, (
-            f"Error while trying to run OpenAiInferenceEngine."
-            f" Please set the environment param '{api_key_env_var_name}'."
+            f"Error while trying to run {inference_engine}."
+            f" Please set the environment param '{api_param_env_var_name}'."
```
lilacheden, line 325: better use `self.__class__.__name__` instead of the `inference_engine` param

assaftibm: sorry, didn't see your comment on time and I see the PR is already merged. 
Will your suggestion work when the code is in the base class and we try to get the name of the derived class?

PR #1239, opened 2024-09-26
Deep copy artifacts that fetched twice

elronbandel:


elronbandel:
Depend on: https://github.com/IBM/unitxt/pull/1240

PR #1241, opened 2024-09-29
Adding of ANLS metric to doc_vqa and info_vqa datasets

alfassy:
Adding of ANLS metric to doc_vqa and info_vqa datasets


elronbandel (approved):
Great job @alfassy !

PR #1243, opened 2024-09-29
Enhancements to inference engines

lilacheden:
1. add wml and gen_ai infer_log_probs
2. allow wml to receive space_id instead of project_id
3. add VLLMRemoteInferenceEngine for remote openai host
4. add InferenceEngineReturn object that returns predictions along metadata, set by param return_meta_data fo infer


elronbandel (changes requested):
Can we leave the InferenceEngineReturn out at this point? We have open PRs with different engines (image, retrieval) that do not conform with this output layout. Alternatively, you could name it `TextGenerationInferenceOutput`.

Given that you want to change the name, it's better to use python's `dataclasses.dataclass` which make the use of the constructor (__init__) unnecessary. Also notice the code for the class appear twice right now.

elronbandel (approved):

PR #1249, opened 2024-09-30
Fix duplicate column deletion bug in pandas serializer

elronbandel:
If column appear twice it just delete it, now it will give it a suffix _1, _2 etc


csrajmohan (approved):
looks good

PR #1256, opened 2024-10-06
Fix answer correctness confidence intervals

matanor:
- While the `main_score` in the `metrics.rag.answer_correctness.token_recall` metric was `recall`, the `score_name` field was `f1` and the confidence interval values were taken from F1. 

See for example, the score value is 0.29 (of recall), but the `score_ci_low`, `score_ci_high` and `score_name` fields refer to F1.
image

This PR:
1. Sets the `score_name` field for instance and global levels to the `main_score` defined for each MetricPipeline.
2. Copies the confidence interval values of the `main_score` (e.g. `recall_ci_low` and `recall_ci_high`) to the `score_ci_low` and `score_ci_high` fields.
3. Also moved the test which involves only recall calculation (i.e. no models involved) to the module top-level, so it runs in CI.
4. Fixed the catalog names of the RAG answer correctness metrics in the module tests.

@elronbandel @yoavkatz 
- Seems to me that steps 1 and 2 are general for every MetricPipeline. WDYT?
- It would have been nice to copy the confidence interval fields only if they exists. Is there a `CopyIfExists` opetator?


yoavkatz:
Hi. I think it's a good point - although it may be a fix which is not backward compatible.  
We we make the change , it should be in 

self.prepare_score = Copy(
            field_to_field=[
                [
                    f"score/instance/{self.metric._add_score_prefix(self.main_score)}",
                    "score/instance/score",
                ],
                [
                    f"score/global/{self.metric._add_score_prefix(self.main_score)}",
                    "score/global/score",
                ],
            ],
        )
        
@elronbandel - What do you think?

elronbandel (approved):

PR #1267, opened 2024-10-08
Fix missing deep copy in MapInstanceValues

yoavkatz:
If the value we map to is a list, we must deepcopy the value, otherwise external  changes to the list effect the value of the mapper .

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -207,8 +207,10 @@ class MapInstanceValues(InstanceOperator):
 
     def get_mapped_value(self, instance, key, mapper, val):
         val_as_str = str(val)  # make sure the value is a string
-        if val_as_str in mapper:
-            return mapper[val_as_str]
+        if (
+            val_as_str in mapper
+        ):  # must deep copy value to avoide external modification to the mapped value (e.g. if it is a list)
+            return copy.deepcopy(mapper[val_as_str])
         if self.strict:
             raise KeyError(
                 f"value '{val}' in instance '{instance}' is not found in mapper '{mapper}', associated with field '{key}'."
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -136,8 +136,8 @@ class MapInstanceValues(InstanceOperator):
     it maps values of instances in a stream using predefined mappers.
 
     Attributes:
-        mappers (Dict[str, Dict[str, str]]): The mappers to use for mapping instance values.
-            Keys are the names of the fields to be mapped, and values are dictionaries
+        mappers (Dict[str, Dict[str, Any]]): The mappers to use for mapping instance values.
+            Keys are the names of the fields to undergo mapping, and values are dictionaries
             that define the mapping from old values to new values.
         strict (bool): If True, the mapping is applied strictly. That means if a value
             does not exist in the mapper, it will raise a KeyError. If False, values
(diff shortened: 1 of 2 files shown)


elronbandel:
I think all the copies in unitxt should be one of these copy functions in utils.py
Where every function add more power with additional price:

- shallow_copy
- recursive_copy
- recursive_shallow_copy
- recursive_deep_copy
- deep_copy

In this case I think that recursive shallow copy is enough. what it does is to recursively recreate every list, dict, tuple, with shallow copy of the object inside. 

In this case of list of images for example it will recreate a new list of images where every image is with new reference, yet the pixels are shared between images.

dafnapension:
Hi, First, just a reminder, that the docstring of MapInstanceValue states:
```
    Attributes:
        mappers (Dict[str, **Dict[str, str]**]): The mappers to use for mapping instance values.
            Keys are the names of the fields to be mapped, and values are dictionaries
            that define the mapping from old values to new values.
```

src/unitxt/operators.py
```diff
@@ -207,7 +215,12 @@ def process(
 
     def get_mapped_value(self, instance, key, mapper, val):
         val_as_str = str(val)  # make sure the value is a string
-        if val_as_str in mapper:
+        if (
+            val_as_str in mapper
+        ):  # carefully check what to return, to avoid external modification to the mapped value (e.g. if it is a list)
+            do_cast, cast_what = self._cast_returned_value[key][val_as_str]
+            if do_cast:
+                return ast.literal_eval(cast_what)
```
elronbandel, line 223: what of the purpose of this stringfication and then de-serialization?

dafnapension: Only of values (to return from the mapping) that are not strings. 
An easy way to generate a new (again: not string) object, to return from the mapping.
On verification time, all the non-string values to return are identified and marked.
Then, on run time, a tiny overhead of looking up that mark is incurred, and then **only the non-strings** go through  `ast.literal_eval`  .  The original string representation of the value is maintained intact at the operator.

elronbandel: So is this a way to achieve deep copy for simple python strucres?

yoavkatz: @dafna @elronbandel - The original deep copy is a reasonable solution for string.  Deepcopy does not  copy immutable objects.

(env2) yoavkatz@Yoavs-MacBook-Pro unitxt % cat a.py                              
import copy

original_string = "hello"
copied_string = copy.deepcopy(original_string)

print(id(original_string))  # Output: some memory address
print(id(copied_string))  # Output: the same memory address 
(env2) yoavkatz@Yoavs-MacBook-Pro unitxt % python a.py
4343111536
4343111536

yoavkatz: Also we should change:

Attributes:
mappers (Dict[str, Dict[str, str]]): The mappers to use for mapping instance values.
Keys are the names of the fields to be mapped, and values are dictionaries
that define the mapping from old values to new values.

to 

Attributes:
mappers (Dict[str, Dict[str, **Any**]]): The mappers to use for mapping instance values.
Keys are the names of the fields to be mapped, and values are dictionaries
that define the mapping from old values to new values.

yoavkatz:
@dafnapension  Thanks for making the changes and adding the test.   Looks good.

I can not approve it (only you or @elronbandel ).

dafnapension:
Thank you, @yoavkatz , I will approve now.

dafnapension (approved):

PR #1268, opened 2024-10-08
A fix for a bug in metric pipeline

elronbandel:
Metric pipeline copy only the score, resulting in wrong metric name and confidence interval which are not copied. This pr fix this.


matanor (approved):
LGTM

PR #1274, opened 2024-10-10
Instruction in SystemFormet demo support.

piotrhelm:
In this PR, I'm introducing the ability to add instructions to the demo section in SystemFormat class. 

This can sometimes be useful when working with prompts that follow the multi-turn convention:
```
<|start_of_role|>user<|end_of_role|>{{Instruction}}
{{ICL_Q1}}
<|end_of_text|>
<|start_of_role|>assistant<|end_of_role|>{{ICL_A1}}
<|end_of_text|>
<|start_of_role|>user<|end_of_role|>{{Instruction}}
{{ICL_Q2}}
<|end_of_text|>
<|start_of_role|>assistant<|end_of_role|>{{ICL_A2}}
<|end_of_text|>
<|start_of_role|>user<|end_of_role|>{{Instruction}}
{{ICL_Q3}}
<|end_of_text|>
<|start_of_role|>assistant<|end_of_role|>{{ICL_A3}}
<|end_of_text|>
<|start_of_role|>assistant<|end_of_role|>
```
 especially when using templates that separate instructions from the input format.

### DIFF
--- a/src/unitxt/formats.py
+++ b/src/unitxt/formats.py
@@ -131,7 +131,7 @@ class SystemFormat(BaseFormat):
 
     """
 
-    demo_format: str = "{instruction}\\N{source}\\N{target_prefix}{target}\n\n"  #  example: "User: {source}\nAgent: {target}\n\n"
+    demo_format: str = "{source}\\N{target_prefix}{target}\n\n"  #  example: "User: {source}\nAgent: {target}\n\n"
     model_input_format: str = (
         "{system_prompt}\\N{instruction}\\N{demos}{source}\\N{target_prefix}"
     )
--- a/src/unitxt/formats.py
+++ b/src/unitxt/formats.py
@@ -131,7 +131,7 @@ class SystemFormat(BaseFormat):
 
     """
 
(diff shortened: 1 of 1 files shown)


src/unitxt/formats.py
```diff
@@ -131,7 +131,7 @@ class SystemFormat(BaseFormat):
 
     """
 
-    demo_format: str = "{source}\\N{target_prefix}{target}\n\n"  #  example: "User: {source}\nAgent: {target}\n\n"
+    demo_format: str = "{instruction}\\N{source}\\N{target_prefix}{target}\n\n"  #  example: "User: {source}\nAgent: {target}\n\n"
```
yoavkatz, line 134: I think we don't want to make this the default (so the empty prompt will have the instruction in the demo).   [ This is what causes the tests to fail].  We should do it, only in specific chat prompts.

elronbandel, line 134:
```suggestion
    demo_format: str = "{source}\\N{target_prefix}{target}\n\n"  #  example: "User: {source}\nAgent: {target}\n\n"
```

Issue #1260, opened 2024-10-07
Add num_instances to scores


yoavkatz:
Right now, we don't report the number of instances (only score, score_name, and confidence interval). 
This is always useful, but more so when we use the group_by options, which splits the original data into groups.

dafnapension:
implemented by PR #1282  , commit  5c7e2c31a45506410287410d1e02fae4459cb415

PR #1275, opened 2024-10-10
A PR suggested for issue #1260: Add num_instances to scores

dafnapension:


yoavkatz:
Thanks @dafnapension .  I think one change to the specification would be to place 

`num_of_evaluated_instances_in_group_group1` as part of the group score dictionary as  `num_of_evaluated_instances`.

This will make groups and global scores more consistent and avoid long names.

dafnapension:
Hi @yoavkatz , I think you closed this PR prematurely..
I liked your idea very much. Currently, Unitxt does not report the scores of the individual groups, and your idea to report them as well looks very fruitful to me.
One issue I would like your advice with:
Currently, both, "score"["instance"] and "score"["global"], are flat. This is relied upon, e.g., when adding prefix to score name.  Would you suggest to turn both scores (instance and global) to nested (and accordingly adjust the prefix, and alike)?

yoavkatz:
Hi.  Unitxt has two way to collect results ("subsets" and "groups").
Unitxt does create a separate entry in the score per group (see https://www.unitxt.ai/en/latest/docs/evaluating_datasets.html)

@elronbandel - we should clarify this in the documentation (by showing example outputs...)

elronbandel:
We just went over it together now. My view is that the only recommended way to group scores is with group_by. Which is documented in few guides and tutorials across the website. The support for presenting the number per group for the old grouping mechanism is unnecessary at this point imo. The number of instances per group in the new mechanism is working out of the box without adjustments (except a small fix dafna is working on).

dafnapension:
Hi @yoavkatz , @elronbandel ,   would you like me to try replace the existing code (that Sam wrote) in `InstanceMetric.process()`  that groups the instances by `task_data/group_id`,  by the general SplitSubsetsAndGroups?  then have the streams (one stream per group) go through the instance metric, which will yield a global-score per each group (!), and finally the streams would go through `JoinSubsetsAndGroups` which will yield a structured, nested global score, which I think, @yoavkatz , is what you had in mind. 
Into such a structure, I can add the lengths of the streams (== the groups) in a nested global score.

Is this an interesting plan?
We still need to think about the CI here
And the naming of the scores (the prefixes  `group_`  and `fixed_group_` )
And the general option to add prefix to a metric. 
And overall:  live with a global score that is not flat,  unlike the global scores that `metrics.py` emits today.

dafnapension:
replaced by PR #1282

PR #1282, opened 2024-10-14
try num of instances in nested global scores

dafnapension:


tests/library/test_fusion.py
```diff
@@ -429,6 +435,7 @@ def test_end_to_end(self):
                         },
                     },
                     "rte": {
+                        "num_of_instances": 5,
                         "f1_macro": 0.333,
```
elronbandel, line 439: line 448 should have number of instances in the global scopes

dafnapension: I do not quite understand this, in particular that I see in `SplitSubsetsAndGroups` that the same instance goes into two different streams of result.

elronbandel (approved):

Issue #1287, opened 2024-10-15
Issue when installing unitxt on MacOS


pawelknes:
I encountered an issue when using unitxt after pulling last changes related to images etc. The issue seems to concern `numpy` and `pytorch` packages. For example, when running `test_api.py` the following error occurres (at the bottom). Also, I was not able to install some packages from requirements due to issue with `numpy`. 
To solve the issue I had to use Python 3.9 (I tested also versions .8, .10 and .11) and downgrade `numpy` to specific version (I used 1.24 but didn't test with other releases).
I know the same issue occurred for other people as well. We tested it on Mac Intel and Mac M2 systems.

```
../../src/unitxt/api.py:114: in load_dataset
    return recipe().to_dataset(features=UNITXT_DATASET_SCHEMA)
../../src/unitxt/stream.py:231: in to_dataset
    {
../../src/unitxt/stream.py:232: in <dictcomp>
    key: Dataset.from_generator(
../../../../opt/anaconda3/envs/fm-eval/lib/python3.11/site-packages/datasets/arrow_dataset.py:1117: in from_generator
    return GeneratorDatasetInputStream(
../../../../opt/anaconda3/envs/fm-eval/lib/python3.11/site-packages/datasets/io/generator.py:28: in __init__
    self.builder = Generator(
../../../../opt/anaconda3/envs/fm-eval/lib/python3.11/site-packages/datasets/builder.py:374: in __init__
    self.config, self.config_id = self._create_builder_config(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <datasets.packaged_modules.generator.generator.Generator object at 0x127adc190>
config_name = None
custom_features = {'source': Value(dtype='string', id=None), 'target': Value(dtype='string', id=None), 'references': Sequence(feature=Va...'string', id=None), 'data_classification_policy': Sequence(feature=Value(dtype='string', id=None), length=-1, id=None)}
config_kwargs = {'features': {'source': Value(dtype='string', id=None), 'target': Value(dtype='string', id=None), 'references': Sequen...me': 'test'}, caching=False, copying=False), 'stream_name': 'test'}, caching=False, copying=False)}>, 'split': 'train'}
builder_config = None

    def _create_builder_config(
        self, config_name=None, custom_features=None, **config_kwargs
    ) -> Tuple[BuilderConfig, str]:
[...]

PR #1288, opened 2024-10-15
Remove split arg to support old hf datasets versions

elronbandel:
For: https://github.com/IBM/unitxt/issues/1287
Closes: https://github.com/IBM/unitxt/issues/1287


elronbandel:
@pawelknes can you see if this branch solve the issue you presented?

pawelknes:
Thanks, this solves the problem with `datasets` itself but there is still an issue with with `numpy` version (`RuntimeError: Numpy is not available`). It doesn't work with versions above 2., and I need it to bound it to some specific version. It may be necessary to add some specific release of the package

PR #1290, opened 2024-10-15
Changes to GenAI, WML  and HF inference engines

pawelknes:
Changes:
* `WMLInferenceEngine`: support for concurrent API calls and for vision models
* `IbmGenAiInferenceEngine` : added execution options
* `TextGenerationInferenceOutput` : added `stop_reason`, `seed` and `input_text` fields
* HF: added support for more types of HF models as well as inferencing with logprobs


src/unitxt/inference.py
```diff
@@ -107,6 +114,10 @@ def _set_inference_parameters(self):
                 if param_inst_val is None:
                     setattr(self, param, param_dict_val)
 
+    def get_model_details(self) -> Dict:
```
OfirArviv, line 117: I think this should return an empty dict or None (And the the return type should be Optional[Dict]).
This way if someone call the method and it was nt implemented, no bug will happen

src/unitxt/inference.py
```diff
@@ -809,16 +855,18 @@ class WMLInferenceEngine(
...
     def verify(self):
         super().verify()
 
-        if self.credentials is not None:
-            for key in self.credentials:
-                if key not in ["url", "apikey", "project_id", "space_id"]:
-                    raise ValueError(
-                        f'Illegal credential key: {key}, use only ["url", "apikey", "project_id", "space_id"]'
-                    )
+        assert (
```
OfirArviv, line 863: Are there no special keys in wml with more than 10?

pawelknes: Unfortunately 10 is the limit imposed by `generate` method of `ModelInference`. I put this check in unitxt as well so that users are aware of it

src/unitxt/inference.py
```diff
@@ -164,86 +175,629 @@ def _is_loaded(self):
...
+            peft_config.base_model_name_or_path
+        )
+
+    def _get_model_args(self) -> Dict[str, Any]:
+        args = super()._get_model_args()
+        args["tokenizer"] = self.processor
+        return args
+
+
+class HFInferenceEngine(HFInferenceEngineBase, LogProbInferenceEngine):
```
OfirArviv, line 411: I don't think we need another abstract class. So many is confusing. If this is the log prob class, the name should reflect it and it should be fully implemented. If there are problem with interfaces mismatch, let solves it @elronbandel

OfirArviv (changes requested):

src/unitxt/inference.py
```diff
@@ -164,86 +175,631 @@ def _is_loaded(self):
...
-    InferenceEngine, PackageRequirementsMixin, LazyLoadMixin
+class HFInferenceEngineBaseParamsMixin(Artifact):
+    max_new_tokens: int
+    do_sample: Optional[bool] = None
+    temperature: Optional[float] = None
+    top_p: Optional[float] = None
+    top_k: Optional[int] = None
+    num_beams: Optional[int] = None
+    repetition_penalty: Optional[float] = None
+    output_scores: Optional[bool] = None
```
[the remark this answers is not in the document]
pawelknes: Yup, I forgot to change that

src/unitxt/inference.py
pawelknes: I wanted to somehow separate pipeline-based HF inference engines (`HFPipelineBasedInferenceEngine` and `HFPeftInferenceEngine`) from "regular" HF inference engines which directly use a model to generate predictions, and also support logprobs (`HFAutoModelInferenceEngine` and `HFLlavaInferenceEngine`), but I didn't come up with some better way to do so.

Maybe just remove the `HFInferenceEngineBase` class, and keep separate implementations for engines:

`HFPipelineBasedInferenceEngine` -> `HFPeftInferenceEngine`

`HFInferenceEngine` (maybe some better name could be used for this) -> `HFAutoModelInferenceEngine` && `HFLlavaInferenceEngine`

What do you think?

src/unitxt/inference.py
```diff
@@ -164,86 +185,645 @@ def _is_loaded(self):
...
-    max_new_tokens: int
+    label: str
+
+    n_top_tokens: int = 5
+
+    device: Any = None
+    device_map: Any = None
+
+    use_fast_tokenizer: bool = True
     use_fp16: bool = True
```
OfirArviv, line 216: I don't see you use these and support this

src/unitxt/inference.py
```diff
@@ -164,86 +185,645 @@ def _is_loaded(self):
...
+    def _init_model(self):
         import torch
-        from transformers import AutoConfig, pipeline
+        from transformers import LlavaForConditionalGeneration
 
-        model_args: Dict[str, Any] = (
-            {"torch_dtype": torch.float16} if self.use_fp16 else {}
+        self.model = LlavaForConditionalGeneration.from_pretrained(
+            self.model_name,
+            torch_dtype=torch.float16,
```
OfirArviv, line 524: should be conditioned on the arg

src/unitxt/inference.py
```diff
@@ -164,86 +185,645 @@ def _is_loaded(self):
...
+            low_cpu_mem_usage=True,
```
OfirArviv, line 525: show have an arg

src/unitxt/inference.py
```diff
@@ -164,86 +185,645 @@ def _is_loaded(self):
...
-        # (even though the model might *just* fit to a single gpu), even if there is a gpu available, and this will
-        # cause an error because the data is always on the gpu
-        if torch.cuda.device_count() > 1:
-            assert device == torch.device(0)
-            model_args.update({"device_map": "auto"})
-        else:
-            model_args.update({"device": device})
 
-        task = (
+    def _init_model_and_processor(self):
```
OfirArviv, line 644: where is this called?

OfirArviv (changes requested):

src/unitxt/inference.py
pawelknes: Right, this should be in pipelines only

src/unitxt/inference.py
```diff
@@ -164,86 +185,645 @@ def _is_loaded(self):
...
+
+    device: Any = None
+    device_map: Any = None
+
+    use_fast_tokenizer: bool = True
     use_fp16: bool = True
+    load_in_8bit: bool = False
+
+    model: Any = InternalField(default=None, name="Inference object")
+    processor: Any = InternalField(default=None, name="Input processor (tokenizer)")
```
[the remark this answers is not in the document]
pawelknes: This is an input processor in general, for example in `HFLlavaInferenceEngine`, we use `AutoProcessor` so I wanted to keep it consistent and generalised

src/unitxt/inference.py
pawelknes: That was old naming, thanks for noticing

OfirArviv (approved):

PR #1295, opened 2024-10-19
recursive copy seems safer here

dafnapension:


elronbandel (approved):
Agree, though wish we could know the affect on performence

PR #1301, opened 2024-10-20
Add LMMSEvalInferenceEngine

elronbandel:


examples/evaluate_image_text_to_text_lmms_eval_inference.py
```diff
@@ -0,0 +1,36 @@
+from tqdm import tqdm
+from unitxt import settings
+from unitxt.api import evaluate, load_dataset
+from unitxt.inference import LMMSEvalInferenceEngine
+from unitxt.text_utils import print_dict
+
+with settings.context(
+    disable_hf_datasets_cache=False,
```
yoavkatz, line 8: We should document this new flag in the documentation in a relatively prominent way. (e.g. in one of the first tutorials that loads data from HF).

examples/evaluate_image_text_to_text_lmms_eval_inference.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    dataset = load_dataset(
+        card="cards.doc_vqa.lmms_eval",
+        template="templates.qa.with_context.title",
+        loader_limit=30,
+        streaming=True,
+    )
+
+    test_dataset = list(tqdm(dataset["test"], total=30))
```
yoavkatz, line 22: That does this do? Why do you nee tqdm?

examples/evaluate_image_text_to_text_lmms_eval_inference.py
yoavkatz: Need to add this example to example main page, and disable the run of this examples  by default in regressions.

examples/evaluate_image_text_to_text_lmms_eval_inference.py
```diff
@@ -0,0 +1,36 @@
...
+
+    predictions = inference_model.infer(test_dataset)
+    evaluated_dataset = evaluate(predictions=predictions, data=test_dataset)
```
yoavkatz, line 25: Do you get same results as with the inference engine  HFLlavaInferenceEngine?

PR #1307, opened 2024-10-24
a more elaborated message from performace-test-summary, and doc-string of card_profiler

dafnapension:


dafnapension:
Hi @elronbandel ,  Currently, the time reported for loading is the `cumtime` of `loader.load_data()`.
For eager mode, this also includes the time of yielding each and every record in the new stream. For non-eager mode -- this is not the case.
I think we should 'charge' the non-eager mode also for the `cumtime` of `loaders.split_limited_load`.  What is your view about this?

I attach the snakeviz for eager mode:

image


and of non-eager mode:

image

elronbandel (approved):

elronbandel:
@dafnapension I think you are absolutely right! the loading time in streaming should counted carefully. yet I dont understand why in eager mode more time spent on loading, I would think otherwise? or maybe is it because of loader limit?

dafnapension:
Hi @elronbandel , 
`loaders.load_data()` returns either `self.limited_load(dataset=dataset)` if loader is limited (which happens to be the case with `card_profiler`, "as a policy") or `MultiStream.from_iterables(dataset)` if not limited.
In both cases, returned is a dictionary of `DynamicStream`s.   In eager mode, the DynamicStream's __init__ exhausts the stream generator, thus yielding each and every instance brought by the loader. 
In non-eager mode, the generators of the DynamicStream continue to live as generators when `loaders.load_data()` ends.  This is why `loaders.load_data()`  takes longer for eager mode.

I will merge this branch now, and in another PR will change just a bit (wrap in names) the generators born by `loaders.load_data()` so that profiler can accurately count the number of times each is invoked, and we can check our calculations.

PR #1314, opened 2024-10-30
Add chat api format with standard open ai chat format

elronbandel:


src/unitxt/formats.py
```diff
@@ -273,19 +312,82 @@ def process(
...
+        target_prefix = self._retrieve_field_and_pop_from_instance(
+            instance=instance, field_name="target_prefix"
+        )
+
+        instance["source"] = (
+            self.tokenizer.apply_chat_template(
+                chat, tokenize=False, add_generation_prompt=True
+            )
+            + target_prefix
         )
```
OfirArviv, line 391: I don't understand why we need this.
Normally we won't be able to have a string "source" string. 
The source should be a dictionary in chat format, no?

OfirArviv: How do you envision the usage of it?
I envisioned it that the dictionary is in the source, and we just pass this to the model

src/unitxt/formats.py
```diff
@@ -200,43 +206,72 @@ def process(
...
-    _requirements_list = ["transformers"]
+class Message(TypedDict):
+    role: Literal["system", "user", "assistant"]
+    content: Union[str, List[Content]]
 
-    def prepare(self):
-        from transformers import AutoTokenizer
 
-        self.tokenizer = AutoTokenizer.from_pretrained(self.model_name)
+class OpenAIFormat(BaseFormat):
```
OfirArviv, line 232: We need documentation. About the output, the use case and so on. I think I understand what you meant to do, but I'm not sure it generalize well... lets make sure by first you writing what you meant to do

src/unitxt/formats.py
```diff
@@ -200,43 +206,72 @@ def process(
...
-        See more details in https://huggingface.co/docs/transformers/main/en/chat_templating
+Content = Union[TextContent, ImageUrlContent, ImageFileContent]
 
-    """
 
-    model_name: str
-    _requirements_list = ["transformers"]
+class Message(TypedDict):
+    role: Literal["system", "user", "assistant"]
+    content: Union[str, List[Content]]
```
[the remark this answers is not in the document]
elronbandel: For simple cases it over complicate things. better leave it simple for the 90% of people that do not use image contents.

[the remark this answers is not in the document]
OfirArviv: it should match the api. Users won't see it. What are apis for HF, OpenAI and etc.?

src/unitxt/formats.py
OfirArviv: Basically the user and unitxt won't know and cannot know what is the format. It only know that the backend somehow handle this

elronbandel: This is just for backward compatibility (for pepole who's using unitxt to get data with hf chat template in a textual format).

src/unitxt/formats.py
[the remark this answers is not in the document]
elronbandel: Pepole do look at what their data look like and they do want to see over complication. What are we losing by giving pepole output that is easier to understand and that is in a format they are mostly familiar with?

src/unitxt/formats.py
elronbandel: done

src/unitxt/formats.py
OfirArviv: Separate it to a different mechanism imo. We don't need to mix stuff that are not related from design pov

src/unitxt/formats.py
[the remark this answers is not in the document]
OfirArviv: it should reflect exactly what going into the model. That the most simple thing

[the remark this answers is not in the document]
OfirArviv: if there are multiple options that these api can recieve thats another matter.

src/unitxt/formats.py
OfirArviv: write what this format is. Give an example. If there are multiple formats list them

yoavkatz:
@elronbandel - I think we should make this self contained  
1. Add a HF inference engine (that uses the chat API)
2. Add en example code that calls this API.
3. Add it to catalog.

OfirArviv:
> * Add a HF inference engine (that uses the chat API)
> * Add en example code that calls this API both as text and with image.

We should wait for Pawel pr for this. The changes in the inference engine is small. No need to duplicate it

PR #1319, opened 2024-11-02
Better separate HF from Unitxt in loading HF datasets, to gain a more comparable loading time for non-eager and eager execution

dafnapension:
Currently, `Load Time` reported in the performance test summary is the cumulative time spent in `loader.load_data()` .  This is good for all types of loaders. 
However, still within `Loader.load_data()`, immediately following the download of the dataset (from HF, for example, in case of `LoadHF`), comes `MultiStream.from_iterables`, the first introduction of the downloaded data into unitxt's recipe. In this introduction, eager-execution differs in its action from non-eager-execution: 
In eager-mode, unitxt loops over **each and every instance**, to include it into the `ListStream`s constituting the  `MultiStream` being generated. Whereas in the non-eager mode, unitxt just creates generators, for the `GeneratorStream`s constituting the `MultiStream` generated in this case, and returns from `load_data()`.

Now, with `load_iterables` recently introduced into `loaders.py` by @elronbandel , we can clearly fetch net load time of datasets, excluding the introduction of each and every instance into unitxt that is only done in eager mode.

Here are demonstrating snakeviz-s:
I tweaked `card_prifler` to only go through load and list streams of the cards (skipping over all preprocessing steps, and verbalization, etc).
For eager mode, more time is spent in `Loader.load_data()` than in `LoaderHF.load_iterables`:
image

in `stream.DynamicStream.__post_init__` we find the time spent in the above gap:
image

Whereas for non-eager mode -- the same (the time to generate the generators is unnoticeable):
image

### DIFF
--- a/performance/card_profiler.py
+++ b/performance/card_profiler.py
@@ -75,7 +75,7 @@ class CardProfiler:
         ms = recipe.verbalization.process(ms)
         return recipe.finalize.process(ms)
 
-    def profiler_print_first_dicts(self, ms: MultiStream, card_name: str):
+    def profiler_list_all_streams(self, ms: MultiStream, card_name: str):
         logger.info(
             f"The multistream generated for card '{card_name}' has {len(ms)} streams of the following lengths:"
         )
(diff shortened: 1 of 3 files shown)


dafnapension:
Hi @elronbandel ,
I am not sure how important is the issue raised here. We usually compare performance of PR against main either when both are eager of both are non-eager, so why bother. On the other hand - why not be accurate, and have each mode 'pay' for its going over the fresh instances in its reported Net Time.  On the third hand: we will need a careful identification of the time spent on 'pure downloading'  (excluding the introduction of new instances into unitxt) for all the different types of Loader-s. Not too complicated, but still. What is your view about this?

dafnapension:
PR #1436 covers this

PR #1329, opened 2024-11-05
Avoid recursive copy in ApplpyMetric for cases of a single metric

dafnapension:
Because that recursive copy is only needed against one metric changing stream contents before another metric is applied.
So overall: payless in time and space for all cases of a single metric.
Earn a few cents in Performance.


dafnapension:
Redundant once PR #1448 is merged into main, on Dec. 24

Issue #1331, opened 2024-11-06
Several processors are added twice to the catalog


martinscooper:
Some or them are:
- processors.extract_mt_bench_rating_judgment
- processors.extract_mt_bench_label_judgment
- processors.regex_parser_from_prediction
- processors.literal_eval

PR #1337, opened 2024-11-10
A swifty PR suggested for issue #1331

dafnapension:
closes: https://github.com/IBM/unitxt/issues/1331

Added a test to identify cases of more than one prepare file generating the same catalog file.
Consistency test works in granularity of prepare-files (can not identify two or more generations of same catalog file, within same prepare file; only across different prepare-files). But for now, and because typically, not too many `add_to_catalog` within same prepare_file, and hence can be manually tested for uniqueness, this PR may suffice.


prepare/processors/processors.py
```diff
@@ -171,31 +171,6 @@
     overwrite=True,
 )
 
-add_to_catalog(
```
elronbandel, line 174: Why did you delete these?

elronbandel (changes requested):

prepare/processors/processors.py
elronbandel: I mean where are they added somewhere else? in the same file?

dafnapension:
Here is an excerpt from `processors.py` of current `main` branch, please notice the line numbers and contents:

image

Now please notice the excerpt from same file of same branch, different lines, immediately following the previous ones, but same contents:

image

Someone pasted twice by mistake, and I removed once.

dafnapension:
On this occasion, 'while the hood is open', I added a test to catelog-consistency, to verify that not two prepare-files generate the same catalog entry (overwriting each other).  There is still no test to catch errors like the one triggered this PR: overwriting in within same prepare-file, because the automatic test works in granularity of whole prepare files.
I commented that errors like this one (that triggered this PR) can be caught (evidently, not too obviously..) by the eye..

elronbandel (approved):

Issue #1042, opened 2024-07-22
Catalog preparation takes over 30 minutes which is too long


yoavkatz:
Also, we don't see the print out to see what takes the time.

https://github.com/IBM/unitxt/actions/runs/10042751612/job/27753878493?pr=1034

PR #1339, opened 2024-11-11
test_preparation in less than 10 min

dafnapension:
closes: https://github.com/IBM/unitxt/issues/1042


dafnapension:
Maybe, concurrency of cards preparation conflicts with the common caches of the loaders.

elronbandel (approved):

PR #1343, opened 2024-11-12
Add multi api inference engine

elronbandel:
With multi api inference engine you can feed into your application:
```python
     MyClass(
           model=CrossProviderInferenceEngine("llama-3.1-405b", api="aws")
      )
 ```
 Then change it by editing the model.api argument or by general settings:
 You can set `unitxt.settings.default_provider="watsonx"`
 or `export UNITXT_DEFAULT_PROVIDER="watsonx"`

### CODE REVIEW

File: examples/evaluate_benchmark_with_custom_api.py [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+from unitxt.text_utils import print_dict
+
+with unitxt.settings.context(
+    default_inference_api="watsonx",  # option a to define your home api
+    default_format="formats.chat_api",
+    disable_hf_datasets_cache=False,
+):
+    data = load_dataset("benchmarks.glue[max_samples_per_subset=5]", split="test")
+
+    model = get_from_catalog(
```
yoavkatz, line 12: I don't see why we need to use the catalog.

I think it's better to show this as:

MultiAPIInferenceEngine(model="llama_3_8b_instruct",api="watsonx",top_k=1)

It's more consistent with the other engines, and does not require non python  syntax.

src/unitxt/catalog/engines/model/llama_3_8b_instruct.json
```diff
@@ -0,0 +1,12 @@
+{
+    "__type__": "multi_api_inference_engine",
```
yoavkatz, line 2: Not sure if adding to the catalog is needed.  The API mapping can be done in a single place, and not repeated per model.

src/unitxt/catalog/system_prompts/general/be_concise.json
```diff
@@ -0,0 +1,4 @@
+{
+    "__type__": "textual_system_prompt",
+    "text": "be concise. at every point give the shortest acceptable answer."
```
yoavkatz, line 3:
```suggestion
    "text": "Be concise. At every point give the shortest acceptable answer."
```

yoavkatz (approved):

docs/docs/inference.rst
```diff
@@ -0,0 +1,114 @@
...
+Inference
+==============
+
+.. note::
+
+   This tutorial requires a :ref:`Unitxt installation <install_unitxt>`.
+
+Introduction
+------------
+Unitxt offers a wide array of :class:`Inference Engines <unitxt.inference>` for running models either locally (using HuggingFace, Ollama, and VLLM) or by making API requests to services like WatsonX, AWS, and Together AI.
```
yoavkatz, line 13:
```suggestion
Unitxt offers a wide array of :class:`Inference Engines <unitxt.inference>` for running models either locally (using HuggingFace, Ollama, and VLLM) or by making API requests to services like WatsonX, OpenAI, AWS, and Together AI.
```

docs/docs/inference.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Finally, evaluate the predictions and obtain final scores:
+
+.. code-block:: python
+
+    evaluate(predictions=predictions, data=dataset)
+
+Calling Models Through APIs
+---------------------------
+Calling models through an API is even simpler and is primarily done using one class: :class:`CrossProviderInferenceEngine <unitxt.inference.CrossProviderInferenceEngine>`.
```
yoavkatz, line 72: We should add at some point the HF  as a provider. So I would not split it to between APIs and local.  Instead you can call specific inference engine , or use the CrossProvider inference engine, which gives a standard API to multiple providers.

yoavkatz (approved):

PR #1350, opened 2024-11-14
Add ability to load_dataset without a template for simpler usage for beginners

elronbandel:


src/unitxt/standard.py
```diff
@@ -411,9 +411,13 @@ def prepare(self):
         assert (
             self.template_card_index is None or self.template is None
         ), f"Specify either template ({self.template}) or template_card_index ({self.template_card_index}) but not both"
-        assert not (
-            self.template_card_index is None and self.template is None
-        ), "Specify either template or template_card_index in card"
+
+        if self.template_card_index is None and self.template is None:
+            self.template_card_index = 0
+            logger.warning(
```
yoavkatz, line 417: I think this will not work if the templates in the card is a dictionary and not a list.

elronbandel: True. Will fix it.

PR #1356, opened 2024-11-17
Align augmenters with task and types mechanisms

elronbandel:
Augmentors mechanism worked differently for different modalities now they all work the same in a safe way.


prepare/augmenters/image_augmentors.py
```diff
@@ -1,21 +1,18 @@
 from unitxt import add_to_catalog
-from unitxt.augmentors import (
-    ImagesAugmentor,
-)
 from unitxt.image_operators import GrayScale, GridLines, Oldify, PixelNoise
```
yoavkatz, line 2: Need to move to "augmenters" directory.

src/unitxt/ui/run.py
```diff
@@ -49,7 +49,7 @@ def run_unitxt_entry(
     system_prompt=None,
     format=None,
     num_demos=0,
-    augmentor=None,
+    augmenter=None,
```
yoavkatz, line 52: Need to revert.

src/unitxt/ui/run.py
```diff
@@ -69,8 +69,8 @@ def run_unitxt_entry(
         format = None
     if not isinstance(num_demos, int):
         num_demos = 0
-    if not isinstance(augmentor, str):
-        augmentor = None
+    if not isinstance(augmenter, str):
```
yoavkatz, line 72: Need to revert.

prepare/augmenters/text_augmentors.py
```diff
@@ -0,0 +1,31 @@
...
+
+operator = AugmentPrefixSuffix(
+    prefixes=pattern_distribution,
+    suffixes=pattern_distribution,
+    prefix_len=5,
+    suffix_len=5,
+    remove_existing_whitespaces=True,
+)
+
+add_to_catalog(operator, "augmentors.text.prefix_suffix", overwrite=True)
```
yoavkatz, line 22: They should be ""augmentors.text.whitespace_prefix_suffix" (There could be other prefix_suffix augmenters, that add other things.

PR #1357, opened 2024-11-17
Add option selecting huggingface inference engine

elronbandel:


OfirArviv (approved):

OfirArviv:
maybe also add a test that make sure you are using all tokens?
Option 1 is A
and 2 is A blah blah

And make sure they are different

PR #1361, opened 2024-11-17
Add target_prefix erasing post processor by default

elronbandel:


src/unitxt/templates.py
```diff
@@ -127,12 +129,19 @@ def process(
             serialized_inputs
         )
 
+        post_processors = self.postprocessors
+        if self.target_prefix is not None:
+            post_processors = [
+                PostProcess(Replace(old=self.target_prefix, new="")),
```
yoavkatz, line 135: Hi. This post processor needs to be added only on prediction.  It should use a more refined regex, that 1. Ensure it's from the start of the prediction and only replaced one. 2. Allow whitespaces (at least a the start).

It also needs to be documented on the template docs.

yoavkatz (changes requested):
See comments below.

yoavkatz (approved):

Issue #1341, opened 2024-11-11
Need a way to rename and deprecate artifacts


yoavkatz:
The catalog today has some wrongly  named artifacts in the wrong local. We need a backward compatible way to migrate to it.

1) Add __deprecation__ string field to each artificat. If set to anything but 'None', a message will be printed when loading the artifact:

"Artifact XXXX was loaded from catalog but is deprecated: {__deprecation__}. " 

The catalog UX should highlight that they are deprecated (e.g. red flag)

2)We would have a way to link an artifact to another artifact.

artifact = ArtifactReference(reference="tasks.my_task" )
add_to_catalog(artifact,"tasks.my_old_task")

This will make loading "tasks.my_old_task" return the "Task.my_task")

PR #1363, opened 2024-11-18
artifact link

dafnapension:
closes: https://github.com/IBM/unitxt/issues/1341


elronbandel:
It looks good, just not sure, how to support the overwrite args.
Not sure how we could pass the arguments from the old artifact to the new one. 
Lets say someone uses `metrics.accurcay[lower_case=True]` and `metrics.accuracy` points to its new name `metrics.exact_match` how can we pass the arguments  to the new one `[lower_case=True]`? @yoavkatz @dafnapension  what do you think?

dafnapension:
Hi @elronbandel ,  
As it is now, the `new_artifact` is a whole: an instantiated artifact. It is returned as is. Potentially, it does not have at all the fields specified in the overwrites of the original artifact referenced. 
I think that a clear documentation can cover this issue.. :-)

elronbandel:
The issue is old code with references to deprecated catalog names. 
If someone  has `metrics.accurcay[lower_case=True]` and we referred `metrics.accurcay->metrics.exact_match` his code wont work as before due to the arguments overwrite not being passed: `[lower_case=True]`

src/unitxt/artifact.py
```diff
@@ -412,6 +412,18 @@ def get_raw(obj):
     return shallow_copy(obj)
 
 
+class ArtifactLink(Artifact):
+    new_artifact: Artifact
+    is_self_deprecated: bool
```
yoavkatz, line 417: In think deprecation and linking are distinct properties.

We can deprecate something without having a new link to it.  So every artifact should have a deprecated field.

src/unitxt/artifact.py
```diff
@@ -138,6 +139,14 @@ class Artifact(Dataclass):
     )
     __id__: str = InternalField(default=None, required=False, also_positional=False)
 
+    __is_deprecated__: bool = NonPositionalField(
```
yoavkatz, line 142: We should add description of the features to the 

https://www.unitxt.ai/en/latest/docs/saving_and_loading_from_catalog.html

src/unitxt/artifact.py
```diff
@@ -138,6 +139,14 @@ class Artifact(Dataclass):
...
+        default=False, required=False, also_positional=False
+    )
+
+    __replacing_artifact__: Any = NonPositionalField(
```
yoavkatz, line 146: I think __replacing_artifact__ should in a dedicated class called ArtifactLink.  , you you had before, because it would be odd that that an artifact of class Operator could link to an artifact of type Card.

tests/library/test_artifact.py
```diff
@@ -217,3 +217,85 @@ def test_misconfigured_data_classification_policy(self):
...
+    def test_artifact_link_with_deprecation_warning(self):
+        with temp_catalog() as catalog_path:
+            rename = Rename(field_to_field={"old_field_name": "new_fild_name"})
+            add_to_catalog(
+                rename,
+                "rename.for.test.dict",
+                catalog_path=catalog_path,
+            )
+
+            class LinkExample2(Artifact):
```
yoavkatz, line 252: There should be only one class called ArtifactLink which will be part of the official code, and be used by all users.

src/unitxt/artifact.py
dafnapension: @yoavkatz , I removed field `__is_deprecated__`  altogether, since we already have the `@deprecation` decorator from `deprecation_utils`.  Decorating class `RenameFields`, for example.  
And moved the link to the other Artifact to class `ArtifactLink` as you suggested.

src/unitxt/artifact.py
dafnapension: Done.

tests/library/test_artifact.py
dafnapension: Done.

src/unitxt/artifact.py
yoavkatz: Thanks Dafna.  
 There is a difference between deprecating a class and deprecating an instance.  
For example, I can have a card saved as "card.sst" and at some point I want to call it "card.sst2".
I want to make the change be backward compatible, so "cards.sst" becomes a deprecated with warning.  It is also replaced with ArtficactLink  to "card.sst2".

Note that "Card" class was never deprecated

I also think a link should always be to acatalog entry and not an object.

src/unitxt/artifact.py
```diff
@@ -237,6 +244,30 @@ def from_dict(cls, d, overwrite_args=None):
     @classmethod
     def load(cls, path, artifact_identifier=None, overwrite_args=None):
```
yoavkatz, line 245: I suggest we override the load method in the derived ArtifactLink class, and not condition the behavior based on the existing the  attribute.

src/unitxt/artifact.py
```diff
@@ -456,20 +509,27 @@ def fetch_artifact(artifact_rep) -> Tuple[Artifact, Union[AbstractCatalog, None]
     (5) Otherwise, check that the artifact representation is a dictionary and build an Artifact object from it.
     """
     if isinstance(artifact_rep, Artifact):
+        if isinstance(artifact_rep, ArtifactLink):
```
yoavkatz, line 512: I'm not sure about this code.  If ww do fetch_artificant(ArtifiactLink("cards.sst2")), then we don't expect to get "cards.sst2", bur rather the real Card from "cards.sst2".  
Can you add a check a test to see this is what we get?

src/unitxt/artifact.py
```diff
@@ -456,20 +509,27 @@ def fetch_artifact(artifact_rep) -> Tuple[Artifact, Union[AbstractCatalog, None]
...
+            return artifact_rep.artifact_linked_to, None
         return artifact_rep, None
 
     # If local file
     if isinstance(artifact_rep, str) and Artifact.is_artifact_file(artifact_rep):
-        return Artifact.load(artifact_rep), None
+        artifact_to_return = Artifact.load(artifact_rep)
+        if isinstance(artifact_rep, ArtifactLink):
+            artifact_to_return = artifact_to_return.artifact_linked_to
```
yoavkatz, line 520: Same here.  The fetch_artificat should return the artificat, not the name of the artifact we link to.

dafnapension: Hi @yoavkatz , and thanks again for your neat idea about the dedicated `load` of `ArtifactLink`. I just expanded it to also allow link to link to link (imagine that..  with the `overwrites` that @elronbandel emphasized floating all the way over the links in the middle, until they can be applied on the final, non-link `Artifact` that can digest them - only it is expected to have these keys that show in the overwrites).

And thanks for hour new comment. I see what you mean, and now fixing.

src/unitxt/artifact.py
```diff
@@ -396,6 +409,53 @@ def verify_instance(
         return instance
 
 
+class ArtifactLink(Artifact):
+    # the artifact linked to, expressed by its catalog id
+    artifact_linked_to: str = NonPositionalField(
+        default=None, required=False, also_positional=False
```
yoavkatz, line 415: Shouldn't it be a required field without a default? Any use case gor not initializing it?

dafnapension: Yes, agree..  just copied over without thinking. Fixed now..

src/unitxt/artifact.py
dafnapension: Hi @yoavkatz , I also added the requested documentation:
image

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
 Modifying Catalog Assets on the Fly
 -----------------------------------
 
-To modify a catalog asset's fields dynamically, use the syntax: `asset.name[key_to_modify=new_value]`. To assign lists, use: `asset.name[key_to_modify=[new_value_0, new_value_1]]`. For instance, to change the metric list of a task:
+To modify a catalog asset's fields dynamically, upon fetching the asset from the catalog, and instantiating it into a python object of type ``unitxt.Artifact``, use the syntax: ``asset.name[key_to_modify=new_value]``.
```
yoavkatz, line 59:
```suggestion
To modify a catalog asset's fields dynamically, upon fetching the asset from the catalog, use the syntax: ``artifact_name[key_to_modify=new_value]``. 
```

I think it's simpler.

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
 Modifying Catalog Assets on the Fly
 -----------------------------------
 
-To modify a catalog asset's fields dynamically, use the syntax: `asset.name[key_to_modify=new_value]`. To assign lists, use: `asset.name[key_to_modify=[new_value_0, new_value_1]]`. For instance, to change the metric list of a task:
+To modify a catalog asset's fields dynamically, upon fetching the asset from the catalog, and instantiating it into a python object of type ``unitxt.Artifact``, use the syntax: ``asset.name[key_to_modify=new_value]``. 
+To assign lists, use: ``asset.name[key_to_modify=[new_value_0, new_value_1]]``.
```
yoavkatz, line 60:
```suggestion
To assign lists, use: ``artifact_name[key_to_modify=[new_value_0, new_value_1]]``. 
```

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
 Modifying Catalog Assets on the Fly
 -----------------------------------
 
-To modify a catalog asset's fields dynamically, use the syntax: `asset.name[key_to_modify=new_value]`. To assign lists, use: `asset.name[key_to_modify=[new_value_0, new_value_1]]`. For instance, to change the metric list of a task:
+To modify a catalog asset's fields dynamically, upon fetching the asset from the catalog, and instantiating it into a python object of type ``unitxt.Artifact``, use the syntax: ``asset.name[key_to_modify=new_value]``. 
+To assign lists, use: ``asset.name[key_to_modify=[new_value_0, new_value_1]]``. 
+To assign dictionaries, use: ``asset.name[key_to_modify={new_key_0=new_value_0,new_key_1=new_value_1}]``.
```
yoavkatz, line 61:
```suggestion
To assign dictionaries, use: ``artifact_name[key_to_modify={new_key_0=new_value_0,new_key_1=new_value_1}]``.
```

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
...
+For instance, to change the metric list and the reference specification of a task:
 
 .. code-block:: python
 
     from unitxt.card import TaskCard
 
     card = TaskCard(
         ...
-        task="tasks.my_task[metrics=[metrics.accuracy, metrics.f1[reduction=median]]]"
+        task="tasks.my_task[metrics=[metrics.accuracy, metrics.f1[reduction=median]],reference_fields={output=int}]"
```
yoavkatz, line 71: The dict example is a bit made it, maybe keep the example with just a list?

If we want a more real example, we can use overriding the credential dictionary fields in WMLInferenfeEgnine

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
...
 
     my_task = get_from_catalog("tasks.my_task")
 
+A Catalog Asset Linking to Another Catalog Asset
+------------------------------------------------
+
+A catalog asset can be just a link to another asset. This feature comes handy when for some reason, ``asset1`` -- the name of an asset, which reflects its place in the catalog, is changed to ``asset2``, while much code already exists where the old name of the asset, ``asset1`` is hard coded.
+In such a case, an asset of type :class:`ArtifactLink <unitxt.artifact.ArtifactLink>`, that links to ``asset2``, can take the place of ``asset1`` in the catalog. 
+When ``asset1`` is accessed from an existing code, Unixt Catalog realizes that the asset fetched from position ``asset1`` is an ``ArtifactLink``, so it continues to ``asset2`` -- the Artifact linked to by ``asset1``, instantiates and returns it.
+If that linked-to asset, ``asset2``, turns out to be an ``ArtifactLink`` as well, Unitxt Catalog continues along the links, until a non-link Artifact is reached, and that one is instantiated as a python object of type non-link ``unitxt.Artifact`` and returned.
```
yoavkatz, line 91: I think we can simply it - and remove the link to link discussion. Can we also add an example of how to ldo the linking?

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
...
 .. code-block:: python
 
     from unitxt import get_from_catalog
 
     my_task = get_from_catalog("tasks.my_task")
 
+A Catalog Asset Linking to Another Catalog Asset
+------------------------------------------------
+
+A catalog asset can be just a link to another asset. This feature comes handy when for some reason, ``asset1`` -- the name of an asset, which reflects its place in the catalog, is changed to ``asset2``, while much code already exists where the old name of the asset, ``asset1`` is hard coded.
```
yoavkatz, line 88:
```suggestion
A catalog asset can be just a link to another asset. This feature comes handy when for some reason, we want to change the catalog name of an existing asset, while there is already code that uses the old name of the asset and we want to avoid non-backward compatible changes.
```

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
...
+In such a case, an asset of type :class:`ArtifactLink <unitxt.artifact.ArtifactLink>`, that links to ``asset2``, can take the place of ``asset1`` in the catalog.
```
yoavkatz, line 89:
```suggestion
In such a case, we can save the asset as as ``asset2`` and create  asset of type :class:`ArtifactLink <unitxt.artifact.ArtifactLink>`, that links to ``asset1`` ``to asset2``
```

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
...
     my_task = get_from_catalog("tasks.my_task")
 
+A Catalog Asset Linking to Another Catalog Asset
+------------------------------------------------
+
+A catalog asset can be just a link to another asset. This feature comes handy when for some reason, ``asset1`` -- the name of an asset, which reflects its place in the catalog, is changed to ``asset2``, while much code already exists where the old name of the asset, ``asset1`` is hard coded.
+In such a case, an asset of type :class:`ArtifactLink <unitxt.artifact.ArtifactLink>`, that links to ``asset2``, can take the place of ``asset1`` in the catalog. 
+When ``asset1`` is accessed from an existing code, Unixt Catalog realizes that the asset fetched from position ``asset1`` is an ``ArtifactLink``, so it continues to ``asset2`` -- the Artifact linked to by ``asset1``, instantiates and returns it.
+If that linked-to asset, ``asset2``, turns out to be an ``ArtifactLink`` as well, Unitxt Catalog continues along the links, until a non-link Artifact is reached, and that one is instantiated as a python object of type non-link ``unitxt.Artifact`` and returned.
+
```
yoavkatz, line 92: Can we also add the deprecation message  description and example?

docs/docs/saving_and_loading_from_catalog.rst
```diff
@@ -56,28 +56,40 @@ To use catalog objects, simply specify their name in the Unitxt object that will
...
     from unitxt import get_from_catalog
 
     my_task = get_from_catalog("tasks.my_task")
 
+A Catalog Asset Linking to Another Catalog Asset
+------------------------------------------------
+
+A catalog asset can be just a link to another asset. This feature comes handy when for some reason, ``asset1`` -- the name of an asset, which reflects its place in the catalog, is changed to ``asset2``, while much code already exists where the old name of the asset, ``asset1`` is hard coded.
+In such a case, an asset of type :class:`ArtifactLink <unitxt.artifact.ArtifactLink>`, that links to ``asset2``, can take the place of ``asset1`` in the catalog. 
+When ``asset1`` is accessed from an existing code, Unixt Catalog realizes that the asset fetched from position ``asset1`` is an ``ArtifactLink``, so it continues to ``asset2`` -- the Artifact linked to by ``asset1``, instantiates and returns it.
```
yoavkatz, line 90:
```suggestion
When ``asset1`` is accessed from an existing code, Unixt Catalog realizes that the asset fetched from position ``asset1`` is an ``ArtifactLink``, so it continues to ``asset2`` -- the Artifact linked to by ``asset1``, instantiates and returns it.
```

src/unitxt/artifact.py
```diff
@@ -237,6 +244,11 @@ def from_dict(cls, d, overwrite_args=None):
     @classmethod
     def load(cls, path, artifact_identifier=None, overwrite_args=None):
         d = artifacts_json_cache(path)
+        if "artifact_linked_to" in d and d["artifact_linked_to"] is not None:
```
yoavkatz, line 247: I think  we should check if it is an instance of of ArtifactLink instead of checking for fields.

src/unitxt/artifact.py
yoavkatz: Yes. Looks right.

tests/library/test_artifact.py
```diff
@@ -217,3 +219,239 @@ def test_misconfigured_data_classification_policy(self):
 
         # "Fixing" the env variable so that it does not affect other tests:
         del os.environ["UNITXT_DATA_CLASSIFICATION_POLICY"]
+
+    def test_artifact_link(self):
+        rename = Rename(field_to_field={"old_field_name": "new_fild_name"})
```
yoavkatz, line 224:
```suggestion
        rename = Rename(field_to_field={"old_field_name": "new_field_name"})
```

src/unitxt/artifact.py
yoavkatz: Thanks!   I've added some followup feedback.

yoavkatz:
I've added example of 2 artifacts that are replaced by links, and one which is deprecated in any case.

Can you make the Catalog  UI reflect deprecated / linked artifacts?

docs/docs/saving_and_loading_from_catalog.rst
dafnapension: Agree! Done.

docs/docs/saving_and_loading_from_catalog.rst
dafnapension: Agree! removed the dict from the example.  I was not sure what `reduction=median` is, but this is besides the point.

docs/docs/saving_and_loading_from_catalog.rst
dafnapension: Agree!
Removed link to link discussion (let them try out, if interested, or peek into `test_artifact` :-)), and showed how to link to `asset2` -- the new home of the `asset1` that their code refers to.

docs/docs/saving_and_loading_from_catalog.rst
dafnapension: took your rephrasing.

docs/docs/saving_and_loading_from_catalog.rst
dafnapension: Thanks, @yoavkatz , I adopted your concise, clear phrasings, and added a subsection about deprecation.
I also added one more test.

image

dafnapension:
Hi @yoavkatz , is that what you meant about reflecting in the UI? You wanted to see how these would look? (I pasted here under)  of you mean if I can find a way to color these fields (of deprecation msg, and artifact linked to) in a special color?

image

image

image

dafnapension:
While the hood is open.. I see the explanation of   does not go through nicely to the UI  (all `\t` and `\n`  show `t` and `n`. I know how to fix this. Will push a fix.

image

dafnapension:
Managed at last..:
image

docs/docs/saving_and_loading_from_catalog.rst
yoavkatz: I think ArtificatLink has a default deprecation message.  
We should have an example where there is no link (just a deprecation message).

yoavkatz: We should add another example with deprecation on a "regular" artifact.

dafnapension:
Hi @yoavkatz ,  I fought the red color, but I think I won. please let me know if this is anywhere near what you had in mind:

image

which also shows on the dir level:

image

yoavkatz:
Thanks Dafna.

Looking good. Two suggestions:
1. "Deprecation message:" as a prefix to the deprecation message.
2. Remove the __deprecation__ message from the displayed yaml (as it is redundant)

This is also true from "__descriptions__" and "__tags__" which would appear in the message and not as fields in the object.

docs/docs/saving_and_loading_from_catalog.rst
yoavkatz: Yes. Thanks.

yoavkatz (approved):

PR #1365, opened 2024-11-18
add serializers to catalog + new table operators

ShirApp:


src/unitxt/struct_data_operators.py
```diff
@@ -312,6 +312,32 @@ def process_rows(self, rows: List[List]) -> str:
         return rows_html
 
 
+class SerializeTableAsConcatenation(SerializeTable):
+    """Cancat Serializer.
```
elronbandel, line 316:
```suggestion
    """Concat Serializer.
```

PR #1382, opened 2024-11-21
Add vectara's hhem2.1 faithfulness model as a metric

lilacheden:


prepare/metrics/hhem.py
```diff
@@ -0,0 +1,48 @@
...
+    ("I am in United States", "I am in California."),
+]
+
+task_data = [{"contexts": [p[0]], "answer": p[1]} for p in pairs]
+
+metric = MetricPipeline(
+    main_score="score",
+    preprocess_steps=[
+        Copy(
+            field_to_field={
```
yoavkatz, line 18: Why do these copies of fields?

If I understand this is a reference less  a faithfulness metrics that should receive the model prediction and the contexts (from task_data).

lilacheden: Just added a fix (that required a bigfix in operator as well), but the idea remains.
This metric needs to support two usecases:
1. regular flow: model prediction, and contexts from task_data 
2. running on external rag data: the input is a dict with the fields "contexts", "answer" 
This should capture both

PR #1385, opened 2024-11-21
Add support for OpenAI custom base url and default headers + RITS Inference engine

martinscooper:
- Added support for passing open base url, default headers, and setting credentials in OpenAIInferenceEnginer
- Added RITSInferenceEngine (also to CrossProviderInferenceEngine)


src/unitxt/inference.py
```diff
@@ -1453,27 +1456,54 @@ class OpenAiInferenceEngine(
     }
     data_classification_policy = ["public"]
     parameters: Optional[OpenAiInferenceEngineParams] = None
+    base_url: Optional[str] = None
+    default_headers: Optional[dict[str, str]] = None
+    credentials: Optional[CredentialsOpenAi] = {}
 
     def get_engine_id(self):
         return get_model_and_label_id(self.model_name, self.label)
```
OfirArviv, line 1465: is this sued anymore?

martinscooper: This allows for the `OpenAiInferenceEngine` to be used against any OpenAI API. We are adding a specific engine class for RITS, but users may the `OpenAiInferenceEngine` against any other OpenAI API endpoint, without needing to create a new class. For example, in the `test_rits_inference()` test, it is shown how the user can also target RITS, without using the `RITSInferenceEngine `.

OfirArviv: Sorry. I wasn't clear. I meant if the `def get_api_param` is needed, as I see you added the new `def _prepare_credentials(self) -> CredentialsOpenAi:`
which does te same and more

yoavkatz: This is used in a derived class.

yoavkatz (approved):

OfirArviv (changes requested):

OfirArviv:
@yoavkatz Is RITS available from outside the network? Maybe that is why the test fails

src/unitxt/inference.py
```diff
@@ -1553,6 +1572,32 @@ def get_return_object(self, predict_result, response, return_meta_data):
...
 
+class VLLMRemoteInferenceEngine(OpenAiInferenceEngine):
+    label: str = "vllm"
+
+
+class RITSInferenceEngine(OpenAiInferenceEngine):
+    label: str = "rits"
+
+    def get_default_headers(self):
+        return {"RITS_API_KEY": self.credentials["api_key"]}
```
OfirArviv, line 1583: why do we need this as an header? It seems strange as we already pass the api key

martinscooper:
@OfirArviv @yoavkatz IMO we should remove the `VLLMRemoteInferenceEngine` as it adds no functionality to `OpenAiInferenceEngine`. In addition, other servers/frameworks can potentially expose LLMs using the OpenAI API, so this is not something exclusive to a remote vLLM instance. Users can directly create an `OpenAiInferenceEngine` and set the lable accordingly.

OfirArviv:
> @OfirArviv @yoavkatz IMO we should remove the `VLLMRemoteInferenceEngine` as it adds no functionality to `OpenAiInferenceEngine`. In addition, other servers/frameworks can potentially expose LLMs using the OpenAI API, so this is not something exclusive to a remote vLLM instance. Users can directly create an `OpenAiInferenceEngine` and set the lable accordingly.

We cannot right now because some users are using it. But I defined it as minimal as it can be, basically just setting the label

src/unitxt/inference.py
OfirArviv: tested it, and it seem we need it. Strange.

OfirArviv (approved):

PR #1386, opened 2024-11-22
Check if unitxt_pkg is None

yhwang:
When using HuggingFace APIs to load datasets and metrics, the unitxt_pkg is None. Check it before accessing its properties. The current code breaks the installation approach documented here: https://www.unitxt.ai/en/latest/docs/lm_eval.html#installation.


ruivieira:
Thank you @yhwang.
TrustyAI's LMEval is being affected by this https://github.com/trustyai-explainability/trustyai-service-operator/issues/367

yoavkatz (approved):

yoavkatz:
Thanks @yhwang .  The test fails because the submission was from a fork and not branch, so I copied it to a branch in  #1387 and merged.

yhwang:
Thanks!

PR #1391, opened 2024-11-26
Safety updates

bnayahu:
Update the provoq card and the harm llm-as-a-judge safety metric to use WML inference instead of BAM.


prepare/cards/safety/provoq.py
```diff
@@ -18,10 +18,10 @@
             }
         ),
         Rename(field_to_field={"question": "input"}),
-        ListFieldValues(fields=["group", "stigma"], to_field="context"),
-        Join(field="context", by=" # "),
+        ListFieldValues(fields=["group", "stigma"], to_field="contexts"),
+        Dictify(field="contexts", with_keys=["group", "stigma"], to_field="contexts"),
     ],
-    task="tasks.generation.with_context[metrics=[metrics.llm_as_judge.safety.llama_3_70b_instruct_ibm_genai_template_harm_rating]]",
```
yoavkatz, line 24: Just to see I understand.  Contexts are not really used as part of the template (they are not placed in the prompt), so how are they used?

bnayahu: We need the contexts post evaluation (for now). There's a possibility we'll need them in the metric computation as well, but that's FFU.
Frankly, I would suggest to add this to the core TaskCard object, allowing for passing arbitrary fields as part of task_data, regardless of the specific Task, Template or Metric being used.

yoavkatz: I suggest use a modified  input for the task:

task="tasks.generation[input_fields={input=str, type_of_input=str,type_of_output=str,stigma=str,group=str},metrics=[metrics.llm_as_judge.safety.llama_3_70b_instruct_ibm_watsonx_template_harm_rating]"

Then you don't need the new generation_with_contexts task and you also don't need to Dictify and ListFields, etc.

And when you want to score per group/sitga , you can use the group_by recipe parameter (https://www.unitxt.ai/en/latest/docs/evaluating_datasets.html)

load_datasets(cards="cards.provoq", group_by=["stigma","group"])

bnayahu: That was my first version, but there was pushback from @OfirArviv or @elronbandel (sorry - don't remember which one of you helped me with this) as an unlawful hack...

bnayahu: By the way, I originally added generation_with_context (singular) for that reason.

yoavkatz: I can think why @elronbandel would object, but in my experience with users, I think doing "generic" building block for reuse - it not always optimal.  Your use case is so simple that even defining the task within the card from scratch would be simpler .  As in: https://github.com/IBM/unitxt/blob/18788006716f9e2f6efef3fd30f35ee3c136a812/examples/standalone_qa_evaluation.py#L24

bnayahu: That's a bad practice in the other extreme...

yoavkatz: I'm not sure. If someone can read and understand what you are doing in 15 lines of code, instead of going to a catalog task with odd fields like "input_text_type" and "contexts", and then need to apply operators to get it into the right form.

yoavkatz: card = TaskCard(
    # Load the data from the dictionary.  Data can be  also loaded from HF, CSV files, COS and other sources
    # with different loaders.
    loader=....
    # Define the QA task input and output and metrics.
    task=Task(
        input_fields={"question": str},
        reference_fields={},
        prediction_type=str,
        metrics=["metrics.accuracy"],
    ),
    templates=TemplateDict(
       { "default" :  InputOutputTemplate(
                   input_format="{question}",
                    output_format=""}
)

# Create a simple template that formats the input.
# Add lowercase normalization as a post processor.

template = InputOutputTemplate(
    instruction="Answer the following question.",
    input_format="{question}",
    output_format="{answer}",
)

bnayahu: Alright. I reverted to the embedded implementation.

yoavkatz (approved):

PR #1392, opened 2024-11-26
Adding custom processing

antonpibm:
I use unitxt to process the data before running evaluation
At a recent change in unitxt all processors are running as operators under the PostProcess processor, which assumes data fields are either prediction or references. 
This PR adds the capability to apply a processor e.g. `LiteralEval` on any data field by dynamically setting the type to `CustomProcess `


yoavkatz:
I'm not sure  I understand the original problem.   Why do you need run a post processor "to  process the data before running evaluation".   Can you share some example code ?

yoavkatz:
We can discuss in a call. But do you have example code where this is new feature is actually  used, it can give me context.

antonpibm:
After discussing the root cause of the limitation I faced with @yoavkatz a different solution was suggested and submitted in https://github.com/IBM/unitxt/pull/1397

PR #1397, opened 2024-11-27
Add processors also as operators

antonpibm:


prepare/processors/processors.py
```diff
@@ -37,173 +37,175 @@
...
+    """Adds a processor and its associated operator to the catalog.
+
+    Args:
+        artifact_name (str): The name of the artifact to associate with the processor and operator.
+        operator (InstanceFieldOperator): The operator instance to be added.
+        process_references (bool, optional): Whether to process references or not. Defaults to True similar to PostProcess.
+        overwrite (bool, optional): Whether to overwrite an existing entry with the same artifact name. Defaults to True.
+    """
+    add_to_catalog(
+        PostProcess(operator, process_references=process_references),
```
yoavkatz, line 56: Maybe for clariity in the catalog, it's worthwhile to add process_predictions=True, so both will appear in the catalog json.
```suggestion
        PostProcess(operator, process_references=process_references,process_prediction=True),
```

yoavkatz (approved):
Approve. One minor change requested. See below.

PR #1401, opened 2024-11-28
Reduce error message clutter

yoavkatz:
Up to now, exceptions texts were concatinated to create long (unreadble) error message.  This change makes sure each level adds only new information and refers to the causing exceptio

### DIFF
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -190,7 +190,7 @@ class MapInstanceValues(InstanceOperator):
             if value is not None:
                 if (self.process_every_value is True) and (not isinstance(value, list)):
                     raise ValueError(
-                        f"'process_every_field' == True is allowed only when all fields which have mappers, i.e., {list(self.mappers.keys())} are lists. Instance = {instance}"
+                        f"'process_every_field' == True is allowed only when for fields whose values are lists, but value of field '{key}' is '{value}'"
                     )
                 if isinstance(value, list) and self.process_every_value:
                     for i, val in enumerate(value):
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -454,7 +454,7 @@ class InstanceFieldOperator(InstanceOperator):
                     old_value = self.get_default
             except Exception as e:
                 raise ValueError(
-                    f"Failed to get '{from_field}' from {instance} due to : {e}"
+                    f"Failed to get '{from_field}' from {instance} due to the exception above."
                 ) from e
             try:
                 if self.process_every_value:
@@ -466,7 +466,7 @@ class InstanceFieldOperator(InstanceOperator):
                     new_value = self.process_instance_value(old_value, instance)
             except Exception as e:
                 raise ValueError(
-                    f"Failed to process '{from_field}' from {instance} due to : {e}"
+                    f"Failed to process '{from_field}' from {instance} due to the exception above."
                 ) from e
             dict_set(
                 instance,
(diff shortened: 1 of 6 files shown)


yoavkatz:
Hi @dafnapension - Can you please review this?

src/unitxt/operators.py
```diff
@@ -190,7 +190,7 @@ def process(
             if value is not None:
                 if (self.process_every_value is True) and (not isinstance(value, list)):
                     raise ValueError(
-                        f"'process_every_field' == True is allowed only when all fields which have mappers, i.e., {list(self.mappers.keys())} are lists. Instance = {instance}"
+                        f"'process_every_field' == True is allowed only when for fields whose values are lists, but value of field '{key}' is '{value}'"
```
dafnapension, line 193: suggesting: 
f"'process_every_field' == True is allowed only for fields whose values are lists, but value of field '{key}' is '{value}'"
(eliminating the word `when`)

src/unitxt/operators.py
```diff
@@ -977,7 +977,7 @@ def process(
             if self.process_every_value:
                 assert isinstance(
                     value, list
-                ), f"'process_every_value' can be set to True only for fields that contain lists, whereas in instance {instance}, the contents of field '{field_name}' is of type '{type(value)}'"
+                ), f"'process_every_value' can be set to True only for fields that contain lists, the contents of field '{field_name}' is of type '{type(value)}'"
```
dafnapension, line 980: suggesting to use same phrasing as above in this case of `process_every_value`:
f"'process_every_field' == True is allowed only for fields whose values are lists, but value of field '{field_name}' is '{value}'"

src/unitxt/operators.py
```diff
@@ -211,7 +211,7 @@ def get_mapped_value(self, instance, key, mapper, val):
             return recursive_copy(mapper[val_as_str])
         if self.strict:
             raise KeyError(
-                f"value '{val}' in instance '{instance}' is not found in mapper '{mapper}', associated with field '{key}'."
+                f"value '{val}' in field '{key}' is not found in mapper '{mapper}'"
```
dafnapension, line 214: `val_as_str` is actually the one missing from the mapper. I suggest to hint about this, in case users will use mappers from non-str (although we warn against this):

suggesting:  f"'{val_as_str}', the casted-to-string value from field '{key}', is not found in mapper '{mapper}'"

src/unitxt/operators.py
```diff
@@ -1194,13 +1194,13 @@ def _is_required(self, instance: dict) -> bool:
                 instance_key = dict_get(instance, key)
             except ValueError as ve:
                 raise ValueError(
-                    f"Required filter field ('{key}') in FilterByCondition is not found in {instance}"
+                    f"Required filter field ('{key}') in FilterByCondition is not found in {instance.keys()} of instance"
```
dafnapension, line 1197: It seems that class `FilterByConditionBasedOnFields`, as its name suggests, refers to instance fields only (and not deeper subfields).
If we want to extend the meaning to subfields as well (which I recommend, actually, why not?) then we need to change the above message to:
 f"Required filter field ('{key}') in FilterByCondition is not found in instance" .
If not, then I think we need to replace both `dict_get`-s (`dict_get(instance, key)` and `dict_get(instance, value)`) by a simple, direct lookup:  `instance[key]` and `instance[value]`, and build the error accordingly.

src/unitxt/operators.py
```diff
@@ -1194,13 +1194,13 @@ def _is_required(self, instance: dict) -> bool:
...
                 ) from ve
             try:
                 instance_value = dict_get(instance, value)
             except ValueError as ve:
                 raise ValueError(
-                    f"Required filter field ('{value}') in FilterByCondition is not found in {instance}"
+                    f"Required filter field ('{value}') in FilterByCondition is not found in {instance.keys()} of instance"
```
dafnapension, line 1203: please see same comment a few lines above

yoavkatz:
Thanks @dafnapension . I updated per your suggestions.

Issue #1400, opened 2024-11-28
Strange Behavior with Copy


yoavkatz:
If we copy a string to a nested field in another string field (which is meaningless) , we get this odd behavior that the target string field becomes a dictionary, with the copied value.
 
```
def test_copy_string_to_string_error(self):
        inputs = [
            {"source": "hello", "task_data" :3}
        ]

        targets = [{"source": "hello","task_data": {'source': 'hello'}}]

        check_operator(
            operator=Copy(field="source", to_field="task_data/source"),  <-- this should fail because task_data does not have a 'source field' . Instead it works.
            inputs=inputs,
            targets=targets,
            tester=self,
        )
```

dafnapension:
Hi @yoavkatz , it is not odd.. it is as planned, pending the flag `not_exist_ok` which is set to True, by default, for `dict_set`, which, in turn, participates in the last stage of `Copy`. It means that if `dict_set` needs to set something, and not all the way to that something exists, then dict_set instantiates all the non-existing:

```python
from unitxt.dict_utils import dict_set

dic = {"source": "hello", "task_data" :3}
try:
    dict_set(dic, "task_data/source", "hello", not_exist_ok=False) 
    print(dic)
except Exception as e:
    print(e)  # prints No path in dic {'source': 'hello', 'task_data': 3} matches query task_data/source
    
dic = {"source": "hello", "task_data" :3}

try:
    dict_set(dic, "task_data/source", "hello", not_exist_ok=True) 
    print(dic) # prints  {'source': 'hello', 'task_data': {'source': 'hello'}}
except Exception as e:
    print(e)  
```

dafnapension:
A solution could be adding a flag, `not_exist_ok`, to `Copy`,  which `Copy` will forward to `dict_set`.
Will that do?

dafnapension:
Here is how `dict_set` participates now in the last stage of `Copy`, actually: in any `InstanceFieldOperator`:
```python
            dict_set(
                instance,
                to_field,
                new_value,
                not_exist_ok=True,   # << note here
            )
  ```
  
We may want to change it to use the flag `not_exist_ok` of `InstanceFieldOperator`.
The same flag is also used for getting the value in the first stage of `InstanceFieldOperator`
So either we use the same flag for both missions, or add another flag and separate.
What would you suggest?

yoavkatz:
[...]

PR #1402, opened 2024-11-29
not_exist_ok for dict_set only allows to add a non existing key to a dictionary, or non existing entry to a list. Not a whole nested structure implied by the query

dafnapension:
closes: https://github.com/IBM/unitxt/issues/1400

Can still set a whole nested structure, but as the value being set, not through processing the query by dict_set.

### DIFF
--- a/prepare/cards/mmmu.py
+++ b/prepare/cards/mmmu.py
@@ -1,7 +1,7 @@
 from unitxt.blocks import LoadHF, TaskCard
 from unitxt.catalog import add_to_catalog
 from unitxt.collections_operators import Filter
-from unitxt.operators import ListFieldValues, MapValues
+from unitxt.operators import ListFieldValues, MapValues, SetEmptyDictIfDoesNotExist
 from unitxt.processors import LiteralEval, Lower
 from unitxt.splitters import RenameSplits
 from unitxt.string_operators import MapReplace
@@ -47,6 +47,7 @@ for name in config_names:
         ),
         preprocess_steps=[
             RenameSplits(mapper={"dev": "train", "validation": "test"}),
+            SetEmptyDictIfDoesNotExist(field="media"),
             ListFieldValues(
                 fields=[f"image_{i}" for i in range(1, 8)], to_field="media/images"
             ),
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -39,6 +39,7 @@ General Operators List:
 ------------------------
 """
 
+import json
 import operator
 import uuid
 import warnings
@@ -285,6 +286,9 @@ class SetEmptyDictIfDoesNotExist(InstanceOperator):
     ) -> Dict[str, Any]:
         if self.field not in instance:
             instance[self.field] = {}
+        if not isinstance(instance[self.field], dict):
+            assert isinstance(instance[self.field], str)
+            instance[self.field] = json.loads(instance[self.field])
         return instance
 
 
--- a/src/unitxt/operators.py
+++ b/src/unitxt/operators.py
@@ -277,6 +277,17 @@ class Set(InstanceOperator):
         return instance
 
 
+class SetEmptyDictIfDoesNotExist(InstanceOperator):
+    field: str
+
+    def process(
+        self, instance: Dict[str, Any], stream_name: Optional[str] = None
+    ) -> Dict[str, Any]:
+        if self.field not in instance:
+            instance[self.field] = {}
+        return instance
+
+
 @deprecation(version="2.0.0", alternative=Set)
 class AddFields(Set):
     pass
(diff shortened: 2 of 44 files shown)


dafnapension:
I think these are rather ugly patches.. and I did not manage to cover all. 
I suggest to return back home..

prepare/cards/mmmu.py
```diff
@@ -47,6 +47,7 @@
         ),
         preprocess_steps=[
             RenameSplits(mapper={"dev": "train", "validation": "test"}),
+            SetEmptyDictIfDoesNotExist(field="media"),
```
yoavkatz, line 50: I think the media field is used by generic code,  so it should be added by the code and not the cards.

src/unitxt/operators.py
```diff
@@ -277,6 +278,20 @@ def process(
         return instance
 
 
+class SetEmptyDictIfDoesNotExist(InstanceOperator):
```
yoavkatz, line 281: The behavior of this is method is more than expected by the name.  It not only creates an empty dict if it does not exist, but also transform the field from string to dict if string is a json.  I think we should avoid such complex behavior, which is more indicative of a bug somewhere else. 

For example, on Friday I found a bug in post processing pipeline:

image

(Still not in main)

dafnapension: Hi @yoavkatz , If this helps, I found what triggered my need to add this loads to `SetEmptyDictIfDoesNotExist` (and again, just for patching with גרזנים ולומיק the need to prepare container elements, to not surprise the user):

This method, shamelessly makes `task_data` into a string:
```python
    def serialize_instance_fields(self, instance, task_data):
        if settings.task_data_as_text:
            instance["task_data"] = json.dumps(task_data)

        if not isinstance(instance["source"], str):
            instance["source"] = json.dumps(instance["source"])
        return instance
```

dafnapension:
Hi @yoavkatz , I suspect that the new `SetEmptyDictIfDoesNotExist` that I added, and the about to be added `SetEmptyListIfDoesNotExist` and the enforcement to use them before each and every existing `Set` and more over: each and every existing Operator that writes anything to any `to_field`, will drive our users crazy. Let alone the runtime it will consume.
I suggest to consider any of the following solutions:
(1) extend the possible values of `not_exist_ok` to : No, at_most_last_component, any_missing_component, 
and throw exception accordingly  (my recommendation)
(2) only allow at_most_last_component (that you feel is the decent, expected behavior), and then once an operator and an input instance like those that started the current saga will meet:
applying operator=Copy(field="source", to_field="task_data/source") on input instance
{"source": "hello", "task_data" :3} will throw a Warning, saying something like:
operator Copy employed dict_set in order to set value "hello" to an element (subfield) "task_data/source" that does not exist in the input instance, nor its implied containing subfield of type dict "task_data". (which exists, but is not a dictionary). dict_set is not allowed to generate into the instance elements implied by the query, other than the element implied by the last component. 
dict_set was invoked instead with the following query and value:  dict_set("task_data", {"source" : "hello"}) which sets a value, of type dict, to an existing element, "task_data", replacing its old value.

Other than these explanation and apologies, the computed results will remain the same, users will not have to change their existing code, and/or write tedious future invocations of dict_set s.

yoavkatz:
> Hi @yoavkatz , I suspect that the new `SetEmptyDictIfDoesNotExist` that I added, and the about to be added `SetEmptyListIfDoesNotExist` and the enforcement to use them before each and every existing `Set` and more over: each and every existing Operator that writes anything to any `to_field`, will drive our users crazy. Let alone the runtime it will consume. I suggest to consider any of the following solutions: (1) extend the possible values of `not_exist_ok` to : No, at_most_last_component, any_missing_component, and throw exception accordingly (my recommendation) (2) only allow at_most_last_component (that you feel is the decent, expected behavior), and then once an operator and an input instance like those that started the current saga will meet: applying operator=Copy(field="source", to_field="task_data/source") on input instance {"source": "hello", "task_data" :3} will throw a Warning, saying something like: operator Copy employed dict_set in order to set value "hello" to an element (subfield) "task_data/source" that does not exist in the input instance, nor its implied containing subfield of type dict "task_data". (which exists, but is not a dictionary). dict_set is not allowed to generate into the instance elements implied by the query, other than the element implied by the last component. dict_set was invoked instead with the following query and value: dict_set("task_data", {"source" : "hello"}) which sets a value, of type dict, to an existing element, "task_data", replacing its old value. This seems to achieve the requested result, and still comply with the rule that dict_set changes at most the element identified by the last component of the query.
> 
> Other than these explanation and apologies, the computed results will remain the same, users will not have to change their existing code, and/or write tedious future invocations of dict_set s.

Hi. I guess what I'm missing is how prevalent is this problem and wherher its actually a problem caused by another issue.. I means media and task data are generic fields that  should always be there and added by generic cide. If not, we need to understand why.  

In general, we need to strongly avoid complex flags, that increase the learning curve of the library.

dafnapension:
I agree (although I like flags..).   Can [this issue](https://github.com/IBM/unitxt/pull/1402#discussion_r1865602637) hint to what you are looking for? I bumped into it while needing to push this LoadJson into the patch.

dafnapension:
Too much existing code trust dict_set to fill in all missing component, and not just the last component

PR #1408, opened 2024-12-03
Adjust binary llmaj to new engines and add rits support

lilacheden:


yoavkatz (approved):

src/unitxt/llm_as_judge.py
```diff
@@ -65,12 +65,12 @@ def verify(self):
             )
 
         if isinstance(self.inference_model, OpenAiInferenceEngine):
-            if self.format and type(self.format) is not SystemFormat:
-                raise ValueError(
-                    "Error in 'LLMAsJudge' metric. Inference model 'OpenAiInferenceEngine' does "
-                    "not support formatting. Please remove the format definition from the recipe"
-                    " (OpenAi Chat API take care of the formatting automatically)."
-                )
+            # if self.format and type(self.format) is not SystemFormat:
```
OfirArviv, line 68: change it to be either None, formats.empty or formats.chat_api

OfirArviv (changes requested):

src/unitxt/llm_as_judge.py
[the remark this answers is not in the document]
yoavkatz: Except if it's empty format , which is ok.

lilacheden: @OfirArviv - the format is the artifact, not the string. How can I check the string from the artifacrt?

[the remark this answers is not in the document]
lilacheden: @yoavkatz - the current check was an error anyway as it allows for any format besides chat_api. see the above discussion with Ofir

lilacheden: @OfirArviv - to be more clear, I can check if the format is a chat api format instance, but don't know how to check if a system format is generated from "formats.empty" - this is probably where this bug emerged from in the first place

OfirArviv (approved):

PR #1409, opened 2024-12-03
Eval Assist integration

martinscooper:
This PR includes all the changes that add EvalAssist-type LLM a Judge evaluators.


src/unitxt/templates.py
```diff
@@ -528,6 +537,17 @@ def prepare(self):
                 "XX",
             ]
 
+    def verify(self):
+        super().verify()
+        print(self.postprocessors[-1])
```
yoavkatz, line 542: I think these checks can be removed. processors.match_closest_option post processor is not a requirement of the MultipleChoiceTemplate.

src/unitxt/processors.py
```diff
@@ -175,6 +175,7 @@ def process_value(self, text: Any) -> Any:
 
 
 class MatchClosestOption(InstanceFieldOperator):
+    __description__ :  str = "Note: this post processor only works with MultipleChoiceTemplate"
```
yoavkatz, line 178: I think this can be removed.  This should work with any task with options in it's task data.

src/unitxt/api.py
```diff
@@ -170,9 +174,17 @@ def infer(
     return_data: bool = False,
     return_log_probs: bool = False,
     return_meta_data: bool = False,
+    previous_messages: Optional[list[dict[str, str]]] = None,
```
yoavkatz, line 177: Hi. I tend to think that adding previous messages is rather specific (and rare) behavior which is beyond the scope of a general infer() method, instead I would inline this code where it is need.

Otherwise, we need to document it, check for cases where the previous messages or current "source"f files are a string and not list, etc...

src/unitxt/catalog/metrics/llm_as_judge/eval_assist/direct_assessment/criterias/answer_relevance.json
```diff
@@ -0,0 +1,33 @@
...
+        },
+        {
+            "__type__": "criteria_option",
+            "name": "Acceptable",
+            "description": "The response is adequate but could be better."
+        },
+        {
+            "__type__": "criteria_option",
+            "name": "Could be Improved",
+            "description": "The response relates to the questions but does not directly answer it."
```
yoavkatz, line 19:
```suggestion
            "description": "The response relates to the question but does not directly answer it."
```

examples/evaluate_eval_assist_direct_predefined_criteria.py
```diff
@@ -0,0 +1,52 @@
...
+    "test": [
+        {"context": {"Question": "How is the weather?"}},
+        {"context": {"Question": "How is the weather?"}},
+        {"context": {"Question": "How is the weather?"}},
+    ]
+}
+
+criteria = "metrics.llm_as_judge.eval_assist.direct_assessment.criterias.temperature"
+metrics = [
+    f"metrics.llm_as_judge.eval_assist.direct_assessment.rits.llama3_1_70b[criteria_or_criterias={criteria}]"
```
yoavkatz, line 19: I think "criteria_or_criterias", should be called "criteria".

examples/evaluate_existing_dataset_by_llm_as_judge_eval_assist.py
```diff
@@ -0,0 +1,49 @@
...
+logger = get_logger()
+settings = get_settings()
+
+with settings.context(allow_unverified_code=True):
+    # Use the HF load_dataset API, to load the squad QA dataset using the standard template in the catalog.
+    # We set loader_limit to 20 to reduce download time.
+    dataset = load_dataset(
+        card="cards.squad",
+        metrics=[
+            "metrics.llm_as_judge.eval_assist.direct_assessment.rits.llama3_1_70b[context_field=context,criteria_or_criterias=metrics.llm_as_judge.eval_assist.direct_assessment.criterias.answer_relevance,score_prefix=answer_relevance_]"
```
yoavkatz, line 17: I think the context_field should be called "context_fields" and be a list, because you may want to pass both "question" and "context" fields.

yoavkatz:
You mentioned in the past, that in some datasets, the criteria is not for all the dataset, but rather each instance has a different criteria.  

We also just saw an important use case in granite feedback program, where the datasets has for each instance
1) question
2) gold answer
3) judgment criteria

Judgement criteria looks like this for one instance:

* Core Point: The model's output should clearly state that the doctor is the boy's mother and female.
* Avoid Irrelevant Information: The output should not include speculative or extraneous information, such as the doctor practicing medicine under a female name due to gender discrimination or personal preference.

And this for another instance:

* The text should contain one question about the user's motivation for attending the university, helpful for assessing their fit for the academic program.
* The ground truth text is just a reference. The exact wordings in model's output could be slightly different. 

I think we need to make this use case easily implementable.

For example, one could either pass a "criteria" to the judge , where the criteria will be taken from the catalog and "criteria_field", where the criteria will be taken from a text field.

src/unitxt/api.py
martinscooper: I think that, as unitxt is tending to primarily use OpenAI chat format, having previous messages won't be that rare of a use case. Anytime the user needs to simulate a conversation with the assistant, like in a chat or CoT, this parameter will be required. 
The verification checks to be done if `previous_messages` is not `None` are:
- `previous_messages` should be a list of messages.
-  the source of the dataset's instances should be a list of messages too. 
- we can go ahead and check that `previous messages` are valid openai chat format too, i.e. have a `source` and a `content` attributes of type string. If format is not `formats.chat_api` we should check that the source of the dataset's instances are valid message lists too.

If you still think this is not the appropriate place to set the previous messages, could you suggest another place?

martinscooper:
> You mentioned in the past, that in some datasets, the criteria is not for all the dataset, but rather each instance has a different criteria.
> 
> We also just saw an important use case in granite feedback program, where the datasets has for each instance
> 
> 1. question
> 2. gold answer
> 3. judgment criteria
> 
> Judgement criteria looks like this for one instance:
> 
> * Core Point: The model's output should clearly state that the doctor is the boy's mother and female.
> * Avoid Irrelevant Information: The output should not include speculative or extraneous information, such as the doctor practicing medicine under a female name due to gender discrimination or personal preference.
> 
> And this for another instance:
> 
> * The text should contain one question about the user's motivation for attending the university, helpful for assessing their fit for the academic program.
> * The ground truth text is just a reference. The exact wordings in model's output could be slightly different.
> 
> I think we need to make this use case easily implementable. I see there is already support , such that if the metric has no criteria specified as a parameter, it tries to load it from the instance "criteria" field. The example:
> 
> examples/evaluate_eval_assist_pairwise_user_criteria.py
> 
> does this but actually has only one fixed criteria.
> 
> We should have an example to show how to do , also for direct metrics.
> 
> So we have 6 usecases (1) Direct / Pair-wise (2) Single criteria from catalog / single user defined criteria / dynamic criteria defined per instance in the dataset

I started working on this. You can see [this example](https://github.com/IBM/unitxt/blob/eval-assist-integration/examples/evaluate_eval_assist_direct_several_criterias.py) where two different criterias are used. I also added the flexibility to specify criterias from the catalog (a `string`) or as a json (a `dict`). @yoavkatz do you think it would be better to express the criterias a `CriteriaWithOptions` instead of `dict`?

yoavkatz:
> > You mentioned in the past, that in some datasets, the criteria is not for all the dataset, but rather each instance has a different criteria.
> > We also just saw an important use case in granite feedback program, where the datasets has for each instance
> > 
> > 1. question
> > 2. gold answer
> > 3. judgment criteria
> > 
> > Judgement criteria looks like this for one instance:
> > 
> > * Core Point: The model's output should clearly state that the doctor is the boy's mother and female.
> > * Avoid Irrelevant Information: The output should not include speculative or extraneous information, such as the doctor practicing medicine under a female name due to gender discrimination or personal preference.
> > 
> > And this for another instance:
> > 
> > * The text should contain one question about the user's motivation for attending the university, helpful for assessing their fit for the academic program.
> > * The ground truth text is just a reference. The exact wordings in model's output could be slightly different.
> > 
> > I think we need to make this use case easily implementable. I see there is already support , such that if the metric has no criteria specified as a parameter, it tries to load it from the instance "criteria" field. The example:
> > examples/evaluate_eval_assist_pairwise_user_criteria.py
> > does this but actually has only one fixed criteria.
> > We should have an example to show how to do , also for direct metrics.
> > So we have 6 usecases (1) Direct / Pair-wise (2) Single criteria from catalog / single user defined criteria / dynamic criteria defined per instance in the dataset
> 
> I started working on this. You can see [this example](https://github.com/IBM/unitxt/blob/eval-assist-integration/examples/evaluate_eval_assist_direct_several_criterias.py) where two different criterias are used. I also added the flexibility to specify criterias from the catalog (a `string`) or as a json (a `dict`). @yoavkatz do you think it would be better to express the criterias a `CriteriaWithOptions` instead of `dict`?

Thanks. I think in this case, when the data usually comes from another source (e.g. a CSV file) the criteria should be a json.  Moreover, in a more common use case we should just provide the textual criteria , where the user provided a column with with what to judge

 "If there is a numerical temperature present, it should be dominated in both Fahrenheit and Celsius".
 "There should be no names in the text"
   
  And build automatically  criteria that returns Yes/No (mapped to 1.0 / 0.0)  on where the response meets the conditions.
  Could also be on "Full / Partially/ No" or another scale.

src/unitxt/api.py
yoavkatz: Hi. I see your points.  I'll be glad to hear @elronbandel .  Maybe given the shift to chat, it is motivated.

martinscooper:
> Thanks. I think in this case, when the data usually comes from another source (e.g. a CSV file) the criteria should be a json. Moreover, in a more common use case we should just provide the textual criteria , where the user provided a column with with what to judge
> 
> "If there is a numerical temperature present, it should be dominated in both Fahrenheit and Celsius". "There should be no names in the text"
> 
> And build automatically criteria that returns Yes/No (mapped to 1.0 / 0.0) on where the response meets the conditions. Could also be on "Full / Partially/ No" or another scale.

I like this suggestion. I'll implement it.

src/unitxt/eval_assist_llm_as_judge_direct.py
```diff
@@ -96,12 +96,28 @@ def get_criterias(self, task_data, eval_count):
                     except:
                         raise Exception(
                             "The criteria dict couldn't be parsed correctly. It has to have the json representation of CriteriaWithOptions"
-                        )
+                        ) from None
                 elif isinstance(task_data_criteria, str):
-                    criteria = fetch_artifact(task_data_criteria)[0]
+                    try:
+                        criteria = fetch_artifact(task_data_criteria)[0]
+                    except Exception:
```
yoavkatz, line 103: This seems a bit error prone.  If I mistype the name of a catalog asset, it will become a textual criteria and get weird results.  It will also have performance issues, searching for the artificat.  
I think this option need to be more explicit.

yoavkatz: For example, have different field names for each use case.

yoavkatz: I have a better suggestion. Keep the code as is, but add a field operator that takes a string and convert it into a dict. So if the raw data has a string creteria it will convert it.  There could also  be multiple diffeentoperator.  One that conerys to Yes/No. Another that converts to Yes/no/partially.

martinscooper: I didn't fully understand your proposed field operator? Can you write some pseudo-code for the `process_value` method?

Just to be in the same page, these are the source and types of criterias we could potentially accept in a direct assessment evaluation. We can narrow the list.
- If it is a single criteria it can be set in the constructor:
  - a catalog criteria string if the evaluator is specified as a catalog string (square brackets). Still to check if a dict representation would work here.
  - any criteria representation if the evaluator is created using code (not from catalog), like a catalog criteria string or a `CriteriaWithOptions` object.
- If multiple criterias, they have to be specified in the `criteria` field of the `input_fields`. It can be:
  - a string representing a criteria in the catalog,
  - a string representing the description of the criteria with a default set of `options` and `option_map` with empty respective description (e.g. Yes/No, Yes/No/Partially)
  - a string of the json representation of the criteria, e.g `"{'description': '', 'options': [{'...'}, {'...'}, {'...'}], 'option_map':{...}}"`
  - a dict: the json representation of the criteria
  - A `CriteriaWithOptions` object, no parsing needed.

yoavkatz: I suggest there will be different operators to convert the input in the original raw data field to the Critera.

class LoadCriteria(FieldOperator):
    def process_value(self, text: Any) -> Any:
         return   fetch_artifact(task_data_criteria)[0]
        
class CreateCriteriaFromDict(FieldOperator):
     def process_value(self, text: Any) -> Any:
         criteria = CriteriaWithOptions(
                             name=task_data_criteria["name"],
                             description=text["description"],
                             options=[
                                 CriteriaOption(
                                     name=option_dict["name"],
                                     description=option_dict["description"],
                                 )
                                 for option_dict in text["options"]
                             ],
           return criteria                  
           
class CreateCriteriaFromString(FieldOperator):
     def process_value(self, text: Any) -> Any:
            criteria = CriteriaWithOptions(
                             name=f"Unknown ({text[:20]}...)",
                             description=text,
                             options=[
                                 CriteriaOption(name="Yes", description=""),
                                 CriteriaOption(name="No", description=""),
                             ],
                             option_map={
                                 "Yes": 1.0,
                                 "No": 0.0,
                             },
                         )
           return criteria
           
Then:
    "test": [
         {"question": "How is the weather?"},
         {"judgement": "Response should discussthe weather, and only the weather")}
     ]
card = TaskCard(
     loader=LoadFromDictionary(data=data, data_classification_policy=["public"]),
     preprocess_steps=[
                                    CreateCriteriaFromString(from_field="judgement", to_field="criteria"),
                                    
                                    ],
     task=Task(
         input_fields={"question": str, "criteria": Criteria},
         reference_fields={},
         prediction_type=str,
         metrics=[
             "metrics.llm_as_judge.eval_assist.direct_assessment.rits.llama3_1_70b[context_fields=question]"
         ],
     ),
 )

src/unitxt/catalog/metrics/llm_as_judge/eval_assist/direct_assessment/criterias/temperature.json
```diff
@@ -0,0 +1,27 @@
+{
```
yoavkatz, line 1: I would rename it as "temperature_in_both_celsius_and_fahrentheit" , so it's clear from the name what is the criteria.

src/unitxt/api.py
martinscooper: Can somehow this be achieved by a formatter? Like `ChatAPIFormatWithPreviousMessages`.

src/unitxt/eval_assist_llm_as_judge_direct.py
yoavkatz: We are checking if this can work - I'll send an example if it does.

yoavkatz: I've implemented the above.

See examples/evaluate_eval_assist_direct_criteria_from_dataset.py

It seems to work, but I think it exposed another bug. There returned score is '1' although the judge selected the 'No' options:

score:
    instance:
        score (int):
            1
        llm_as_a_judge_score (int):
            1
        positional_bias (bool):
            False
        selected_option (str):
            No
        positional_bias_selected_option (str):
            No
        assessment (str):
            No
            
            The response is not funny as it's a weak play on words, and the punchline "To cat to the other side" is a
            n obvious and unoriginal variation of the classic "To get to the other side" phrase.
        positional_bias_assessment (str):
            No

martinscooper: Thanks for contributing this @yoavkatz . The bug is related to the fact that the `option_map` attribute was not being correctly set and the default value for the `option_map` was 1. Do you think the `option_map` should be required? It is currently optional. For simple usage, it makes sense to keep it as optional, we could set default float values for the options.

The criteria in the `task_data` objects is currently a dict (with `__type__` key) although a `CriteriaWithOptions` object is returned by `CreateYesNoCriteriaFromString`. So, we have to call `fetch_artifact()`. Is this expected?

yoavkatz: > For simple usage, it makes sense to keep it as optional, we could set default float values for the options.

How could we give defaults to option_map? Don't we need to set it always to convert it.

I will check regarding desearlizing the object in the task dat - for now - try to fetch_artificat().

martinscooper: To give defaults to `option_map` we could assume that the options are an ordinal variable in crescending order and assign the value of `i/n` to the _ith_ option, where `n` is the number of options. If we make the `option_map` required, which values should a user set to a nominal categorical variable, where options aren't ordered?

yoavkatz: I see your point.  I think that if the options are   "Happy" , "Sad", "Angry" we create should scores like "happy_score" and have it 1, if the chosen category is "happy" or else otherwise.    This will allow to aggregate over these score, over all instances (if we use the same criteria).

Now the question what will be the default "score" field.  If option_map is not set, then it could be assign the value of i/n to the ith option, where n is the number of options.   If option_map  is set, it should be used to map.

examples/evaluate_eval_assist_direct_several_criterias.py
```diff
@@ -0,0 +1,73 @@
+from typing import Any
```
yoavkatz, line 1: Currently support criteria from dataset using Operator.  So this example can be removed, but add different way to load the criteria.

examples/evaluate_eval_assist_direct_user_criteria_no_catalog.py
```diff
@@ -0,0 +1,68 @@
+from unitxt import get_logger
+from unitxt.api import create_dataset, evaluate
+from unitxt.eval_assist_constants import (
```
yoavkatz, line 3: from unitxt.eval_assist import Criteria,EvalAssistLLMsJudgeDirect

yoavkatz: llm_as_judge_from_criteria - suggest rename

examples/evaluate_eval_assist_direct_user_criteria_no_catalog.py
```diff
@@ -0,0 +1,68 @@
...
+    CriteriaWithOptions,
+    EvaluatorNameEnum,
+)
+from unitxt.eval_assist_llm_as_judge_direct import EvalAssistLLMAsJudgeDirect
+from unitxt.inference import LiteLLMInferenceEngine
+from unitxt.text_utils import print_dict
+
+logger = get_logger()
+
+criteria = CriteriaWithOptions(
```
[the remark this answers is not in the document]
martinscooper: I ended up implementing `from_jsons()`(calls `from_obj`) and `from_obj()`(receives a dict), which are used by the operators.

.pre-commit-config.yaml
```diff
@@ -8,7 +8,12 @@ repos:
       # Run the linter on all files except the specific one
       - id: ruff
         args: [--fix]
-        exclude: src/unitxt/metrics.py|examples/evaluate_existing_dataset_no_install.py
+        exclude: src/unitxt/metrics.py|examples/evaluate_existing_dataset_no_install.py|src/unitxt/eval_assist_llm_as_judge_pairwise.py
+      # Run the linter on the specific file with the ignore flag
+      - id: ruff
+        name: ruff (src/unitxt/eval_assist_llm_as_judge_pairwise.py)
+        files: src/unitxt/eval_assist_llm_as_judge_pairwise.py
+        args: [--fix, --ignore, C901]
```
elronbandel, line 16: This changes should be moved to the pyproject.toml

elronbandel (changes requested):

examples/evaluate_eval_assist_direct_user_criteria_no_catalog.py
```diff
@@ -0,0 +1,76 @@
...
+            "description": "The temperature reading is provided either in Fahrenheit or Celsius, but not both."
+        },
+        {
+            "name": "Pass",
+            "description": "There is no numerical temperature reading in the response."
+        }
+    ],
+    "option_map": {"Yes": 1.0, "No": 0.5, "Pass": 0.0}
+}
+"""
```
elronbandel, line 34: can we have this working by a `from_dict` function? it seems like from json is not suitable in this case and is unnatural python code

examples/evaluate_eval_assist_pairwise_predefined_criteria.py
```diff
@@ -0,0 +1,68 @@
...
+            "question": "Tell me a joke about cats",
+            "criteria": "metrics.llm_as_judge.eval_assist.pairwise_comparison.criterias.funny_joke",
+        },
+    ]
+}
+
+card = TaskCard(
+    loader=LoadFromDictionary(data=data, data_classification_policy=["public"]),
+    preprocess_steps=[
+        LoadCriteria(field="criteria", to_field="criteria"),
```
elronbandel, line 28:
```suggestion
        LoadCriteria(field="criteria"),
```
equivalent

examples/evaluate_eval_assist_direct_user_criteria_no_catalog.py
martinscooper: There is a `from_obj()` class method that returns a `CriteriaWithOptions` object from a dict. I didn't use the `from_dict()` method because it needs the dict to have a `__type__` attribute. If not, it throws:
`unitxt.artifact.MissingArtifactTypeError: Missing '__type__' parameter`.

elronbandel (approved):
Great work @martinscooper !

PR #1411, opened 2024-12-04
Fix model strings for Llama 3 on Together AI

yifanmai:
The current model strings for Llama 3 are incorrect. This updates the model strings to match the current [Together AI documentation](https://docs.together.ai/docs/serverless-models).

### DIFF
--- a/src/unitxt/inference.py
+++ b/src/unitxt/inference.py
@@ -2739,8 +2739,8 @@ class CrossProviderInferenceEngine(InferenceEngine, StandardAPIParamsMixin):
             "granite-3-8b-instruct": "ibm/granite-3-8b-instruct",
         },
         "together-ai": {
-            "llama-3-8b-instruct": "together_ai/togethercomputer/llama-3-8b-instruct",
-            "llama-3-70b-instruct": "together_ai/togethercomputer/llama-3-70b-instruct",
+            "llama-3-8b-instruct": "together_ai/meta-llama/Llama-3-8b-chat-hf",
+            "llama-3-70b-instruct": "together_ai/meta-llama/Llama-3-70b-chat-hf",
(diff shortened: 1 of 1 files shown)


yifanmai:
Test failures seem to be due to `huggingface-cli login --token` failing and don't seem related to this PR.

yoavkatz (approved):

yoavkatz:
Yes. We don't accept merges from forks easily.  Do you know how do you configure GitHub to allow running action on forks (after review)?

yifanmai:
> Do you know how do you configure GitHub to allow running action on forks (after review)?

Unfortunately no. GitHub just doesn't allow using secrets with actions triggered by forks for security reasons. I ran into the same issue with HELM as well, and didn't have a good solution. I ended up disabling pull request actions that need secrets.

[Source](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions):

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

yoavkatz:
> > Do you know how do you configure GitHub to allow running action on forks (after review)?
> 
> Unfortunately no. GitHub just doesn't allow using secrets with actions triggered by forks for security reasons. I ran into the same issue with HELM as well, and didn't have a good solution. I ended up disabling pull request actions that need secrets.
> 
> [Source](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions):
> 
> > With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

Thanks.  I did not find that before, and  gpt4 hallucinated that that it is possible, but with instructions I could not recreate.  

Good to know.

PR #1437, opened 2024-12-14
metadata is set only once in recipe

dafnapension:


elronbandel (commented):
I remember there was a reason for that (unfortunately no test for that as It seems).

elronbandel (approved):
Great catch thanks!

PR #1441, opened 2024-12-16
Add parallelization to RITS inference

arielge:
Adds parallelization from the client side. Could probably be improved further with multiprocessing but this already provides a very significant speedup


elronbandel (approved):
This can definitely work, though not necessarily ensure you dont hit the limit. For now it's good, in the future it is better to use the implementation we have for LightLLM where the below the limit is gurenteed.

PR #1445, opened 2024-12-19
try see if TestExample is triggered

dafnapension:


dafnapension:
Passed all tests, including Example Test alright.

PR #1459, opened 2024-12-26
Add MapReduceMetric a new base class to integrate all metrics into

elronbandel:


src/unitxt/api.py
```diff
@@ -198,8 +198,12 @@ def load_dataset(
     ).with_transform(loads_instance)
 
 
-def evaluate(predictions, data) -> EvaluationResults:
-    return _compute(predictions=predictions, references=data)
+def evaluate(
+    predictions, dataset: Union[Dataset, IterableDataset], data=None
+) -> EvaluationResults:
+    if data is not None:
+        dataset = data  # for backward compatibility
```
dafnapension, line 205: What if both are not None?

elronbandel: I wanted to rename `data` to `dataset` to have consistent name with load_dataset and create_dataset and DatasetRecipe but yet have backward compatibility for code using `data=`

dafnapension (approved):

src/unitxt/metrics.py
```diff
@@ -317,6 +317,318 @@ def update_and_adjust_global_score(
...
+
+class ConfidenceIntervalMixin(Artifact):
+    n_resamples: int = 1000
+    confidence_level: float = 0.95
+    ci_score_names: List[str] = None
+
+    @abstractmethod
+    def _sample_to_scores(self, sample: List[Any]) -> Dict[str, Any]:
+        pass
+
```
dafnapension, line 335: to my understanding, score_names might include instance metrics as well as global metric.
For instance metric, the instance scores are already computed, and why compute them again?

PR #1465, opened 2025-01-01
Mm updates

alfassy:
DocVQA lmms-eval version added. Default template set and tested vs original. Tested vs lmms-eval with slightly lower accuracy - to be fixed in later update
InfoVqa lmms-eval version added. Default template set and tested vs original. Tested vs lmms-eval with slightly lower accuracy - to be fixed in later update
ChartQA lmms-eval version added. Default template set and tested vs original. Added relaxed_corectness metric. added qa.with_context.with_type task. Accuracy similar to original - ready for use.
AI2D default template set and tested vs original. Added exact_match_mm metric. Accuracy similar to original - ready for use.


alfassy:
I see the errors, will fix and commit fixes.

elronbandel (approved):

PR #1481, opened 2025-01-06
Add place_correct_choice_position to set the correct choice index and…

eliyahabba:
Add place_correct_choice_position to set the correct choice index and update tests


elronbandel (approved):
Thanks again for great contributions!

PR #1493, opened 2025-01-07
complete json_cache to read missing artifacts also from github catalog

dafnapension:


elronbandel:
Fixed in: https://github.com/IBM/unitxt/pull/1494

Issue #1479, opened 2025-01-06
Error message when reading an artifact from github-catalog 


yoavkatz:
This code fails on 1.16.0 and above, and works on 1.15.9.  

```python
import unitxt
from unitxt import load_dataset 
unitxt.settings.allow_unverified_code=True
recipe = "card=cards.doc_vqa.en,template=templates.qa.with_context.simple,loader_limit=100,demos_taken_from=train,augmentor=augmentors.no_augmentation,demos_removed_from_data=True,max_test_instances=20"
dataset = load_dataset(dataset_query=recipe)['test'].select(range(20))
```

This is the error:

```
  File "/Users/yoavkatz/migrate/v1/unitxt/src/unitxt/artifact.py", line 596, in fetch_artifact
    artifact_to_return = catalog.get_with_overwrite(
  File "/Users/yoavkatz/migrate/v1/unitxt/src/unitxt/catalog.py", line 63, in get_with_overwrite
    return self.load(name, overwrite_args=overwrite_args)
  File "/Users/yoavkatz/migrate/v1/unitxt/src/unitxt/catalog.py", line 53, in load
    return Artifact.load(
  File "/Users/yoavkatz/migrate/v1/unitxt/src/unitxt/artifact.py", line 288, in load
    return artifact_link.load(overwrite_args)
  File "/Users/yoavkatz/migrate/v1/unitxt/src/unitxt/artifact.py", line 506, in load
    d = artifacts_json_cache(path)
  File "/Users/yoavkatz/migrate/v1/unitxt/src/unitxt/utils.py", line 127, in artifacts_json_cache
    return load_json(artifact_path)
  File "/Users/yoavkatz/migrate/v1/unitxt/src/unitxt/utils.py", line 131, in load_json
    with open(path) as f:
FileNotFoundError: [Errno 2] No such file or directory: 'https://raw.githubusercontent.com/IBM/unitxt/1.16.0/src/unitxt/catalog/tasks/qa/with_context.json'
```

It seems the code tries load an artifact from a  remote  catalog as a local file. 

 Probably relates to 
 
 
```python
 def load(self, overwrite_args: dict) -> Artifact:
        # identify the catalog for the artifact_linked_to
        assert (
            self.artifact_linked_to is not None
        ), "'artifact_linked_to' must be non-None in order to load it from the catalog. Currently, it is None."
        assert isinstance(
            self.artifact_linked_to, str
[...]

PR #1494, opened 2025-01-08
Simplify artifact link [Non Backward Compatible!]

elronbandel:
Closes: https://github.com/IBM/unitxt/issues/1479


src/unitxt/catalog/temp_recipe_name.json
```diff
@@ -0,0 +1,39 @@
+{
```
yoavkatz, line 1: What is this file?

elronbandel: Deleted

PR #1506, opened 2025-01-13
Tables bench

ShirApp:


src/unitxt/struct_data_operators.py
```diff
@@ -237,7 +237,7 @@ def serialize_table(self, table_content: Dict) -> str:
 
         return (
             "pd.DataFrame({\n"
-            + json.dumps(data_dict)
+            + json.dumps(data_dict)[1:-1]
```
elronbandel, line 240: Is this on purpose?

elronbandel (commented):
Seems ok should we ask @csrajmohan to review?

src/unitxt/struct_data_operators.py
csrajmohan: fine as it is done to avoid duplicate paranthesis(one hard-coded in line 239 so we can probably have a newline after that and another coming from dumps output in line 240).

prepare/cards/tablebench_visualization.py
csrajmohan, line 1: @ShirApp Should we keep this card as we discussed about this that it wont work in current form due to mapping and evaluation issues in the task output and ground truth?

ShirApp: It is not part of the benchmark, but let's remove it to prevent accidental use. Thanks for noticing @csrajmohan !

elronbandel (approved):

PR #1508, opened 2025-01-13
Remove IBM GenAI  support after service was sunset

yoavkatz:


prepare/metrics/llm_as_judge/rating/llama_3_table2text_template.py
```diff
@@ -27,6 +25,6 @@
 
     add_to_catalog(
         metric,
-        f"metrics.llm_as_judge.rating.{model_label}_template_{template_label}",
+        f"metrics.llm_as_judge.rating.{model_label}.{template_label}",
```
yoavkatz, line 28: @elronbandel @ShirApp - Please see the change in metric name I made to make table2text rating be consistent with other rating engine.   As far as I saw the only use was in scigen.py.

ShirApp: It looks fine. Yes, we used it only for scigen.

elronbandel (approved):

PR #1520, opened 2025-01-17
add descriptions and short descriptions

MikolajCharchut:
Adding descriptions and short_descriptions to some of the datasets


elronbandel (changes requested):
In order to implement it you will need to add a new field to the Artifact  class in artifact.py. Can you explain the use case?

MikolajCharchut:
Hi @elronbandel I misunderstood my task, no PR to unitxt is needed. Apologise for the fuss!

PR #1529, opened 2025-01-20
Added key value extraction evaluation and example with images

yoavkatz:
Added a new key value extraction task - which works on both text and images.


docs/docs/examples.rst
```diff
@@ -279,6 +279,18 @@ Evaluate Image-Text to Text Models with different templates and explore the sens
 
 Related documentation: :ref:`Multi-Modality Guide <multi_modality>`, :ref:`Inference Engines <inference>`.
 
+Evaluate Image Key Value Extraction task
++++++++++++++++++++++++++++++++++++++++++
+
+This example demonstrates an image key value extraction task.  It renders several images og given texts and then prompts a vision model to extract key value pairs from the images.
```
elronbandel, line 285:
```suggestion
This example demonstrates an image key value extraction task.  It renders several images of given texts and then prompts a vision model to extract key value pairs from the images.
```

elronbandel (approved):

PR #1541, opened 2025-01-21
Set use 1 shot for wikitq in tables_benchmark

yifanmai:
Use 1 shot instead of 5 for wikitq in tables_benchmark because 5 shots will typically exceed the context length of 32K token models.


yifanmai:
This does make `1_demos` part of the recipe name. I can change it to `1_demo` if you prefer.

ShirApp (approved):

PR #1545, opened 2025-01-22
Renamed criterias in LLM-as-a-Judge metrics to criteria.

tejaswini:


martinscooper:
I would use italic for CrossProviderInferenceEngine

src/unitxt/llm_as_judge.py
```diff
@@ -8,43 +8,16 @@
 from .error_utils import UnitxtError
 from .inference import (
     InferenceEngine,
-    OptionSelectingByLogProbsInferenceEngine,
```
martinscooper, line 11: Although this imports are unused, they are here in order to let users import anything from unitxt.llmasjudge. Lets re-add them.

martinscooper (changes requested):

yoavkatz:
Since we already announced it - we need to merge it as soon as possible.  Any blockers - from doing it to day? Smaller fixes can be done later.

elronbandel (approved):
More changes (related to @martinscooper comments can be made in a separate PR)

PR #1547, opened 2025-01-22
add a filter to wikitq

ShirApp:
there are a few large tables, that cannot be added to the prompts (context window)


csrajmohan (approved):
Looks good to me..

PR #1556, opened 2025-01-26
Improve assistant

elronbandel:
- Fix error in session id handeling
- Separate Model and View
- Add Like/Dislike buttons


src/unitxt/assistant/app.py
```diff
@@ -28,258 +24,298 @@ def wrapped_validate_environment(self, *args, **kwargs):
...
+            np.load(os.path.join(assistant_dir, "embeddings.npz"))["embeddings"],
+            dtype=torch.float32,
+        )
+
+        self.model_name = model_name
+        self.tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
+
+    def search(self, query, max_tokens=5000, min_text_length=50):
+        response = litellm.embedding(
+            model="watsonx/intfloat/multilingual-e5-large",
```
eladven, line 50: That should be a datamember, same as model

elronbandel: Thanks. Will do.

elronbandel: Done

Issue #1559, opened 2025-01-27
slow dataset loading


assaftibm:
It seems that loading a dataset from HF using Unitxt is much slower than doing it using the `datasets` package.

Compare this:
```python
from datasets import load_dataset
from time import time
import os
from uuid import uuid4

path = os.path.join(f"cache/{uuid4()}")

t0 = time()
ds = load_dataset("PrimeQA/clapnq_passages", cache_dir=path)
t1 = time()

print(t1-t0)

print(len(ds))
```

To:
```python
from time import time

from unitxt import load_dataset


t0 = time()
ds = load_dataset('card=cards.rag.documents.clap_nq.en')
t1 = time()

print(t1-t0)

print(len(ds))

```

The Unitxt version takes about x5 times longer.

In both cases a fresh new copy is downloaded.

dafnapension:
Hi, I combined both cases, and ran each to its end, not just load the beginning of the datasets, but actually do the whole process. I also deleted the default huggingface cache before each run:

```python
from datasets import load_dataset as hf_load_dataset
from time import time
import os
from uuid import uuid4
from unitxt.api import load_dataset
import shutil 

path = "/home/dafna/.cache/huggingface"
shutil.rmtree(path)
t0 = time()
ds = hf_load_dataset("PrimeQA/clapnq_passages")
print(f"hf ds: {ds}")
t1 = time()
instances = list(ds["train"])
t2 = time()
print(f"num of instances in ds = {len(instances)}")
print(f"from start to after load_dataset: {t1-t0}")
print(f"from after list of ds to after load_dataset = {t2-t1}")
print()
shutil.rmtree(path)
t0 = time()
ds = load_dataset('card=cards.rag.documents.clap_nq.en')
print(f"unitxt ds: {ds}")
t1 = time()
instances = list(ds["train"])
t2 = time()
print(f"num of instances in ds = {len(instances)}")
print(f"from start to after load_dataset: {t1-t0}")
print(f"from after list of ds to after load_dataset = {t2-t1}")
```

dafnapension:
The printout is:
```
(virtual39) dafna@DESKTOP-GM8R3J7:~/workspaces/unitxt$ python -m tests.mine.checktimes_assaf
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
[...]

PR #1560, opened 2025-01-27
remove IterableDataset, use only Dataset

dafnapension:
Primitively closes: https://github.com/IBM/unitxt/issues/1559


dafnapension:
wad just meant to make a point. Not a real PR

PR #1566, opened 2025-01-29
Add granite documents format

elronbandel:


src/unitxt/formats.py
```diff
@@ -25,6 +26,55 @@ class Format(InstanceOperator):
     pass
 
 
+class GraniteDocumentsFormat(Format):
+    model: str = "ibm-granite/granite-3.1-8b-instruct"
```
OfirArviv, line 30: why set the model? Can't there be multiple models?

tests/library/test_formats.py
```diff
@@ -327,6 +338,132 @@ def test_hf_system_format(self):
...
+            # "metrics.rag.response_generation.recommended.llmaj_watsonx.all",
+            # "metrics.rag.response_generation.recommended.llmaj_rits.all"
+            # "metrics.rag.response_generation.recommended.llmaj_azure.all"
+        ]
+
+        # Verbalize the dataset using the template
+        dataset = load_dataset(
+            card=card,
+            template_card_index="simple",
+            format=GraniteDocumentsFormat(),
```
OfirArviv, line 449: set the model name in explicit way, just in case things change in the future and there are multiple granite ones

OfirArviv (approved):

PR #1568, opened 2025-02-02
Add global mmlu lite sensitivity cards

eliyahabba:
feat: add Global-MMLU-Lite CS/CA task cards

Add two task cards for Global-MMLU-Lite dataset:
- CS card for culturally sensitive questions
- CA card for culturally agnostic questions

Both cards include:
- Support for 14 languages
- Multiple choice QA format
- Topic mapping and preprocessing steps


elronbandel (changes requested):
Is all the difference between the 3 files of global mmlu is the filtering lambda? 
can you just have them in one python file with loop over the lambdas:
`for func in [None, "lambda x: x['cultural_sensitivity_label'] == 'CA'", ...`
Also can you run `make pre-commit` before committing to fix the style of the code (once you run it once it will persist to affect your code before new commits)

eliyahabba:
Not exactly. I combined the two files of cultural_sensitivity_label, but there are important differences between these files and the `global_mmlu` file:

1. different datasets: Global-MMLU-Lite vs Global-MMLU
2. The processing approach is different: the Global-MMLU-Lite files create **one card per language covering all subjects,** while the Global-MMLU file creates **a separate card for each language-subject** combination.

elronbandel (approved):

PR #1572, opened 2025-02-03
include inference time in performance report only when not-mocked

dafnapension:
because inference is mocked, and evaluation is not indicative if runs over mocked predictions


dafnapension:
covered by #1600

PR #1576, opened 2025-02-04
[Breaking change] Add support for all Granite Guardian risks 

martinscooper:
This PR adds support for all available Granite Guardian risks, extending current support of RAG risks, plus bug fixes and several improvements. In summary, this PR:
- BREAKING CHANGE: `GraniteGuardianWMLMetric` was re-implemented and `GraniteGuardianRagRisk` can be used in its place.
- The new implementation defines one base class `GraniteGuardianBase` and classes with name `GraniteGuardian<risk_type>Risk` inherit from it by implementing methods `verify_granite_guardian_config()` and `process_input_fields()`.
- Adds all supported risks (the ones available in [this demo](https://huggingface.co/spaces/ibm-granite/granite-guardian-3.1-8b)).
- Adds support for using any `LogProbInferenceEngine` that offers Granite Guardian models.
- Adds flexibility for input field names.
- Adds the concept of risks types: users risks, assistant risks, RAG risks, agentic risks and custom.
- Adds examples for every risk type.
- Adds catalog for every risk.
- Updates default model version to 3.1.
- Adds verification for early error detections.
- Adds debugging logs.
- Adds generation parameters to match those used in the Granite Guardian cookbook notebooks.
- Fix probability of risk score: certainty was being used.
- Adds support for 'Bring your own risk'.


martinscooper:
I don't agree on calling 'question ' to the default `user_message_field` as it is not always a question. I would prefer `user_message` or `user_prompt`. This change may be not possible because of backward compatibility.

martinscooper:
The RAG Granite Guardian metrics are duplicated now, but not exactly. The current Granite Guardian RAG metric is defined differently in the catalog. `Copy` is used to set data in the task data param ([link to file](https://github.com/IBM/unitxt/blob/main/prepare/metrics/rag_granite_guardian.py)), but there is no example on how to use it. 

@elronbandel let me know if this is the way I should define the metrics in the [prepare file](https://github.com/IBM/unitxt/blob/granite-guardian-support/prepare/metrics/granite_guardian.py).

martinscooper:
@arielge FYI, we are adding all the missing Granite Guardian risks.

elronbandel (approved):

src/unitxt/llm_as_judge_constants.py
```diff
@@ -80,12 +80,10 @@ class EvaluatorNameEnum(str, Enum):
     O1_PREVIEW = "o1-Preview"
     O1_MINI = "o1-Mini"
     GRANITE_13B = "Granite-13b"
-    GRANITE3_2B = "Granite3.0-2b"
-    GRANITE3_8B = "Granite3.0-8b"
+    GRANITE3_2B = "Granite3-2b"
```
elronbandel, line 83: @martinscooper is this breaking change or just internal renaming?

src/unitxt/metrics.py
```diff
@@ -6012,76 +6014,265 @@ def compute(
...
+        RiskType.RAG: ["context_relevance", "groundedness", "answer_relevance"],
+        RiskType.AGENTIC: ["function_call"],
+    }
 
-    _requirements_list: List[str] = ["ibm_watsonx_ai", "torch", "transformers"]
+    _requirements_list: List[str] = ["torch", "transformers"]
 
-    def prepare(self):
-        self.reduction_map = {"mean": [self.main_score]}
+    def verify_guardian_config(self, task_data):
```
elronbandel, line 6083: The following lines are very complicated. Can we instead use Inheritance? So we have `BaseGuerdianGraniteMetric`
Then `RagGuerdianGraniteMetric` where minimal things are changed? Not sure if possible, but I think it will might make this code section as well as it usage more simple

martinscooper: I think it makes sense, I will implement it.

martinscooper: I am having second thoughts on this. I agree on the fact that the code is complex, but adding one inheritence per risk type makes it harder for users to use in case they don't want to use the catalog and they create the classes directly.

martinscooper: @elronbandel done

PR #1577, opened 2025-02-04
Know your splits: proof of the concept of automatically adding split names to cards

dafnapension:
Begin with cards whose loader is LoadHF.


dafnapension:
Having found what we needed: closing this PR:
The only card whose Loader is LoadHF, and whose splits can not be found by `get_dataset_config_info` (or by `get_dataset_split_names` that return the same list of split names, is `prepare/cards/xlsum.py`, for which the split names, found thrugh actually loading the dataset are: `['test', 'train', 'validation']`, and are now added to its card.

PR #1583, opened 2025-02-07
fix: minor bug when only space id is provided for WML inference

tsinggggg:
WML inference is expected to support either 'WML_SPACE_ID' or 'WML_PROJECT_ID' specified, however currently the code does not handle the case where only space id is specified.


tsinggggg:
Hi @elronbandel , would you please review this minor fix? Thank you!

elronbandel (approved):
Thanks

PR #1585, opened 2025-02-08
move gradually to lazy loader

dafnapension:
begin with just better printout of compare bluebench performance


dafnapension:
already covered

PR #1598, opened 2025-02-11
add ragbench faithfulness cards

lilacheden:
1. add ragbench faithfulness cards
2. fix single-reference rag-faithfulness error
3. add mistral large rag judges

### DIFF
--- a/src/unitxt/inference.py
+++ b/src/unitxt/inference.py
@@ -1623,34 +1623,50 @@ class OpenAiInferenceEngine(
 
     @run_with_imap
     def _get_chat_completion(self, instance, return_meta_data):
+        import openai
         messages = self.to_messages(instance)
-        response = self.client.chat.completions.create(
-            messages=messages,
-            model=self.model_name,
-            **self._get_completion_kwargs(),
-        )
-        prediction = response.choices[0].message.content
-        return self.get_return_object(prediction, response, return_meta_data)
+        try:
+            response = self.client.chat.completions.create(
+                messages=messages,
+                model=self.model_name,
+                **self._get_completion_kwargs(),
+            )
+            prediction = response.choices[0].message.content
+            return self.get_return_object(prediction, response, return_meta_data)
+        except openai.BadRequestError as e:
+            logging.error(f"Error predicting instance {messages}:{e}. Retuning empty predictions")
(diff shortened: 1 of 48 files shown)


src/unitxt/inference.py
```diff
@@ -1623,34 +1623,50 @@ def _infer_log_probs(
...
-        return self.get_return_object(prediction, response, return_meta_data)
+        try:
+            response = self.client.chat.completions.create(
+                messages=messages,
+                model=self.model_name,
+                **self._get_completion_kwargs(),
+            )
+            prediction = response.choices[0].message.content
+            return self.get_return_object(prediction, response, return_meta_data)
+        except openai.BadRequestError as e:
```
yoavkatz, line 1636: @lilacheden @elronbandel - when do these errors occur? Is it right to mask them?

lilacheden: I saw them in azure_openai, when the content fails any of their integrated content filters (hate, violence etc). No reason to throw away the entire process/batch in that case

yoavkatz: Thanks. Can you add small comment to explain this?

lilacheden: done

src/unitxt/inference.py
```diff
@@ -1623,34 +1623,50 @@ def _infer_log_probs(
...
+            logging.error(f"Error predicting instance {messages}:{e}. Retuning empty predictions")
```
yoavkatz, line 1638:
```suggestion
            logging.error(f"Error predicting instance {messages}:{e}. Returning empty predictions")
```
```suggestion
            logging.error(f"Error predicting instance {messages}:{e}. Retuning empty predictions")
```

src/unitxt/inference.py
yoavkatz: Did you mush the change?

src/unitxt/inference.py
lilacheden: fixed

src/unitxt/inference.py
lilacheden: yes, added `# catch in case of content_filtering failure`
do you want more details?

yoavkatz (approved):

PR #1599, opened 2025-02-11
touch the dataset to make sure loaded correctly with the streaming value you chose

dafnapension:
For lazy_loaders, this must be done in load_dataset, because it is harmful later.


dafnapension:
moved to #1624

PR #1602, opened 2025-02-12
Make MultipleSourceLoader lazy and fix its use of fusion

elronbandel:


yoavkatz (commented):
Why didn't the  Unitxt unit tests not fail ? Can we make a unit test that validates the fix?

elronbandel:
Added test that validated the fix @yoavkatz

PR #1611, opened 2025-02-17
Fix Ollama inference engine

eladven:


elronbandel (changes requested):
change maybe to if v is not None?
Also this solution should be in to_dict which suppose in my opinion to return only values that were set by the user and no defaults

PR #1624, opened 2025-02-22
Add test to verify integrity of loaded datasets, Avoid copying=True for LoadHF and two more loaders. Remove loader cache (python knows how to manage main memory).

dafnapension:
avoid copying=true (upon iterating over the loaded dataset) for the streams generated by the loader for loaders which watch out after the datasets they load - letting no one modify them. This is the case for LoadHF, LoadCSV, and LoadSKLearn.  For the rest -- copying=True remains.   Also, remove the loader cache. We are not sure what goes there (a whole listed dataset? just its generators?) and python better manages its main memory knowing the whole picture (and remove to file system whole pages as it sees fit)

E.g that demonstrates how the HF dataset, returned by LoadHF, is not modifiable.
``` python

    from datasets import load_dataset as hf_load_dataset
    ds = hf_load_dataset("PrimeQA/clapnq_passages")
    print(ds)
    iter_train = iter(ds["train"])
    iter_train2 = iter(ds["train"])
    for _ in range(5):
        instance = next(iter_train)
        print(instance["id"])
        instance["id"] = 5
    
    for _ in range(5):
        instance = next(iter_train2)
        print(instance["id"])
        
    iter_train = iter(ds["train"])
    iter_train2 = iter(ds["train"])
    for _ in range(5):
        instance = next(iter_train)
        old_value = instance.pop("id")
        instance["id_new"] = old_value
    
    for _ in range(5):
        instance = next(iter_train2)
        print(instance["id"])
        print("id_new" in instance)


```
```python

    DatasetDict({
        train: Dataset({
            features: ['id', 'text', 'title'],
            num_rows: 178890
        })
    })
    827849752_115-357
    827849752_358-743
    827849752_744-1186
    827849752_1187-1426
    827849752_1604-3189
    827849752_115-357
    827849752_358-743
    827849752_744-1186
    827849752_1187-1426
    827849752_1604-3189
    827849752_115-357
    False
    827849752_358-743
    False
    827849752_744-1186
    False
    827849752_1187-1426
    False
    827849752_1604-3189
    False
```

### DIFF
--- a/src/unitxt/loaders.py
+++ b/src/unitxt/loaders.py
@@ -56,11 +56,12 @@ from typing import (
 import pandas as pd
 import requests
 from datasets import (
+    Dataset,
     DatasetDict,
     DownloadConfig,
     IterableDataset,
     IterableDatasetDict,
-    get_dataset_split_names,
+    get_dataset_config_info,
 )
 from datasets import load_dataset as _hf_load_dataset
 from huggingface_hub import HfApi
@@ -75,7 +76,6 @@ from .operators import Set
 from .settings_utils import get_settings
 from .stream import DynamicStream, MultiStream
 from .type_utils import isoftype
-from .utils import LRUCache, recursive_copy
 
 logger = get_logger()
 settings = get_settings()
@@ -83,7 +83,7 @@ settings = get_settings()
 def hf_load_dataset(path: str, *args, **kwargs):
     if settings.hf_offline_datasets_path is not None:
         path = os.path.join(settings.hf_offline_datasets_path, path)
-    return _hf_load_dataset(
+    dataset_to_return = _hf_load_dataset(
         path,
         *args, **kwargs,
             download_config=DownloadConfig(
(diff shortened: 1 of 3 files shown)


src/unitxt/loaders.py
```diff
@@ -194,11 +193,17 @@ def get_splits(self) -> List[str]:
...
-            splits = [self.split]
-        else:
-            splits = self.get_splits()
+    def load_iterables(
+        self
+    ) -> Dict[str, DynamicStream]:
+        # log only once, here:
+        # log once for all splits, as they are limited the same
+        if self.get_limit() is not None:
+            self.log_limited_loading()
```
elronbandel, line 202: This is unnecassry as inside the log limited loading there is a mechanism to log only once. Which means it is logging only once and only when the data is actually loaded.

dafnapension: I think that the mechanism hides problem. Like it hid from you the loop in LoadCSV. 
This logging is cleaner, and happens just once, and does not introduce new variables.

src/unitxt/loaders.py
```diff
@@ -173,7 +172,7 @@ def load_data(self) -> MultiStream:
             raise UnitxtError(f"Error in loader:\n{self}") from e
         if isoftype(iterables, MultiStream):
             return iterables
-        return MultiStream.from_iterables(iterables, copying=True)
+        return MultiStream.from_iterables(iterables)
```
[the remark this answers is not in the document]
dafnapension: Currently, we only reach this line. No Loader returns a MultiStream from its load_iterables.

dafnapension:
closed for no public interest

PR #1627, opened 2025-02-24
small changes to profiler

dafnapension:
+ make `bluebench_profiler` profile any input `dataset_query`, or a list thereof, including a list of cards. Thus making `card_profiler` and `compare_performance_results` (that only dealt with a list of cards) -- unnecessary, hence deleted.
+ add a list of explicit split names to the only card that needs it -- `prepare/cards/xlsum.py` (all the other can retrieve their split names easily, through HF `get_dataset_splits_names`  (or `get_dataset_config_info` which returns the same)
+ small change in docstring of `operators.py` to eliminate a red line in `make docs-server`


elronbandel (approved):

elronbandel:
Thanks @dafnapension !

Issue #1632, opened 2025-02-26
Add hotpotqa dataset a rag end-to-end dataset


BenjSz:
Add hotpotqa dataset a rag end-to-end dataset

https://huggingface.co/datasets/hotpotqa/hotpot_qa

PR #1633, opened 2025-02-26
Benjams/add hotpotqa + change type of metadata field to dict (non backward compatible)

BenjSz:
https://github.com/IBM/unitxt/issues/1632


yoavkatz:
Note the change to the metadata field is not backward compatible, but I'm not sure anyone else is using it.

elronbandel:
What is the purpose of the metadata_field? is it metadata? or reference to a field that holds the metadata? and if no card is using it can we just delete it or ignore it by adding it defaults at the task? @yoavkatz

elronbandel (approved):

PR #1637, opened 2025-02-27
Fix of typo in usage of attributes inside IntersectCorrespondingFields

pklpriv:
As in title, this typo makes code throw out misleading error message.


src/unitxt/operators.py
```diff
@@ -1523,7 +1523,7 @@ def verify(self):
 
         if not isinstance(self.allowed_values, list):
             raise ValueError(
-                f"The allowed_field_values is not a type list but '{type(self.allowed_field_values)}'"
+                f"The allowed_values is not a type list but '{type(self.allowed_values)}'"
```
yoavkatz, line 1526:
```suggestion
                f"The allowed_values is not of type list but of type '{type(self.allowed_values)}'"
```

yoavkatz (approved):

PR #1642, opened 2025-03-02
Performance to report accurate times based on end-to-end time() diffs, rather than accumulate cProfile numbers over methods that seemed relevant.

dafnapension:
-  completed the mechanism to load HF datasets from a local file-system, set up by @elronbandel, to also save into such (all pending settings values), and extended for CSV too.
-  introduced a mechanism to just load the instances that were actually pulled from the raw (typically HF) dataset into unitxt, in order to complete the recipe. Thus, together, can now report more accurately, based on genuine time() reads, the performance times of: 
   - load from HF (or other web sites) down to local file system
   - load from file system to main mem and iterate over the exact instances that actually participated in the benchmark
   - do all the rest: 
      - prepare for inference 
      - infer, and 
      - evaluate.
-  removed the `max_samples_per_subset=30` from `bluebench dataset query`, since when that limitation is in effect, most recipes do not participate in the fused result.

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -44,12 +44,17 @@ jobs:
     - name: Prepare the dirs for performance evaluation in main
       run: |
         mkdir -p performance_action
+        mkdir -p performance_action/hf_fs_cache
         cp performance/bluebench_profiler.py performance_action/bluebench_profiler.py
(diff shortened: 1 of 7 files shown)


dafnapension:
replaced by #1783

PR #1644, opened 2025-03-04
Use elaborated cache key and use it for filelock semaphore

elronbandel:


src/unitxt/api.py
```diff
@@ -132,43 +133,69 @@ def create_dataset(
     card = TaskCard(loader=LoadFromDictionary(data=data), task=task)
     return load_dataset(card=card, split=split, **kwargs)
 
+def to_str(obj):
```
yoavkatz, line 136: I suggest documenting this function (motivation), and giving it a more meanful names.

src/unitxt/api.py
```diff
@@ -132,43 +133,69 @@ def create_dataset(
...
+        lock_file = os.path.join(os.path.expanduser("~"), ".cache", "unitxt", f"{config_name}.lock")
+        os.makedirs(os.path.dirname(lock_file), exist_ok=True)
+
+        # Create a file lock
+        lock = filelock.FileLock(lock_file, timeout=lock_timeout)
+
+        # Only protect the download_and_prepare operation with the lock
+        try:
+            with lock:
+                ds_builder.download_and_prepare(
```
yoavkatz, line 184: If caching its disabled (which is the default) - why not set the cache dir to temporary folder?

If we go with this approach - then we should test it with FM-eval before continuing as well as add a unit test.

src/unitxt/dataclass.py
```diff
@@ -296,6 +296,35 @@ def _asdict_inner(obj):
 
     return copy.deepcopy(obj)
 
+def to_dict(obj, func=copy.deepcopy, _visited=None):
```
yoavkatz, line 299: Needs documentation.

src/unitxt/api.py
elronbandel: Done.

src/unitxt/dataclass.py
elronbandel: Done.

src/unitxt/api.py
OfirArviv: I will test it today

OfirArviv: Will Original error:

[Errno 2] No such file or directory: '/dccstor/fme/users/ofir.arviv/cache/huggingface/datasets/unitxt/recipe-36158aed/1.19.0.incomplete/unitxt-test-00000-00000-of-NNNNN.arrow'

PR #1651, opened 2025-03-06
add 'show more' button for imports from unitxt modules

dafnapension:
E.g.:
image
and 
image


elronbandel:
button
Ideally something like this

PR #1657, opened 2025-03-08
Minor bug fix affecting Text2SQL execution accuracy

oktie:
This fixes a minor bug - the SQL executor can wrap the results in a dictionary inside a 'results' field, and we were comparing the output and gold DataFrames, including the 'results' field. As a result, column names were not ignored. I also revised the logic for comparing DataFrames.


perlitz (approved):
LGTM

PR #1667, opened 2025-03-10
Test branch to evaluate impact of different format 

yoavkatz:


yoavkatz:
Debug completed. Main difference are due to temperature.

yoavkatz:
There is a new need to set template instruction in user role and not ion the system role.  This in some models (e.g. granite) a user system overrides the default system prompt (E.g. one thast supports thinking).

elronbandel (approved):

PR #1677, opened 2025-03-13
HFPipelineInferenceEngine - add loaded tokenizer to pipeline

eladven:
Currently, HFPipelineInferenceEngine failed to load complex checkpoints (e.g. granite after FT). This simple change allow it.


elronbandel (approved):
I approve but is it related to the fast tokenizer? do we understand whats going on?

PR #1682, opened 2025-03-13
Fix some bugs in inference engine tests

elronbandel:


src/unitxt/api.py
```diff
@@ -129,7 +131,11 @@ def create_dataset(
             f"No 'template' was passed to the create_dataset() and the given task ('{task.__id__}') has no 'default_template' field."
         )
 
-    card = TaskCard(loader=LoadFromDictionary(data=data), task=task)
+    args = {"data": data}
+    if data_classification_policy is not None:
+        args["default_data_classification_policy"] = data_classification_policy
+
+    card = TaskCard(loader=LoadFromDictionary(**args), task=task)
```
yoavkatz, line 138: why not :

card = TaskCard(loader=LoadFromDictionary(data=data), task=task, data_classification_policy=data_classification_policy) ?

src/unitxt/schema.py
```diff
@@ -86,6 +85,24 @@ def loads_instance(batch):
         batch["task_data"] = [json.loads(d) for d in batch["task_data"]]
     return batch
 
+def loads_instance(instance):
+    if (
+        "source" in instance
+        and isinstance(instance["source"], str)
+        and (
+            instance["source"].startswith('[{"role":')
```
yoavkatz, line 93: This seems very risk, a minor added space, will cause the code to fail.  Need to consider alternatives.

PR #1690, opened 2025-03-18
Fixes to GraniteGuardian metric,, safety evals cleanups

bnayahu:
- Granite Guardian metric to properly handle data_classification_policy
- Interim fix to make 'prediction' available to the metric.
- Further cleanup to safety evals.


elronbandel (approved):

martinscooper:
@bnayahu @elronbandel note the the following added lines makes to impossible to set an input field named 'prediction' as it will be overwritten:
```
# TODO replace with logic inside verify_granite_guardian_config and process_input_fields
task_data["prediction"] = prediction
```

In addition, we should discuss how the fix should be implemented. Using the predictions only makes sense in the case where the risk is evaluating a generated response (the response). So it would be all the assistant risks and some of the RAG test cases.

bnayahu:
That's true. Perhaps a more delicate way would be to add a flag field to the class (e.g., use_generated_prediction) that could be set in the specification of the metric in the card, then be checked in verify_granite_guardian_config and process_input_fields to either only use task_data or use the prediction field for the assistant field?

PR #1695, opened 2025-03-20
Fix llm judge artifacts

martinscooper:
LLM Judge artifacts are missing the provider param


elronbandel (approved):

prepare/metrics/llm_as_judge/direct/llama_3_3_70b_instruct_adherence_completeness.py
```diff
@@ -89,7 +89,7 @@
 # now = define the judge metric using the criteria
 adherence_metric = LLMJudgeDirect(
     inference_engine=CrossProviderInferenceEngine(  # or your favorite inference model
-        model="llama-3-3-70b-instruct", max_tokens=1024
+        model="llama-3-3-70b-instruct", max_tokens=1024, temperature=0, provider="watsonx"
```
yoavkatz, line 92: Why do we need to set a hard coded provider . The metric name does not include it 

"metrics.rag.response_generation.adherence_with_format.llama_3_3_70b_instruct_judge"

PR #1700, opened 2025-03-25
Vision templates

alfassy:
Hi are these all the changes you wanted?


eladven (approved):

prepare/cards/ai2d.py
```diff
@@ -3,16 +3,10 @@
...
 from unitxt.test_utils.card import test_card
 
 templates = get_from_catalog("templates.qa.multiple_choice.with_context.no_intro.all")
-template = MultipleChoiceTemplate(
-    input_format="{context}\n{question}\n{choices}\nAnswer with the option's letter from the given choices directly.",
-    choices_separator="\n",
-    target_field="answer",
-    enumerator="capitals",
-)
+default_template = get_from_catalog("templates.qa.multiple_choice.with_context.ai2d")
```
yoavkatz, line 9: I don't think you need to fetch form the catalog, you can just place the string and it will automatically fetch it.

eladven (approved):

PR #1701, opened 2025-03-26
Support HFPipelineBasedInferenceEngine with PEFT model

eladven:
Peft models don't have config.json, and therefore AutoConfig.from_pretrained may fail.

### DIFF
--- a/src/unitxt/inference.py
+++ b/src/unitxt/inference.py
@@ -939,14 +939,26 @@ class HFPipelineBasedInferenceEngine(
         if settings.hf_offline_models_path is not None:
             path = os.path.join(settings.hf_offline_models_path, path)
 
-        self.task = (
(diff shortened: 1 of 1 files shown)


src/unitxt/inference.py
```diff
@@ -939,14 +939,26 @@ def _define_task(self):
...
-                path,
-                trust_remote_code=True,
-            ).is_encoder_decoder
-            else "text-generation"
-        )
+        try:
+            # Try loading as a full model (HF model or local full model)
+            config = AutoConfig.from_pretrained(path, trust_remote_code=True)
+
+        except Exception:
```
elronbandel, line 946: What is the Exception you expect to see here?

elronbandel (approved):

PR #1702, opened 2025-03-28
Text2sql metrics fixes

oktie:
This PR fixes a bug with the text2sql metric execution accuracy scores, making the comparison of DataFrames column-order independent again. It also modifies the subset score calculation, returning true if either DataFrame is a subset of the other, while optionally considering row order if the ground truth (gt) includes an ORDER BY clause.


perlitz (approved):
Looks good, results also makes sense

PR #1721, opened 2025-04-09
Fix errors with peft inference engine implementation.

eladven:


src/unitxt/inference.py
```diff
@@ -885,10 +885,10 @@ def _init_model(self):
 
         model_class = (
```
OfirArviv, line 886: this should be model_name

OfirArviv (approved):

src/unitxt/inference.py
```diff
@@ -885,10 +885,10 @@ def _init_model(self):
...
             AutoPeftModelForSeq2SeqLM
```
OfirArviv, line 887: the left model doesn't have this arg in the config?

PR #1722, opened 2025-04-09
Allow overriding of CI method in GlobalMetric

lga-zurich:
This is related to [Exponential time complexity when confidence interval calculation is enabled](https://github.com/IBM/unitxt/issues/1008), and provides a means for GlobalMetrics to override the default "BCa" method to compute confidence intervals.

As it is now, any non-trivial metric computation or moderate number of samples will take forever, as the computation is performed n times over n-1 samples. This allows a GlobalMetric to select the "percentile" method, for example.


elronbandel (approved):
Thanks. As far as I understand, we moved many metrics to inherit from MapReduce base metric instead of GlobalMetric, in that metric the heavy computation is done once per instance and the confidence interval is just resampling the  already mapped values, reducing them each them each time with a cheap computation. 

Your modification is great, just adding for the record and future viewers that with MapReduceMetric we have a way to enjoy both the accuracy of "BCa" and avoid computationally heavy outcomes. 

Thanks for the contribution!

PR #1724, opened 2025-04-09
Add data_files support in HFLoader + tests

elronbandel:


src/unitxt/loaders.py
```diff
@@ -351,6 +351,10 @@ def _maybe_set_classification_policy(self):
     def get_splits(self):
         if self.splits is not None:
             return self.splits
+        if self.data_files is not None:
+            if isinstance(self.data_files, dict):
+                return list(self.data_files.keys())
+            return ["train"]
```
yoavkatz, line 357: Why return "train" by default, if there are no datafile?
What happens if I specify a list of data_files?

elronbandel: This is aligned with all the possible outcomes possible with the huggingface api, each option: list, dict, str is tested.

yoavkatz (approved):

src/unitxt/loaders.py
elronbandel: Done.

PR #1732, opened 2025-04-14
Disable litellm cache

martinscooper:


elronbandel (changes requested):
What is saved in this cache? can you add one sentence as a documentation?

elronbandel:
Closes: https://github.com/IBM/unitxt/issues/1733

martinscooper:
@elronbandel LiteLLM was caching all the responses from the models. Thus, unitxt was caching everything twice, both at the `inference_engine_cache`

> What is saved in this cache? can you add one sentence as a documentation?

@elronbandel I took a closer look. LiteLLM is not caching anything because the `cache` attribute is no longer set. However, `caching=True` is set in the call to `acompletion()`, which is confusing. So, this PR just removes that `caching=True` line.

_Note: I thought LiteLLM caching was being used because I was getting the same result over and over, but it was caused because of the seed._

elronbandel (approved):
Thanks for sorting it out!

PR #1740, opened 2025-04-16
a potential fix for helm

dafnapension:


dafnapension:
I do not understand why a cached copy of the faulty evaluate_cli  is still in /home/runner/.cache/huggingface

PR #1757, opened 2025-04-25
trying to fix PERFORMANCE:  use a github repository as a replacement for the gone HF 'lmsys/arena-hard-browser'

dafnapension:
`'lmsys/arena-hard-browser'` was gone from HF.  `test_preparation` does not catch faulty `prepare/cards/arena_hard/generation/english_gpt-4-0314_reference.py` that tries to access this (gone) dataset, because a missing dataset is considered by `test_preparation`  an error to be ignored.   Performance, however, failed on it. 
So here a potentially good enough dataset is used instead:     
https://github.com/lmarena/arena-hard-auto  that is pointed at from https://paperswithcode.com/dataset/arena-hard

### DIFF
--- a/prepare/cards/arena_hard/generation/english_gpt-4-0314_reference.py
+++ b/prepare/cards/arena_hard/generation/english_gpt-4-0314_reference.py
@@ -2,7 +2,7 @@ from unitxt.blocks import (
     TaskCard,
 )
 from unitxt.catalog import add_to_catalog
-from unitxt.loaders import LoadFromHFSpace
+from unitxt.loaders import LoadCSV
 from unitxt.operators import (
     Apply,
     Copy,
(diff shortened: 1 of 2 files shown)


prepare/cards/arena_hard/generation/english_gpt-4-0314_reference.py
```diff
@@ -13,35 +13,39 @@
 from unitxt.stream_operators import DeleteSplits, JoinStreams
 from unitxt.test_utils.card import test_card
 
+# 'lmsys/arena-hard-browser' (https://huggingface.co/spaces/lmarena-ai/arena-hard-browser) was gone from HF.
+# https://paperswithcode.com/dataset/arena-hard points at git repository  https://github.com/lmarena/arena-hard-auto
+# from which the dataset is now fetched:
+
+question_git_repo_file_path = "https://raw.githubusercontent.com/lmarena/arena-hard-auto/main/data/arena-hard-v0.1/question.jsonl"
+model_answer_git_repo_file_path = "https://raw.githubusercontent.com/lmarena/arena-hard-auto/main/data/arena-hard-v0.1/model_answer/gpt-4-0314.jsonl"
```
elronbandel, line 21: Can you tie it to a specific commit like here: https://github.com/IBM/unitxt/blob/main/prepare/cards/bfcl.py
?
More robust for future changes

elronbandel (changes requested):
See comment

prepare/cards/arena_hard/generation/english_gpt-4-0314_reference.py
dafnapension: Yes, Done. A good idea! I did not think about this.

elronbandel (approved):

PR #1767, opened 2025-05-04
fixed a typo in examples_tests.yml, and now see that it fails for other reasons

dafnapension:
Fixed a typo in examples_tests.yml to make it run. And now it fails for other reasons.


dafnapension:
Included in https://github.com/IBM/unitxt/pull/1769

PR #1779, opened 2025-05-08
Fix HF AutoModel tokenization issue with chat template + issue with multi GPU

OfirArviv:
When tokenizing a chat template, one should use `"add_special_token"=False`
We are currently not doing so, and because of that our tokenized text start with 2 `"<begin_of_text>"` tokens.
This is also done in the pipeline code, see https://github.com/huggingface/transformers/blob/5c47d08b0d6835b8d8fc1c06d9a1bc71f6e78ace/src/transformers/pipelines/text_generation.py#L311

Also removed the line that put the model on a specific gpu, which caused conflicts with the multi-gpu mechanism.


elronbandel (changes requested):
Great catch! can you remove test2.py or add it to the actual inference tests?

OfirArviv:
> Great catch! can you remove test2.py or add it to the actual inference tests?

done

src/unitxt/inference.py
```diff
@@ -714,17 +714,17 @@ def _init_model(self):
...
     def prepare_inputs(self, data: Iterable) -> Mapping:
+        tokenizer_kargs = {}
         if isinstance(data[0], list):
             data = self.processor.apply_chat_template(
                 data,
                 tokenize=False,
                 add_generation_prompt=True,
                 **self.chat_kwargs_dict,
             )
+            tokenizer_kargs.update({"add_special_tokens": False})
```
elronbandel, line 727:
```suggestion
            tokenizer_kargs["add_special_tokens"] = False
```

OfirArviv: done

elronbandel (approved):

PR #1786, opened 2025-05-11
FIx granite guardian agentic metric and align it with unitxt built in tool calling types

elronbandel:
Other than fixing this broken example, the alignment with unitxt types can allow unitxt users to evaluate benchmarks like bfcl with granite guardian metric


src/unitxt/metrics.py
```diff
@@ -6250,7 +6250,7 @@ def compute(self, references: List[Any], prediction: Any, task_data: Dict) -> di
         return result
 
     def create_message(self, role: str, content: str) -> List[Dict[str, str]]:
-        return [{"role": role, "content": content}]
+        return [{"role": role, "content": str(content)}]
```
martinscooper, line 6253: From your changes, we are already checking that the content is a string (for call and tool fields), so I don't think a conversion to string is required.

examples/evaluate_granite_guardian_agentic_risks.py
```diff
@@ -1,33 +1,105 @@
...
+from unitxt.types import Tool, ToolCall
 
 print("Evaluation of Function Calling Hallucination in Agentic workflow")
 
 data = [
     {
-        "tools": '[\n  {\n    "name": "comment_list",\n    "description": "Fetches a list of comments for a specified IBM video using the given API.",\n    "parameters": {\n      "aweme_id": {\n        "description": "The ID of the IBM video.",\n        "type": "int",\n        "default": "7178094165614464282"\n      },\n      "cursor": {\n        "description": "The cursor for pagination to get the next page of comments. Defaults to 0.",\n        "type": "int, optional",\n        "default": "0"\n      },\n      "count": {\n        "description": "The number of comments to fetch. Maximum is 30. Defaults to 20.",\n        "type": "int, optional",\n        "default": "20"\n      }\n    }\n  }\n]',
-        "user_prompt": "Fetch the first 15 comments for the IBM video with ID 456789123.",
-        "assistant_response": '[\n  {\n    "name": "comment_list",\n    "arguments": {\n      "video_id": 456789123,\n      "count": 15\n    }\n  }\n]',
+        "tools": [
```
martinscooper, line 13: I noticed you slighly changed the structured of the tools content in the example. The `properties` level didn't exist before. Is there a specific reason why it was added? I am not aware of the standard of the struture of the tools nor with which tools GG was trained with. 

I guess that if the example returns a `Yes` from GG and that if we correct the assistant response GG returns a `No` it is fine.

elronbandel: yes, this is the standard way to define tools, also when we define it that way with unitxt Tool type it verifies it is defined well.

PR #1793, opened 2025-05-14
Add ToRR Benchmark Readme file

csrajmohan:
Added a simple, dedicated README file with basic details and run instructions for the ToRR benchmark in Unitxt.


ShirApp (commented):
It looks great!
Should we also modify the current examples for running ToRR to use the same inference engine as here? (since this works for sure)

PR #1799, opened 2025-05-16
Make api_key_env_var optional in LoadFromAPI

martinscooper:
As the docstring states, `api_key_env_var` should be optional:

```python
 api_key_env_var (str, optional):
     The name of the environment variable holding the API key.
```

Currently, it has a default value of `SQL_API_KEY`, forcing users to set that env var even if the endpoint doesn't require any api key.

This PR makes `api_key_env_var` optional.


elronbandel:
@perlitz will this new default break any of the sql pipelines?

perlitz:
I'll make sure it doesn't

PR #1804, opened 2025-05-19
Add more llmjudge benchmarks

martinscooper:


martinscooper:
@elronbandel The newsroom dataset poses a challenge. Both summary and article are stored in the same string. See dataset [here](https://raw.githubusercontent.com/dmg-illc/JUDGE-BENCH/refs/heads/master/data/newsroom/newsroom.json). For example:

```
"You will be shown the response of a language model to a problem of propositional logic. Your task is to judge whether the model's reasoning is sound, i.e. logically valid.\nYou are first presented with the PROBLEM STATEMENT that the model has been given. Subsequently, the model's RESPONSE is shown. Indicate with a 'yes' if the model's response (in particular its rationale) is sound, and conclude 'no' if it is not.\n\n### PROBLEM STATEMENT\n\nExplain whether the conclusion logically follows from the set of statements below. Please report all your reasoning steps. \nEnd your reasoning with: Conclusion: True/False.\n\nStatements:\nEither there is a white marble in the box or else there is a black marble in the box, but not both\nEither there is a black marble in the box or else there is a pink marble in the box, but not both\nThere is a pink marble in the box if and only if there is a gray marble in the box\nConclusion: If there is a white marble in the box then there is a gray marble in the box\n\nLet's think step by step.\n\n### MODEL RESPONSE\n\n Great, let's start!\n\nOkay, so I have the set of statements given to me:\n\n1. Either there is a white marble in the box or else there is a black marble in the box, but not both\n2. Either there is a black marble in the box or else there is a pink marble in the box, but not both\n3. There is a pink marble in the box if and only if there is a gray marble in the box\n4. If there is a white marble in the box then there is a gray marble in the box\n\nLet me think aloud as I work through this problem:\n\nOkay, so the first statement says that there is either a white marble or a black marble in the box, but not both. That means that if there is a white marble in the box, then there can't be a black marble in the box, and vice versa.\n\nThe second statement says that there is either a black marble or a pink marble in the box, but not both. So if there is a black marble in the box, then there can't be a pink marble in the box, and vice versa.\n\nThe third statement says that if there is a pink marble in the box, then there is a gray marble in the box.\n\nNow, let's apply these statements to the conclusion: If there is a white marble in the box, then there is a gray marble in the box.\n\nSo, let's think about this... if there is a white marble in the box, then there can't be a black marble in the box, because of statement 1. And if there is no black marble in the box, then there must be a gray marble in the box, because of statement 3. Therefore, the conclusion logically follows from the given statements, so I will say:\n\nConclusion: True\n\n"
```

Do you think it is possible to parse the content and store in `article` and `summary` fields?

martinscooper:
The same applies to other datasets in this benchmark

martinscooper:
Example on how to run a judge bench card:

```python
from unitxt.api import evaluate, load_dataset
from unitxt.inference import MetricInferenceEngine


dataset = load_dataset(
    card="cards.judge_bench.dices.safety",
    split="test",
    loader_limit=3
)


params = "[criteria_field=criteria,context_fields=user_prompt,check_positional_bias=False]"

metric_inference_engine = MetricInferenceEngine(
    metric=f"metrics.llm_as_judge.direct.watsonx.llama3_3_70b{params}",
    prediction_field="assistant_response",
)

predictions = [p["assistant_message_general_harm"] for p in metric_inference_engine.infer(dataset)]

results = evaluate(predictions=predictions, data=dataset)
parsed_results = {"spearmanr": results.global_scores["spearmanr"], "accuracy": results.global_scores["accuracy"]}
print(parsed_results)
```

PR #1805, opened 2025-05-20
Fix: number of batches calculation is incorrect

martinscooper:
This PR corrects the following line:

```python
number_of_batches = len(dataset) // self.cache_batch_size + 1
```

If the result of the `//` calculation has no decimal part it is not required to add 1. For example if dataset len is 200 and batch_size is 100: 200 // 100 + 1 = 3 and not 2.

Correcte line:

```python
number_of_batches = math.ceil(len(dataset) / self.cache_batch_size)
```


elronbandel (approved):
Great catch! I think this should be merged!

PR #1816, opened 2025-05-29
Refactor JaccardIndex and StringContainment metrics to use ReductionInstanceMetric

elronbandel:


src/unitxt/metrics.py
```diff
@@ -2062,24 +2058,18 @@ def compute(
...
-
-    prediction_type = Any  # string representation is compared
+    reduction = MeanReduction()
 
-    def compute(
-        self, references: List[Any], prediction: Any, task_data: List[Dict]
-    ) -> dict:
-        result = {
+    def map(
+        self, prediction: str, references: List[str], task_data: Dict[str, Any]
```
yoavkatz, line 2066: containment, can also work on other types than str. So "Any" type is right, no?

yoavkatz:
Tehre is a failure in one of the prepare files.

src/unitxt/metrics.py
```diff
@@ -2008,21 +2008,20 @@ def _compute_f1(sa, sb):
...
 
-    def compute(
-        self, references: List[Any], prediction: Any, task_data: List[Dict]
-    ) -> dict:
         if not isinstance(prediction, set):
             prediction = set(prediction)
         references = [set(reference) for reference in references]
 
-        result = {
+        return {
```
dafnapension, line 2024: suggesting to simplify:
float( len(reference.intersection(prediction))  / len(reference.union(prediction)) )

PR #1827, opened 2025-06-11
Add support for tool calling in HFAutoModelInferenceEngine

elronbandel:


src/unitxt/struct_data_operators.py
```diff
@@ -759,13 +760,17 @@ class ToolCallPostProcessor(FieldOperator):
     allow_failure: bool = False
 
     def process_value(self, value: str) -> ToolCall:
+        match = re.search(r"(\{.*\}|\[.*\])", value, re.DOTALL)
```
yoavkatz, line 763: What does this do? Add a comment for reason.

PR #1838, opened 2025-06-23
Improved error messages

elronbandel:
Above every error message there is context for the error:
```bash
raise ValueError(
┌──────────────────────────────┐
│ Unitxt Error Context         │
│  - Object: LoadFromDictionary│
│  - Unitxt: 1.24.0            │
│  - Python: 3.10.18           │
│  - Function: verify()        │
│  - Help: link                │
└──────────────────────────────┘

ValueError: Passed data to LoadFromDictionary is not of type Dict[str, List[Dict[str, Any]]].
Expected data should map between split name and list of instances.
Received value: 
```
Where the object name link to its docs, and the help link is set differently in different places.

### DIFF
--- a/examples/evaluate_tool_calling.py
+++ b/examples/evaluate_tool_calling.py
@@ -7,7 +7,7 @@ from unitxt.inference import (
 logger = get_logger()
 
 sum_tool = {
-    "name": "add_numbers",
+    "mame": "add_numbers",
     "description": "Add two numbers together and return the result.",
     "parameters": {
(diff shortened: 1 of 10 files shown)


yoavkatz:
Hi. This is very nice and a great improvement! I can not wait to see to the users' input.

I think we should agree on the stages:

Dataset loading 
     Raw dataset download
     Schema Validation
     
Inference
     
Metrics
      Metric loading
      Metric evaluation

PR #1845, opened 2025-06-26
Allow using python functions instead of operators (e.g in pre-processing pipeline)

elronbandel:
If you cannot find operators fit to your needs simply use function to modify every instance in the data:

```python
        def my_function(instance, stream_name=None):
            instance["x"] += 42
            return instance
```

Or a function that modify the entire stream:

```python

        def my_other_function(stream, stream_name=None):
            for instance in stream:
                instance["x"] += 42
                yield instance
```
Both functions can be plugged in every place in unitxt requires operators, e.g pre-processing pipeline.

### DIFF
--- a/prepare/cards/xlam_function_calling.py
+++ b/prepare/cards/xlam_function_calling.py
@@ -12,17 +12,6 @@ from unitxt.splitters import RenameSplits
 from unitxt.struct_data_operators import LoadJson
 from unitxt.test_utils.card import test_card
 
-
-def extract_required_parameters(instance, stream_name=None):
-    for tool in instance["tools"]:
-        required_params = []
-        for param_name, param_info in tool["parameters"]["properties"].items():
-            if "optional" not in param_info["type"].lower():
-                required_params.append(param_name)
-        tool["parameters"]["required"] = required_params
-    return instance
-
-
 card = TaskCard(
     loader=LoadHF(
         path="Salesforce/xlam-function-calling-60k",
--- a/prepare/cards/bfcl.py
+++ b/prepare/cards/bfcl.py
@@ -5,7 +5,6 @@ from unitxt.loaders import LoadJsonFile
 from unitxt.operators import (
     Copy,
     ExecuteExpression,
-    FixJsonSchemaOfToolParameterTypes,
     Set,
 )
 from unitxt.stream_operators import JoinStreams
@@ -34,7 +33,7 @@ with unitxt.settings.context(allow_unverified_code=True):
                 ),
                 Copy(field="question/0/0/content", to_field="query"),
                 Copy(field="function", to_field="tools"),
-                FixJsonSchemaOfToolParameterTypes(),
+                "operators.fix_json_schema",
                 # Process ground truth data in this dataset, which is a provided as a list of options per field,
                 # and convert it into a list of explicit tool calls
                 #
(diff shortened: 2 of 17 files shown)


src/unitxt/text_utils.py
```diff
@@ -295,6 +296,36 @@ def construct_dict_as_python_lines(d, indent_delta=4) -> List[str]:
...
+                        result_lines.append(line[base_indent:])
+                    else:
+                        result_lines.append(line.lstrip())
+
+            return result_lines
+
+        except (OSError, TypeError):
+            # If source is not available
+            return [f"<function {d.__name__} (source unavailable)>"]
+
```
dafnapension, line 328: very nice,  this is how the result looks at the catalog:

image

prepare/cards/xlam_function_calling.py
```diff
@@ -3,14 +3,26 @@
...
     Copy,
-    ExecuteExpression,
     Move,
     Set,
 )
 from unitxt.splitters import RenameSplits
 from unitxt.struct_data_operators import LoadJson
 from unitxt.test_utils.card import test_card
 
+
```
dafnapension, line 13: this very dataset is gated,  I suggest to use an example that is accessible to all:

`Dataset 'Salesforce/xlam-function-calling-60k' is a gated dataset on the Hub. You must be authenticated to access it.`

prepare/cards/xlam_function_calling.py
```diff
@@ -30,10 +42,7 @@
             to_field="tools/*/parameters/properties",
             set_every_value=True,
         ),
-        ExecuteExpression(
-            to_field="required",
-            expression="[[p for p, c in tool['parameters']['properties'].items() if 'optional' not in c['type']] for tool in tools]",
-        ),
+        extract_required_parameters,
         Copy(
```
dafnapension, line 46: Being unauthorized for this dataset, I copied the one example instance from 
`https://huggingface.co/datasets/Salesforce/xlam-function-calling-60k`

and ran the following:
[xlam_function_calling.pdf](https://github.com/user-attachments/files/20960177/xlam_function_calling.pdf)

prepare/cards/xlam_function_calling.py
```diff
@@ -30,10 +42,7 @@
...
             field="required",
```
dafnapension, line 47: Please see detailed comment above

dafnapension (changes requested):

prepare/cards/xlam_function_calling.py
elronbandel: @dafnapension it is very easy to access it with huggingface-login, it almost became a standard this days

dafnapension:
Hi @elronbandel , I think I now managed to correctly fix the jsonschema issue (rather than translate all to 'type': 'object' which I erroneously suggested Friday).   
Here is a notebook that runs through the whole recipe output, emphasizing instance 15  that has this special case of 'str, optional' which failed the runs before.

dafnapension:
image

dafnapension:
For the record: to this PR I added Operator  `FixJsonSchemaOfToolParameterTypes` to deal with complicated recursive types which `"operators.fix_json_schema"` does not deal with. E.g. attached from dataset `Salesforce/xlam-function-calling-60k`.  Besides that, `bfcl` still has problems which we deal with in a separate PR.

image

PR #1874, opened 2025-07-20
Handle a case in ToolCallPostProcessor where prediction is an empty list of tools

yoavkatz:
Also improved error message printouts


src/unitxt/operators.py
```diff
@@ -537,7 +541,10 @@ def process(
                     old_value = self.get_default
 
             with error_context(
-                self, field=from_field, action="Process Field", value=old_value
+                self,
+                field=from_field,
+                action="Process Field",
+                value=_limited_repr(old_value),
```
elronbandel, line 547:
```suggestion
```
Not sure about this, I think the error context should point on where there error was not what the cause is, this should be the role of the error itself. my fear is that it will get too busy to read and hide the actual error.

elronbandel (approved):
See my suggestion regarding the context error.  I approved since I dont want to delay the unrelated bug fix.

PR #1879, opened 2025-07-27
extend condition to also filter by field exists or not

dafnapension:
Currently, Filter by condition throws an exception if a field participating in the condition is missing from the instance.
Suggesting here to extend conditions to also handle the very existence of the field in the instance as a measure of this instance being required.  
This can be handy for cleaning datasets in some of the cards, which now throw exception for a missing field (breaking the pre-processing pipeline), and also, indirectly, shed light on cards that are totally erroneous, and are better be removed from the catalog:
E.g. for helpful cleaning:
Currently:
image

whereas [pushing the new operators in the card](https://github.com/IBM/unitxt/pull/1879/files#diff-c0511cc100f69c845f48afdd069a0e19f4c08138fd4e0cb81a8d35adefbc5ca5), yields:

image

Another example, that makes us consider a removal of a totally erroneous card from the catalog:
image

while with the new filter, we find that the whole subset is missing that field, and we may want to remove the card (for subset 'experts') from the catalog altogether:

image

Just to prove the point that the whole dataset, of subset experts, does not match the preprocessing steps:
image

as compared to a subset that does match:
image

### DIFF
--- a/prepare/cards/judge_bench/inferential_strategies.py
+++ b/prepare/cards/judge_bench/inferential_strategies.py
@@ -9,7 +9,7 @@ from unitxt.blocks import (
 from unitxt.catalog import add_to_catalog
 from unitxt.llm_as_judge_constants import DirectCriteriaCatalogEnum
 from unitxt.loaders import LoadJsonFile
-from unitxt.operators import Copy, Set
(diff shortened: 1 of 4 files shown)


elronbandel (approved):
Thanks @dafnapension !

PR #1880, opened 2025-07-28
For load_dataset, use_cache default value is taken from settings

eladven:


elronbandel (approved):
I would use a more specific name like :
use_dataset_cache (not sure)
and worth aligning with other caching names (inference, hf datasets)
And finally add to `settings.rst` which documents all the available settings

PR #1884, opened 2025-07-30
fix performance test 

dafnapension:
through removal from performance evaluation of `_source_to_dataset`, which generated a 403 to the dataset of recipe `cfpb_product_2023`, although the access to this dataset is free no problem:

image


bnayahu:
@dafnapension Shalom Dafna! Please consider reverting this change, once #1888 is merged.

PR #1887, opened 2025-07-30
Biggen benchmark and pearson correlation metric

martinscooper:


elronbandel (approved):
Thanks for the simplification of the correlation metrics! nice!

PR #1928, opened 2025-08-24
remove redundant split from airbench2024

dafnapension:
**BEFORE:**

[safety_airbench2024_main.pdf](https://github.com/user-attachments/files/21957492/safety_airbench2024_main.pdf)

**AFTER:**
[safety_airbench2024_fixed.pdf](https://github.com/user-attachments/files/21957500/safety_airbench2024_fixed.pdf)


elronbandel (approved):

bnayahu:
@dafnapension actually that split is needed, but I haven't yet figured out the proper way to use it...
It contains the judge prompts that AIrbench use for each of the prompt categories. At the time I couldn't find a way to pass is to the LLM as Judge metric (effectively a dynamic template). Perhaps there's a way to do it now? Any help would be appreciated...

cc: @martinscooper

PR #1939, opened 2025-09-20
light fast removal of register_all_artifacts  for unitxt classes

dafnapension:
remove `register_all_artifacts` from `unitxt.__init__`.  Instantiate classes upon demand (only the needed ones) by finding the module, and then use import tools.

### DIFF
--- a/src/unitxt/artifact.py
+++ b/src/unitxt/artifact.py
@@ -48,40 +48,30 @@ def import_module_from_file(file_path):
     return module
 
 
-# type is read from a catelog entry, the value of a key "__type__"
-def get_class_from_artifact_type(type: str):
-    if type in Artifact._class_register:
-        return Artifact._class_register[type]
+# snake_case_class_name is read from a catelog entry, the value of a key "__type__"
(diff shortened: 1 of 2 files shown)


yoavkatz:
Hi Dafna. Indeed the requirement to register classes is a cumbersome one. If I understand right, this proposed solution, searchs for a class in the unitxt folders.   Unitxt allows extension in external catalogs and code, and people use it, so the search path to find classes is not only unitxt.  I'm not sure how this approach can be extended to cover this.

dafnapension:
Thank you, @yoavkatz , this indeed is one missing piece of information for me:  Where else is `register_all_classes` invoked, and over which modules/classes?   Does the usual mode of operation of a user contain the `register_all_classes` that I see in `src/unitxt/__init__`, as well as other `register_all_artifacts`?  Will you share an example of such with me?
Are all classes, registered in all points of registration, registered together in `Artifact._class_register` ?

Will a high-level registration of such points of class-registration in the project directory do as a solution for the issue you raised?

PR #1963, opened 2026-02-16
fix: Disable test_card for MILU dataset due to gated access

yoavkatz:
The MILU (Multilingual Language Understanding) dataset is gated and requires special authentication/approval to access. This change comments out the test_card() call to prevent test failures in automated environments while maintaining the card preparation functionality for users with proper access.

Changes:
- Commented out test_card(card, strict=False) in prepare/cards/milu.py
- Added explanatory comment about gated dataset requirement


yoavkatz:
Added HF Token to CI instead.

